PR #2, opened 2016-04-01
Add RedisClient class and ScopeLogger class

kcudnik:


stcheng:
i remove the previous commit.

common/logger.h
```diff
@@ -27,6 +29,19 @@ class Logger
     static Logger &getInstance();
     static void setMinPrio(Priority prio);
     void write(Priority prio, const char *fmt, ...);
+
+    class ScopeLogger
```
lguohan: why define this inside the Logger class? why not outside of it?

common/redisclient.h
```diff
@@ -0,0 +1,63 @@
...
+
+        void set(std::string key, std::string value);
+
+        void hset(std::string key, std::string field, std::string value);
+
+        void mset(std::unordered_map<std::string, std::string> map);
+
+        void hmset(std::string key, std::unordered_map<std::string, std::string> map);
+
+        std::shared_ptr<std::string> get(std::string key);
```
lguohan: is this function implemented? I cannot find it.

common/redisclient.cpp
```diff
@@ -0,0 +1,166 @@
...
+RedisClient::RedisClient(swss::DBConnector *db):
+    m_db(db)
+{
+}
+
+int64_t RedisClient::del(std::string key)
+{
+    char *temp;
+    int len = redisFormatCommand(&temp, "DEL %s", key.c_str());
+    std::string del(temp, len);
```
lguohan: isn't this an extra copy operation that we can avoid?

common/redisclient.cpp
```diff
@@ -0,0 +1,166 @@
...
+    free(temp);
+
+    RedisReply r(m_db, del, REDIS_REPLY_INTEGER, true);
+
+    if (r.getContext()->type != REDIS_REPLY_INTEGER)
+        throw std::runtime_error("DEL operation failed");
+
+    return r.getContext()->integer;
```
lguohan: why not return true or false as you only remove one key. the integer can be either 1 or 0.

common/logger.h
kcudnik: there is an issue to fully separate it from logger class since they both depend on each oder

common/redisclient.h
kcudnik: they are not implemented yet, but no one is using them yet, so no compiler error

common/redisclient.cpp
kcudnik: what copy operation ?

common/redisclient.cpp
kcudnik: i keep consistent with redis interface, that del return integer

common/redisclient.cpp
lguohan: string copy, copy the temp string into new del string.

kcudnik: this is probably not optimal now, but dedis takes string, and if i pass char* it may be possible to miss free if redis reply throws

common/redisclient.h
```diff
@@ -0,0 +1,63 @@
...
+namespace swss
+{
+
+class RedisClient
+{
+    public:
+
+        RedisClient(swss::DBConnector *db);
+
+        int64_t del(std::string key);
```
eladraz: I'm not sure I understand why we need this class at all. We have started to implement generic function under Table.h.
Most of the suggested function are already implemented there (In your previous commit):

Table::del
Table::delField
Table::setField

Why do we need this class?
@kcudnik @lguohan @stcheng

common/logger.h
eladraz: I think it's a very good patch, however, as Guohan wrote, I don't think it should be within Logger class.
In addition, it would be nice to expand this patch and include thread-id

common/logger.h
```diff
@@ -43,10 +45,12 @@ class Logger
     };
 
 private:
-    Logger() {};
+    Logger();
     Logger(const Logger&);
     Logger &operator=(const Logger&);
 
+    std::string m_self;
```
eladraz: m_self is just set but never used.

common/redisreply.cpp
```diff
@@ -13,24 +13,35 @@ namespace swss {
 
 RedisReply::RedisReply(DBConnector *db, string command, int exepectedType, bool isFormatted)
```
eladraz: I would like to split this patch into two patches

common/redisclient.cpp
```diff
@@ -129,6 +129,40 @@ int64_t RedisClient::decr(std::string key)
...
+        std::shared_ptr<std::string> ptr(new std::string(reply->str));
+        freeReplyObject(reply);
+        return ptr;
+    }
+
+    freeReplyObject(reply);
+
+    throw std::runtime_error("GET failed, memory exception");
+}
+
```
eladraz: Why do we need it? We are using HSET and HGETALL (implemented in Table::get)
Why do we need to add another layer of complexity and duplication?

common/redisclient.h
kcudnik: Table class will not provide all functionality that i need, and those functions from table will be removed later on

common/logger.h
kcudnik: not yet used, but saved for future use

common/redisclient.cpp
kcudnik: those functions from table will be removed later on

PR #4, opened 2016-04-11
Swss.common.multi.consumer

hrachya-mellanox:


stcheng:
@lguohan @eladraz

common/consumertable.cpp
```diff
@@ -37,7 +37,10 @@ ConsumerTable::ConsumerTable(DBConnector *db, string tableName) :
         }
         catch (...)
         {
-            delete m_subscribe;
+            if(m_subscribe) {
```
kcudnik: on constructor entry you could add m_subscribe = NULL;

common/select.cpp
```diff
@@ -8,17 +10,28 @@ using namespace std;
 
 namespace swss {
 
-void Select::addSelectable(Selectable *c)
+void Select::addSelectable(_in_ Selectable *selectable)
 {
-    m_objects.push_back(c);
+    if(std::find(m_objects.begin(), m_objects.end(), selectable) != m_objects.end()) {
+        SWSS_LOG_WARN("Selectable:%p already been added to the list, ignoring\n.");
```
kcudnik: "\n." is not needed, we can enforce in logs to add \n at the end but currently it goes to syslog so no need

common/table.h
```diff
@@ -38,11 +38,11 @@ class Table {
     void delField(std::string key, std::string field);
 
     virtual ~Table();
+    std::string getTableName()const {return m_tableName;};
```
kcudnik: lets keep all code in cpp files ?

eladraz: This function isn't in used. Why do you need it?

eladraz:
In general, you should break this commit to 3 differet commits. There are too many logical changes here.

common/consumertable.cpp
```diff
@@ -37,7 +37,10 @@ ConsumerTable::ConsumerTable(DBConnector *db, string tableName) :
...
+                delete m_subscribe;
+                m_subscribe = NULL;
```
eladraz: There is no needed to "if (m_subscribe)", it's ok to delete NULL.
As for "m_subscribe = NULL;" it's a good change.

common/consumertable.cpp
```diff
@@ -47,7 +50,9 @@ ConsumerTable::ConsumerTable(DBConnector *db, string tableName) :
 
 ConsumerTable::~ConsumerTable()
 {
-    delete m_subscribe;
+    if(m_subscribe) {
+        delete m_subscribe;
+    }
```
eladraz: This change is not needed. It's ok to delete NULL in C++

common/scheme.h
```diff
@@ -1,17 +1,32 @@
...
+#define APP_INTF_TABLE_NAME             "INTF_TABLE"
+#define APP_NEIGH_TABLE_NAME            "NEIGH_TABLE"
+#define APP_ROUTE_TABLE_NAME            "ROUTE_TABLE"
+
+#define APP_TC_TO_QUEUE_MAP_TABLE_NAME  "TC_TO_QUEUE_MAP_TABLE"
+#define APP_SCHEDULER_TABLE_NAME        "SCHEDULER_TABLE"
+#define APP_DSCP_TO_TC_MAP_TABLE_NAME   "DSCP_TO_TC_MAP_TABLE"
+#define APP_QUEUE_TABLE_NAME            "QUEUE_TABLE"
+#define APP_PORT_QOS_MAP_TABLE_NAME     "PORT_QOS_MAP_TABLE"
+#define APP_WRED_PROFILE_TABLE_NAME     "WRED_PROFILE_TABLE"
```
eladraz: Should be in different commit.

common/scheme.h
```diff
@@ -1,17 +1,32 @@
...
+
+
+
```
eladraz: Why so many empty lines?

common/select.cpp
```diff
@@ -8,17 +10,28 @@ using namespace std;
...
+        return;
+    }
+    m_objects.push_back(selectable);
```
eladraz: I think it should be ASSERT() and in DEBUG mode only. There is no point for runtime check like that in production.

common/select.cpp
```diff
@@ -8,17 +10,28 @@ using namespace std;
...
+}
+    
+void Select::addSelectables(_in_ std::vector<Selectable *> &selectables)
+{
+    for(auto it : selectables) {
+        addSelectable(it);
+    }
```
eladraz: I'm against these type of changes. I think that the Select class should do one thing, and adding vector of Selectable * sin't one of them.

common/select.cpp
```diff
@@ -8,17 +10,28 @@ using namespace std;
...
+    }    
 }
 
 void Select::addFd(int fd)
 {
     m_fds.push_back(fd);
 }
 
-int Select::select(Selectable **c, int *fd, unsigned int timeout)
+int Select::select(_out_ Selectable **c, _out_ int *fd, _in_ unsigned int timeout)
```
eladraz: @stcheng are you ok with the _out_ _in_ weird convention?

common/select.h
```diff
@@ -7,13 +7,15 @@
 #include <hiredis/hiredis.h>
 #include "selectable.h"
 
+
 namespace swss {
```
eladraz: Why the extra space?

common/table.h
```diff
@@ -38,11 +38,11 @@ class Table {
...
 
 protected:
     /* Return the actual key name as a comibation of tableName:key */
     std::string getKeyName(std::string key);
-
```
eladraz: This should be a new patch

common/select.cpp
stcheng: for my point of view, i don't think the #define in #define out #define inout is necessary in the code. the type itself indicates that.

common/consumertable.cpp
hrachya-mellanox: It really depends only OS as well. In past I had issues with this. I checked, it seems for Linux its ok.

common/select.cpp
hrachya-mellanox: You can probably assume that Selectable **c should be out parameter, but int *fd wouldn't tell at all.
I don't think it is wierd, but usefull making declarations obviously clear, and self documenting.
Usually OS APIs, and other low level code have such annotations, especially when there is no documentation.
SAI has this as well.
In Windows I believe there are static analysis tools which will treat absence of in/out as error. I understand this is not Windows, but it has benefits outlined above.

stcheng:
thanks

PR #5, opened 2016-04-12
common: add removeSelectable function in Select class

stcheng:
Used by teamsyncd.

### DIFF
--- a/common/select.cpp
+++ b/common/select.cpp
@@ -1,7 +1,7 @@
-#include <algorithm>
 #include "common/selectable.h"
 #include "common/logger.h"
 #include "common/select.h"
+#include <algorithm>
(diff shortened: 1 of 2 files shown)


common/select.cpp
```diff
@@ -13,6 +13,18 @@ void Select::addSelectable(Selectable *c)
     m_objects.push_back(c);
 }
 
+void Select::removeSelectable(Selectable *c)
+{
+    for (auto i = m_objects.begin(); i != m_objects.end(); i++)
```
kcudnik: use find() on m_objects ?

stcheng: yes. i'll modify this.

kcudnik:
Ack

PR #6, opened 2016-04-13
Add notification consumer and producer

kcudnik:


lguohan:
can we have unit test for this?

kcudnik:
I can add this in next cr

lguohan:
ack

PR #7, opened 2016-04-13
Consumer should populate table, not producer

kcudnik:
Adding alterTable flag which will help to address db freeze scenario:
OA is restarting, but in the same time some APP decided to add/del some entries in app view, then after OA restarts, current app view could be altered (added/chanded/deleted entries) and also the same message is still in the producer queue which is not desired: for example event was to set entry, and OA will read the same operation again and execute it.

Also note: should this implemented behavior should be as default?


lguohan:
check current redis-ut.cpp to see if the test is still passing?

kcudnik:
Yes, test are passing, just fixed lib order to make test compile

lguohan:
ack

PR #8, opened 2016-04-14
Add SelectableEvent class

kcudnik:
Purpose of this class is to allow threads waiting on selectable objects to be notified by user if for example user wants to finish thread, instead of periodic cause timeouts and check if thread needs to end

### DIFF
--- a/tests/redis_ut.cpp
+++ b/tests/redis_ut.cpp
@@ -2,6 +2,7 @@
 #include "common/producertable.h"
 #include "common/consumertable.h"
 #include "common/select.h"
+#include "common/selectableevent.h"
 #include <iostream>
 #include <memory>
 #include <thread>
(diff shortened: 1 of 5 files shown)


lguohan:
can you add a unit test for this, which will also serve as an example for how to use this class. Since it has a new api notify(), not sure how to use it. 

I assume we can no longer use Select::select() since it has the timeout hardcoded.

kcudnik:
I will add unittest for this in next cr, also select() is method of Select class not Selectable, notify function is only provided to trgger select that some message arrived.

lguohan:
I do like to see an example how to use this selectableevent class, especially the notify api. Do you have an example?

tests/redis_ut.cpp
```diff
@@ -221,3 +222,45 @@ TEST(DBConnector, multitable)
...
+
+    s.addSelectable(ev);
+
+    swss::Selectable *sel;
+
+    int fd;
+
+    std::cout << "listening ... " << std::endl;
+
+    int result = s.select(&sel, &fd, 2000);
```
lguohan: still have the timeout? I thought this selectevent can get rid of timeout which is the purpose?

kcudnik: yes, i put timeout explicitly that if something goes wrong, it will not set value, and test will fails, since there is thread.join() and it will block unittest. Purpose of this test is to see if event notification is working correctly and it is.

lguohan:
ack

PR #10, opened 2016-04-19
Add NotificationConsumer/Producer unittest

kcudnik:


common/producertable.cpp
```diff
@@ -57,7 +57,7 @@ void ProducerTable::del(string key, string op)
 
     multi();
 
-    enqueueDbChange(key, "{}", op);
+    enqueueDbChange(key, "{}", "D" + op);
```
lguohan: what's the use case for this passing arbitrary DEL command?

tests/redis_ut.cpp
```diff
@@ -222,6 +224,73 @@ TEST(DBConnector, multitable)
...
+
+    std::vector<swss::FieldValueTuple> values;
+
+    swss::FieldValueTuple e("foo", "bar");
+
+    values.push_back(e);
+
+    std::cout << "sending ntf" << std::endl;
+
+    np.send("a", "b", values);
```
lguohan: what's the difference between the data and FieldValueTuple?

common/producertable.cpp
kcudnik: Added just for flexibility

tests/redis_ut.cpp
kcudnik: I don't get the quesion, data is strign since api takes "key" "op" as strings and values as vector of FieldValues, data is just a name here which was intented as "data" send inside notification

lguohan: void send(std::string op, std::string data, std::vector<FieldValueTuple> &values);

here is the send api for notificationproducer. there is no key. The second arg is data. The question is to explain the difference between the data and values for the API.

kcudnik: Data is just arbitrary string that is passed by user, values is just for convenience, will be used for attribute id/attribute value vector so user will not have to split it manually. Same concept as in consumer and producer

lguohan: maybe we should put your description to notificationproducer.h to explain how to use this API.

lguohan:
ack

PR #12, opened 2016-04-27
Splitting common library into two libraries

stcheng:
one with libnl
one without


kcudnik:
What is purpose of that split?

lguohan:
so that syncd does not need to depends on libnl

lguohan:
ack

stcheng:
Delayed for future evaluation.

kcudnik:
ack

stcheng:
close for now. leave it for future discussion.

PR #13, opened 2016-04-28
Improve performance on table set command

kcudnik:


common/table.cpp
```diff
@@ -73,9 +73,8 @@ void Table::set(std::string key, std::vector<FieldValueTuple> &values,
 {
     /* We are doing transaction for AON (All or nothing) */
     multi();
```
lguohan: mulit is not needed.

kcudnik: removed

lguohan:
ack

PR #20, opened 2016-05-07
table: Adding getTableContent function

stcheng:
This function is used to retrieve the whole content of a table.

It will future be used for orchagent restarting logics.

### DIFF
--- a/common/table.cpp
+++ b/common/table.cpp
@@ -126,6 +126,38 @@ void Table::delField(std::string key, std::string field)
                            "DEL operation failed");
 }
 
+void Table::getTableContent(std::vector<KeyOpFieldsValuesTuple> &tuples)
+{
+    std::vector<std::string> keys;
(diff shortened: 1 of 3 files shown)


common/table.cpp
```diff
@@ -126,6 +126,43 @@ void Table::delField(std::string key, std::string field)
...
+    {
+        std::vector<FieldValueTuple> values;
+        std::string op = "";
+
+        get(key, values);
+        tuples.push_back(make_tuple(key, op, values));
+    }
+}
+
+bool Table::getTableKeys(std::vector<std::string> &keys)
```
kcudnik: return could be void since if no elements then you will get empty result
you are also not checking output of this function

kcudnik:
ack

common/table.cpp
stcheng: thanks. fixed.

PR #25, opened 2016-06-17
[swss-common] Introduction of COPP table

hrachya-mellanox:
*Added COPP table name to scheme.h

Signed-off-by: hrachya@mellanox.com


stcheng:
Already checked in.

PR #26, opened 2016-06-21
netlink: Setting nl_socket buffer size to 256KB

stcheng:


eladraz:
Acked-by: Elad Raz <eladr@mellanox.com>

PR #30, opened 2016-07-28
[common] - Added PFC related table names

hrachya-mellanox:
* Added table names for PFC mappings

signed-off-by: hrachya@mellanox.com


lguohan:
Ack

PR #34, opened 2016-08-29
producertable: Adding dump file to record all producer activities to file

stcheng:
- Use json.hpp to construct the JSON data and store into the m_dumpFile,
    which could be used as input file for swssconfig.
- Add unit test for verification


kcudnik:
Ack

PR #36, opened 2016-09-08
IpPrefix supports ipv6, add unit test

qiluo-msft:


common/Makefile.am
```diff
@@ -1,5 +1,6 @@
 INCLUDES = -I $(top_srcdir)
 
+bin_PROGRAMS = unittest
```
stcheng: move to test/ folder.

common/ipaddress.h
```diff
@@ -59,11 +59,6 @@ class IpAddress
                );
     }
 
-    inline bool operator!=(const IpAddress &o) const
```
stcheng: why remove not equal?

common/ipprefix.h
```diff
@@ -24,7 +25,38 @@ class IpPrefix
...
-        return IpAddress(htonl((0xFFFFFFFFLL << (32 - m_mask)) & 0xFFFFFFFF));
+        switch (m_ip.getIp().family)
+        {
+            case AF_INET:
+            {
+                return IpAddress(htonl((0xFFFFFFFFLL << (32 - m_mask)) & 0xFFFFFFFF));
+            }
+            case AF_INET6:
+            {
+                ip_addr_t ipa;
```
stcheng: Use IpAddress class?

common/ipaddress.h
qiluo-msft: Not necessary for interacting with standard libraries.

common/ipprefix.h
qiluo-msft: It's more convenient to operate on underlying array directly.

stcheng:
Thanks

PR #39, opened 2016-09-20
logger: Preventing flooding same message multiple times

stcheng:


lguohan:
this kind of intelligent should be put in higher logic, should not be here.

stcheng:
agree

PR #43, opened 2016-10-04
Fix getMask() buffer overflow

qiluo-msft:


common/ipprefix.h
```diff
@@ -26,10 +26,14 @@ class IpPrefix
 
     inline IpAddress getMask() const
     {
+        int mask = m_mask < 0 ? 0 : m_mask;
```
lguohan: it is better to change the constructor to validate the m_mask there, instead of here. Here we should assume the m_mask is in correct range.

common/ipprefix.h
```diff
@@ -26,10 +26,14 @@ class IpPrefix
...
+        
         switch (m_ip.getIp().family)
         {
             case AF_INET:
             {
+                if (mask > 32)
```
lguohan: validation in the constructor.

common/ipprefix.h
```diff
@@ -38,18 +42,23 @@ class IpPrefix
...
+                // mask / 8 - 1 <= n < mask / 8
+                if (mask > 128)
+                    mask = 128;
+                int mid = (mask + 7) / 8 - 1;
+                if (mid >= 0)
+                {
+                    int leftbit = (mask + 7) / 8 * 8 - mask;
+                    ipa.ip_addr.ipv6_addr[mid] = 0xFF << leftbit;
+                    memset(ipa.ip_addr.ipv6_addr, 0xFF, mid);
+                }
```
lguohan: mid = mask >> 3;
bits = mask & 0x7
ipa.ip_addr.ipv6_addr[mid] = 0xFF << (8 - bits)
if (mid > 0)
{
  memset(ipa.ip_addr.ipv6_addr, 0xFF, mid);
}
memset(ipa.ip_addr.ipv6_addr + mid + 1, 0, 16 - mid - 1);

common/ipprefix.h
qiluo-msft: Resolved.

common/ipprefix.h
qiluo-msft: Resolved.

common/ipprefix.h
qiluo-msft: Rewrite and tune for the boundary cases.

stcheng:
thanks for fixing

PR #44, opened 2016-10-07
add isZero() for IpAddress and isDefaultRoute() for IpPrefix

lguohan:


common/ipaddress.h
```diff
@@ -29,6 +29,14 @@ class IpAddress
         return m_ip.family == AF_INET;
     }
 
+    inline bool isZero() const
+    {
+        unsigned char zerobuf[16] = { 0 };
```
qiluo-msft: static unsigned char: to prevent initialization each time. alternatively, compared to 0 in a loop without a buffer.
If use 'static', then the function should be 'static inline'

PR #46, opened 2016-10-18
Switch Consumer and Producer to use lua script in redis

kcudnik:


common/consumertable.cpp
```diff
@@ -51,77 +51,110 @@ ConsumerTable::~ConsumerTable()
...
+            &temp,
+            "EVALSHA %s 4 %s %s %s %s '' '' '' ''",
+            sha.c_str(),
+            getKeyQueueTableName().c_str(),
+            getOpQueueTableName().c_str(),
+            getValueQueueTableName().c_str(),
+            getTableName().c_str());
+
+    string command = string(temp, len);
+    free(temp);
```
qiluo-msft: if redisFormatCommand() failed, len gets -1, and temp is unitialized. ref: https://github.com/redis/hiredis/blob/5f98e1d35dcf00a026793ada2662f6e1ba77eb17/hiredis.c#L530.

Then string constructor will convert len(-1) to size_t. Free will use an invalid pointer.
Many occurrences in this PR.

common/consumertable.cpp
```diff
@@ -51,77 +51,110 @@ ConsumerTable::~ConsumerTable()
...
-        // HACK, this needs to be done in different way
-        // when get operation is performed we don't
-        // want to put anything to db
-        return;
-    }
-
-    multi();
-
-    if (dbop == DEL_COMMAND)
+    for (size_t i = 2; i < ctx->elements; i += 2)
```
qiluo-msft: Is it possible that i+1==(ctx->elements), and then index out of range?

common/consumertable.cpp
```diff
@@ -51,77 +51,110 @@ ConsumerTable::~ConsumerTable()
...
-    rpop_op += getOpQueueTableName();
-    enqueue(rpop_op, REDIS_REPLY_STRING);
-
-    rpop_value += getValueQueueTableName();
-    enqueue(rpop_value, REDIS_REPLY_STRING);
-
-    exec();
+    static std::string luaScript =
+
+        "local key   = redis.call('RPOP', KEYS[1])\n"
```
qiluo-msft: Where is it defined?

kcudnik: Where is defined what ?

common/consumertable.cpp
kcudnik: returns -1 when first argument is NULL or when malloc failed, first argument is never null, and when malloc will fail, then we have bigger problem

qiluo-msft: Yes, I am talking about malloc failure. It that happen, we should die decently instead of continue run:
1. string(temp, -1)
2. free(random value)

common/consumertable.cpp
kcudnik: keys is internal redis variable in lua scripts

common/consumertable.cpp
kcudnik: if malloc is not working then we wont be able even to alloc memory for exception to throw, only suitable action to make is abort in this case, there are also some bits that control this via mallopt() which approach you propose to choose ? throw exception ? abort or set bits via mallopt ?

qiluo-msft: 1. Throw exception is ok. The exception object could be created on stack.
2. Or, If this function is possible to fail by design, maybe change the type of return value, indicating the status.
Not familiar with mallopt(). The standard behavior of malloc() is good enough.

common/consumertable.cpp
```diff
@@ -51,77 +51,110 @@ ConsumerTable::~ConsumerTable()
...
+        "    if i == nil then break end\n"
+        "    j = string.find(json, '\"', i + 1)\n"
+        "    local field = descape(string.sub(json, i + 1 , j))\n"
+        "    i = string.find(json, '\"', j + 1)\n"
+        "    j = string.find(json, '\"', i + 1)\n"
+        "    local value = descape(string.sub(json, i + 1, j))\n"
+        "    i = j\n"
+        "    table.insert(ret, field)\n"
+        "    table.insert(ret, value)\n"
+        "end\n"
```
kcudnik: i will do that in next pr

common/consumertable.cpp
kcudnik: Fixed

qiluo-msft:
Please also check with other reviewers.

stcheng:
it breaks the unit test. please create another one and fix the issue.

PR #50, opened 2016-10-22
Make produce and consumer use LUA script

kcudnik:
+ update compiler warnings and fix unittest


configure.ac
```diff
@@ -33,7 +33,50 @@ AC_ARG_ENABLE(gtest,
...
+CFLAGS_COMMON+=" -fPIC"
+CFLAGS_COMMON+=" -std=c++11"
+CFLAGS_COMMON+=" -Wall"
+CFLAGS_COMMON+=" -Wcast-align"
+CFLAGS_COMMON+=" -Wcast-qual"
+CFLAGS_COMMON+=" -Wconversion"
+CFLAGS_COMMON+=" -Wdisabled-optimization"
+CFLAGS_COMMON+=" -Werror"
+CFLAGS_COMMON+=" -Wextra"
+CFLAGS_COMMON+=" -Wextra"
```
kcudnik: fixed

common/consumertable.cpp
```diff
@@ -51,77 +52,94 @@ ConsumerTable::~ConsumerTable()
...
+        "    table.insert(ret, k)\n"
+        "    table.insert(ret, v)\n"
+        "end\n"
+
+        "if op == 'get' or op == 'getresponse' then\n"
+        "return ret\n"
+        "end\n"
+
+        "local keyname = KEYS[4] .. ':' .. key\n"
+        "if key == '' then\n"
```
lguohan: how can if key == '' happen?

kcudnik: it can and its used actually :)

lguohan: in what case, when the list is empty?

tests/redis_ut.cpp
```diff
@@ -348,9 +362,8 @@ TEST(Table, test)
     vector<KeyOpFieldsValuesTuple> tuples;
     t.getTableContent(tuples);
 
-    unsigned int size_t = 2;
     cout << "Get total " << tuples.size() << " number of entries" << endl;
-    EXPECT_EQ(tuples.size(), size_t);
+    EXPECT_EQ(tuples.size(), 2);
```
stcheng: To use unsigned int size_t = 2 to suppress the warning: 
../googletest/googletest/include/gtest/gtest.h:1392:11: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
   if (lhs == rhs) {

stcheng: btw, better not to use size_t name as size_t is a keyword in c++

common/consumertable.cpp
lguohan: when op is set, the key is not empty string.

PR #54, opened 2016-10-27
Use json array in consumer-producer to save fields order

kcudnik:


common/json.cpp
```diff
@@ -10,11 +10,13 @@ namespace swss {
 
 string JSon::buildJson(const vector<FieldValueTuple> &fv)
 {
-    nlohmann::json j;
+    nlohmann::json j = nlohmann::json::array();
 
+    // we use array to save order
```
qiluo-msft: How about json array of key-value pair?

kcudnik: i explicitly didn't used that to reduce overhead

PR #58, opened 2016-11-03
Add Table dump method

kcudnik:


lguohan:
what is purpose of this dump? is it for user debugging purpose or something else?

kcudnik:
It can be used in 2 places, 1 in syncd during restart, now its 2 step process - get keys and then query each key separetly, this is doing everything in one shot, and second usage im writing asicdump tool that will dump this table to stdout for debugging purpose to see current state, since navigating via redis-cli is hard to list all attributes in all objects.

qiluo-msft:
@kcudnik Is the first place an existing feature, ie. are you just refactoring?
If the purpose is only debug, we can just write lua script file in database container.

kcudnik:
I have already tool asicdump which dumps output from db using this method, i was holding this for another pr, and i was planning to reuse that also in syncd like mentioning in previous comment

common/table.cpp
```diff
@@ -309,4 +311,85 @@ std::string Table::scriptLoad(const std::string& script)
...
+        "local keys = redis.call(\"keys\", KEYS[1] .. \":*\")\n"
+        "local res = {}\n"
+
+        "for i,k in pairs(keys) do\n"
+        "   local skeys = redis.call(\"HKEYS\", k)\n"
+        "   local sres={}\n"
+
+        "   for j,sk in pairs(skeys) do\n"
+        "       sres[sk] = redis.call(\"HGET\", k, sk)\n"
+        "   end\n"
```
lguohan: it probably does not make much performance difference here, but you can also use HGETALL to construct hash

local hgetall = function (key)
  local bulk = redis.call('HGETALL', key)
	local result = {}
	local nextkey
	for i, v in ipairs(bulk) do
		if i % 2 == 1 then
			nextkey = v
		else
			result[nextkey] = v
		end
	end
	return result
end

PR #59, opened 2016-11-04
Fix duplicated notifications bug

kcudnik:
In previous implementation on high volume of notifications on the stream messages were swallowed in readCache method, and previous "m_msg" was not populated, and since readCache returned DATA result, readMe method (which populates m_msg) was not called which was causing duplicated notifications poped. Now queue was introduced and populated every time we get valid redis reply, and pop is accessing queue directly.

### DIFF
--- a/common/notificationconsumer.cpp
+++ b/common/notificationconsumer.cpp
@@ -7,8 +7,8 @@
 #define REDIS_PUBLISH_MESSAGE_ELEMNTS (3)
 
 swss::NotificationConsumer::NotificationConsumer(swss::DBConnector *db, std::string channel):
-    m_db(db), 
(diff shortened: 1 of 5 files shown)


qiluo-msft:
Check and fix the similar bug in https://github.com/Azure/sonic-swss-common/blob/master/common/consumertable.cpp?

kcudnik:
There is no bug there since table consumer is not reading notification stream, but data is read from queue, in other words this consumer only cares "is i have been notified"

lguohan:
can you provide some summary for this bug in the commit message. For example, previously there is no queue in the notification stream, new message will overwrite old message so that the consumer get duplicated message. This patch introduces queue so that the consumer get message one by one.

PR #65, opened 2016-11-08
Add operation counters for ConsumerStateTable/ProducerStateTable

qiluo-msft:


common/producerstatetable.cpp
```diff
@@ -53,21 +56,22 @@ void ProducerStateTable::del(std::string key, std::string op /*= DEL_COMMAND*/)
     static std::string luaScript =
         "redis.call('SADD', KEYS[2], ARGV[2])\n"
         "redis.call('DEL', KEYS[3])\n"
+        "redis.call('HINCRBY', KEYS[4], 'del', 1)\n"
```
lguohan: this is a critical path, instead of using lua script, I think increase the counter in the library and periodically put it back to db has better performance.

PR #68, opened 2016-11-18
Add prefix feature to Consumer/Producer table

kcudnik:


tests/redis_ut.cpp
```diff
@@ -398,3 +398,32 @@ TEST(Table, test)
...
+
+    DBConnector db(TEST_VIEW, "localhost", 6379, 0);
+    ProducerTable p(&db, tableName);
+
+    std::vector<FieldValueTuple> values;
+
+    FieldValueTuple t("f", "v");
+    values.push_back(t);
+
+    p.set("key", values, "op", "prefix_");
```
qiluo-msft: Why this feature is needed? Does it equal to:

    ProducerTable p(&db, "prefix_" + tableName);

without code changes.

kcudnik: no, idea is that we use the same communication channel for different "prefixed" tables, since we want to keep order of function calls, looks ugly :( opened to other solutions

qiluo-msft: OK, I understand the need.

Is it achievable by prefixing key before calling set()/del(), and trimming after calling pop()?

kcudnik: problem here is channel, which is "TableName" and all other queues op/value/key derive from that tablename, also channel derives that which we want to keep the same

lguohan: can we do something like

producerTable p(&db, tableName, instanceId);

p.set("key", values, "op");

ConsumerTable c(&db, tableName, instanaceId);
c.pop(kco);

I suggest to bring the instance concept to the table, so that a table can have multiple instances and they all share the the same channel.

for example, ASIC_TABLE instance 0 is ASIC_TABLE_0, instance 1 is ASIC_TABLE_1. By default, if instance is not specificed, it is instance 0.

kcudnik: no we can't sence that will be on different connection and it will need to be on selectable and 2 selectable events will be required

PR #69, opened 2016-11-23
IpPrefix and MacAddress classes extension.

oleksandrivantsiv:
- Added define for mirror table.
- Extended IpPrefix class.
- Extended MacAddress class.


common/macaddress.h
```diff
@@ -37,6 +37,11 @@ class MacAddress
         return !(*this == o);
     }
 
+    inline bool operator<(const MacAddress &o)
+    {
+        return to_uint64() < o.to_uint64();
```
qiluo-msft: This function is not necessary. You may use memcpy.

common/ipaddress.h
```diff
@@ -72,6 +72,11 @@ class IpAddress
                );
     }
 
+    inline bool operator!=(const IpAddress &o) const
```
qiluo-msft: This function is not necessary. You may use std::rel_ops: http://en.cppreference.com/w/cpp/utility/rel_ops/operator_cmp

common/ipprefix.h
```diff
@@ -66,6 +66,43 @@ class IpPrefix
         return (m_mask == 0);
     }
 
+    inline bool isAddressInSubnet(const IpAddress& addr) const
```
qiluo-msft: Is it possible to move code to https://github.com/Azure/sonic-swss/blob/master/orchagent/swssnet.h, and reuse code there? Also add unit test for fundamental functions.

stcheng: i'm okay with putting this function here as it is more related to ipprefix class which sits here. the file swssnet.h contains more about ip address types defined in sai.

stcheng: also unit test is needed.

common/ipaddress.h
oleksandrivantsiv: I saw that in other classes operators are implemented in the same way. https://github.com/Azure/sonic-swss-common/blob/master/common/ipaddresses.h#L38W
Will remove it.

common/ipprefix.h
oleksandrivantsiv: Unit test is added

lguohan:
unit test fails

oleksandrivantsiv:
It is strange. I've run this test continuously  for 100 times and didn't saw a failure.
Is there a possibility to re-run tests?

lguohan:
https://sonic-jenkins.westus.cloudapp.azure.com/job/common/job/swss-common-pr-build/31/console

```
ipprefix_ut.cpp:85: Failure
Value of: prefix3.isAddressInSubnet(ip3)
  Actual: false
Expected: true
ipprefix_ut.cpp:86: Failure
Value of: prefix4.isAddressInSubnet(ip3)
  Actual: true
Expected: false
```

tests/ipprefix_ut.cpp
```diff
@@ -63,3 +63,34 @@ TEST(IpPrefix, compare)
     EXPECT_TRUE(!(ip2 < ip1));
     EXPECT_TRUE(!(ip1 == ip2));
 }
+
+TEST(IpPrefix, subnet)
+{
+    IpPrefix prefix1("0.0.0.0/0");
```
qiluo-msft: >0 [](start = 30, length = 1)

add test cases for ipv4/32, and ipv6/128?

oleksandrivantsiv: Added

oleksandrivantsiv:
Found and fixed issue in implementation of isAddressInSubnet for IPv6 address

PR #72, opened 2016-12-02
Fix #71: ProducerStateTable does not support field/value names with space characters

### LINKED ISSUE #71: ProducerStateTable does not support field/value names with space characters

qiluo-msft:
Fix #71: ProducerStateTable does not support field/value names with space characters

### DIFF
--- a/common/producerstatetable.cpp
+++ b/common/producerstatetable.cpp
@@ -1,10 +1,13 @@
 #include <stdlib.h>
 #include <tuple>
 #include <sstream>
+#include <algorithm>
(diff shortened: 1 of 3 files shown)


common/producerstatetable.cpp
```diff
@@ -20,31 +23,38 @@ void ProducerStateTable::set(std::string key, std::vector<FieldValueTuple> &valu
     static std::string luaScript =
```
stcheng: remove std:: as it is already included

PR #74, opened 2016-12-06
Implement buffered pop for ConsumerTable/ConsumerStateTable

qiluo-msft:
No change on interface.

### DIFF
--- a/common/consumertable.cpp
+++ b/common/consumertable.cpp
@@ -17,6 +17,7 @@ namespace swss {
 ConsumerTable::ConsumerTable(DBConnector *db, string tableName)
     : RedisTransactioner(db)
     , TableName_KeyValueOpQueues(tableName)
+    , POP_BATCH_SIZE(128)
 {
     for (;;)
     {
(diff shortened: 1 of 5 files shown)


common/consumertable.cpp
```diff
@@ -35,81 +36,126 @@ ConsumerTable::ConsumerTable(DBConnector *db, string tableName)
...
-        "       redis.call('HSET', keyname, ret[st], ret[st+1])\n"
-        "       st = st + 2\n"
-        "   end\n"
+void ConsumerTable::pops(deque<KeyOpFieldsValuesTuple> &vkco, string prefix)
+{
+    static string luaScript =
+        "local rets = {}\n"
+        "local keys   = redis.call('LRANGE', KEYS[1], -ARGV[1], -1)\n"
+        "local ops    = redis.call('LRANGE', KEYS[2], -ARGV[1], -1)\n"
+        "local values = redis.call('LRANGE', KEYS[3], -ARGV[1], -1)\n"
```
lguohan: ARGV[1] is 128, so we start from -128?

LRANGE key start stop

common/consumertable.cpp
```diff
@@ -17,6 +17,7 @@ namespace swss {
 ConsumerTable::ConsumerTable(DBConnector *db, string tableName)
     : RedisTransactioner(db)
     , TableName_KeyValueOpQueues(tableName)
+    , POP_BATCH_SIZE(128)
```
lguohan: >POP_BATCH_SIZE(128) [](start = 6, length = 19)

increase the default batch size to 1k?

lguohan: what is the reason to have the batch size? why not popall?

---
In reply to: [91067133](https://github.com/Azure/sonic-swss-common/pull/74#discussion_r91067133) [](ancestors = 91067133)

qiluo-msft: It's a tradeoff between throughput and latency. Will tune together with producer in an end-to-end test.

---
In reply to: [91067133](https://github.com/Azure/sonic-swss-common/pull/74#discussion_r91067133) [](ancestors = 91067133)

common/consumertable.cpp
```diff
@@ -35,81 +36,126 @@ ConsumerTable::ConsumerTable(DBConnector *db, string tableName)
 
 void ConsumerTable::pop(KeyOpFieldsValuesTuple &kco, string prefix)
```
kcudnik: you are changing consumer table as well, not sure about this if you will not break anything here, add some unittest for that ?

qiluo-msft: I am using the unit tests in redis_ut.cpp. And also did an integration test in a lab switch.

PR #77, opened 2016-12-21
ProducerTable/ProducerStateTable add buffered mode

qiluo-msft:
In buffered mode, all the "set" and "del" operations will be accumulated into hiredis output buffer until buffer overflow, or flush(), or other operations. This optimization is also called redis pipeline, reducing the chance of block waiting for redis replies.

Also original non buffered behavior is kept to maintain backward compatibility.


lguohan:
can you investigate why the pr build fails?

qiluo-msft:
Build resolved.

common/producertable.cpp
```diff
@@ -78,6 +95,11 @@ void ProducerTable::set(string key, vector<FieldValueTuple> &values, string op,
     }
 
     enqueueDbChange(key, JSon::buildJson(values), "S" + op, prefix);
+    // Only buffer continuous "set/set" or "del" operations
+    if (!m_buffered || op != "set")
+    {
+        m_pipe->flush();
+    }
```
lguohan: why cannot we mix the set and del operations?

lguohan:
:shipit:

common/producertable.cpp
qiluo-msft: Actually yes. We can mix "set/set" and "del" continuously in the pipeline without flushing. Others such as "set/getResponse" will trigger the flushing.

---
In reply to: [93956717](https://github.com/Azure/sonic-swss-common/pull/77#discussion_r93956717) [](ancestors = 93956717)

common/producertable.cpp
```diff
@@ -12,10 +12,25 @@ using json = nlohmann::json;
 
 namespace swss {
 
-ProducerTable::ProducerTable(DBConnector *db, string tableName)
+ProducerTable::ProducerTable(DBConnector *db, std::string tableName)
+    : ProducerTable(new RedisPipeline(db, 1), tableName, false)
```
kcudnik: what is "1" here?

common/producerstatetable.cpp
```diff
@@ -5,34 +5,60 @@
...
-void ProducerStateTable::set(string key, vector<FieldValueTuple> &values,
-                 string op /*= SET_COMMAND*/, string prefix)
+ProducerStateTable::ProducerStateTable(RedisPipeline *pipeline, string tableName, bool buffered)
+    : TableName_KeySet(tableName)
+    , m_buffered(buffered)
+    , m_pipeowned(false)
+    , m_pipe(pipeline)
 {
-    static string luaScript =
+    string luaSet =
```
kcudnik: static is needed here to not create string each time

common/producertable.cpp
```diff
@@ -12,10 +12,25 @@ using json = nlohmann::json;
...
+}
+
+ProducerTable::ProducerTable(RedisPipeline *pipeline, std::string tableName, bool buffered)
     : TableName_KeyValueOpQueues(tableName)
-    , m_db(db)
+    , m_buffered(buffered)
+    , m_pipeowned(false)
+    , m_pipe(pipeline)
 {
+    string luaEnque =
```
kcudnik: this can be static

common/producertable.cpp
qiluo-msft: It is the pipeline size, ie. it can hold 1 message, ie. each new request will trigger pipeline flushing. It is conservative to maintain the behavior of previous implementation. All the caller will switch to new constructer if the pipeline is tested.

---
In reply to: [94517331](https://github.com/Azure/sonic-swss-common/pull/77#discussion_r94517331) [](ancestors = 94517331)

common/producerstatetable.cpp
qiluo-msft: Ideally each Redis server will load the script once at most. It is not easy for this class to differentiate multiple servers, so loose it to once per ProducerStateTable. It will make no harm to load existing script, and the running time impact can be ignored in our usage. I could optimize loadRedisScript() to achieve the ideal if you prefer.

---
In reply to: [94517595](https://github.com/Azure/sonic-swss-common/pull/77#discussion_r94517595) [](ancestors = 94517595)

common/producertable.cpp
qiluo-msft: Replied under another comment.

---
In reply to: [94517782](https://github.com/Azure/sonic-swss-common/pull/77#discussion_r94517782) [](ancestors = 94517782)

PR #81, opened 2017-01-20
[logger]: Enable loglevel setting via redis DB

marian-pritsak:
Extend schema
Add registration functionality
Add cli tool swssloglevel


lguohan:
@qiluo-msft, can you take a look at this?

common/logger.h
```diff
@@ -73,14 +80,19 @@ class Logger
             std::chrono::time_point<std::chrono::high_resolution_clock> m_start;
     };
 
+    void setMinPrio(Priority prio);
+    Priority getMinPrio();
```
qiluo-msft: Why make them non-static?

common/logger.cpp
```diff
@@ -1,33 +1,78 @@
...
-Logger::Logger()
+void Logger::linkToDb(std::string dbName, PriorityChangeObserver *priorityChangeObservers)
 {
-    std::ifstream self("/proc/self/cmdline", std::ios::binary);
+    auto& logger = getInstance();
 
-    m_self = "unknown";
+    auto swssNotify = [ &logger ](std::string component, std::string prioStr)
+    {
+        logger.m_minPrio = priorityStringMap.at(prioStr);
```
qiluo-msft: Handle invalid data from redis?

common/logger.cpp
```diff
@@ -1,33 +1,78 @@
...
+};
 
-Logger::Logger()
+void Logger::linkToDb(std::string dbName, PriorityChangeObserver *priorityChangeObservers)
 {
-    std::ifstream self("/proc/self/cmdline", std::ios::binary);
+    auto& logger = getInstance();
 
-    m_self = "unknown";
+    auto swssNotify = [ &logger ](std::string component, std::string prioStr)
```
qiluo-msft: It captures logger by reference, and logger is a local variable. If the lambda is invoked out of this scope (expected for this case), undefined behavior.

common/logger.h
```diff
@@ -73,14 +80,19 @@ class Logger
...
 private:
-    Logger();
+    Logger(){};
```
qiluo-msft: Why remove original implementation?

common/logger.cpp
```diff
@@ -1,33 +1,78 @@
...
+        {
+            SWSS_LOG_NOTICE("Invalid loglevel entry in DB for %s", key.c_str());
+            continue;
+        }
+
+        logger.m_currentPrios[i.first] = *newPrio;
+        i.second(i.first, *newPrio);
     }
+
+    logger.m_prioThread.reset(new std::thread(&Logger::prioThread, &logger));
```
qiluo-msft: If there is an old thread running, stop it?

common/logger.h
marian-pritsak: Instance of logger is already static
If we have all members outside it, there's no reason for getInstance()

common/logger.cpp
marian-pritsak: logger itself is a reference to static variable from getInstance(), so lambda captures reference to this static instance

common/logger.h
marian-pritsak: Original implementation was getting name for m_self from /proc and it was buggy
In case process was started with some arguments, they were part of m_self too
IMO specifying things like name explicitly is easier to read and understand

common/logger.cpp
marian-pritsak: It will be stopped on destructor call
However, this function should be called only once
Do you think assertion is needed to prevent multiple calls?

marian-pritsak:
This will allow:
1. a permanent loglevel value in Redis DB
2. unified way for setting loglevel for all swss-common clients
3. enable loglevel setting for SAI APIs

common/logger.cpp
qiluo-msft: Could you provide a proof for "It will be stopped on destructor call"? There is no mechanism to guarantee " this function should be called only once". Assertion has no runtime effect here.

marian-pritsak: Regarding thread stop, http://en.cppreference.com/w/cpp/thread/thread/~thread
By assertion I mean throwing runtime error in case linkToDb is called more than once

common/logger.cpp
marian-pritsak: Done

common/logger.cpp
qiluo-msft: This is cutting edge of C++ and different compilers may not have the compliant implementation currently. How about making logger as lamda's parameter?

common/logger.cpp
qiluo-msft: Throwing may be an option. You can also support multiple call. I notice that prioThread will never exit, and this is a problem for either case.

common/logger.cpp
oleksandrivantsiv: Is there any plans to use compiler other than GCC?
Any way all C+11 feature are supported by most of all popular compilers:
http://en.cppreference.com/w/cpp/compiler_support

common/logger.cpp
marian-pritsak: There's no reason for multiple calls
Its a one time action executed at program startup
I've added throwing runtime error in case it is called more than once

qiluo-msft: I am ok that you decide not to support multiple calls. Still prioThread will never exit, does it make the logger dtor hangs?

common/logger.cpp
qiluo-msft: I didn't say that there is any plan to use other compilers. I didn't say that your implementation is wrong in specific compiler. I prefer some easier syntax to achieve the same goal without cutting edge technology in C++ standard and compiler implementations.

To be more specific, I propose codes like below.

```c++
static void Logger::Notify(Logger& logger, std::string component, Priority prio)
{
    self >> m_self;
    logger.m_minPrio = prio;
}

for (const auto& i : logger.m_priorityChangeObservers)
{
    ...
    logger.m_currentPrios[i.first] = priorityStringMap.at(*newPrio);
    i.second(logger, i.first, priorityStringMap.at(*newPrio));
}
```

qiluo-msft:
I marked some comments as closed and reply the some open ones.

common/logger.cpp
marian-pritsak: It won't hang, std::terminate() will be called for thread on destructor call

marian-pritsak:
@lguohan @qiluo-msft
Did I miss any comment? Or it can be merged?
Also notice that there are pull requests in swss and sairedis that depend on this one.

qiluo-msft:
Open issues:
https://github.com/Azure/sonic-swss-common/pull/81#discussion_r98978653
https://github.com/Azure/sonic-swss-common/pull/81#discussion_r97464012

common/logger.h
qiluo-msft: The modification will break existing code using it.

---
In reply to: [97550682](https://github.com/Azure/sonic-swss-common/pull/81#discussion_r97550682) [](ancestors = 97550682)

common/logger.cpp
marian-pritsak: Done

marian-pritsak:
Fixed comments

common/logger.cpp
qiluo-msft: Thanks!

PR #86, opened 2017-04-26
Add declaration of unix socket

kcudnik:


lguohan:
can you update the submodule in buildimage?

PR #88, opened 2017-04-28
add debug symbol package

lguohan:


debian/control
```diff
@@ -11,6 +11,12 @@ Depends: ${shlibs:Depends}, ${misc:Pre-Depends}
 Section: libs
 Description: Tihs package contains Switch State Service common library.
```
jleveque: Could you fix this unrelated comment while you're at it? "Tihs" -> "This"

PR #89, opened 2017-05-02
Add support for bulkset in consumer/producer

kcudnik:


common/consumertable.cpp
```diff
@@ -84,7 +84,30 @@ void ConsumerTable::pops(deque<KeyOpFieldsValuesTuple> &vkco, string prefix)
...
+            "    key = key:sub(1, string.find(key, ':') - 1)\n"
+
+            "   local len = #ret\n"
+            "   local st = 3\n"         // since 1 and 2 is key/op
+            "   while st <= len do\n"
+            "       local field = ret[st]\n"
+            // keyname is ASIC_STATE : OBJECT_TYPE : OBJECT_ID
+            "       local keyname = KEYS[4] .. ':' .. key .. ':' .. field\n"
+
+            // value can be multiple a=v|a=v|... we need to split using gmatch
```
lguohan: in the saidump, you use | as a separator, should we use another separator here?

kcudnik: saidump is not using "|" separators, did you mean saiplayer?

lguohan: yes

kcudnik: but we use the same separator here also look at gmatch

Issue #91, opened 2017-05-05
parseMacString() has issues


pavel-shirshov:
parseMacString() considers 52:54:00:25::E9 and 52:54:00:25:E9 as good mac addresses

kcudnik:
Solled in #92

pavel-shirshov:
It wasn't. It's a separate issue.

pavel-shirshov:
Fixed by https://github.com/Azure/sonic-swss-common/pull/171

PR #95, opened 2017-05-10
Handle errors in parseMacString

qiluo-msft:
Solve #91


stcheng:
@pavel-shirshov i think it is okay right now. without this fix the orchagent also cannot work properly.

qiluo-msft:
Created an issue to track this: https://github.com/Azure/sonic-swss/issues/213

PR #96, opened 2017-05-12
[tokenize]: Add tokenizeFirst function

stcheng:
This function is helpful when trying to tokenize a key-value string
that the tokenizer used between key and value also appears in value.

i.e. "neigh:00:00:00:00:00:00" and "NEIGH_TABLE:lo:fc00::79"

### DIFF
--- a/common/tokenize.cpp
+++ b/common/tokenize.cpp
@@ -15,5 +15,23 @@ vector<string> tokenize(const string &str, const char token)
 
     return ret;
 }
+vector<string> tokenize(const string &str, const char token, const size_t firstN)
+{
+    vector<string> ret;
+    string tmp = str;
+    size_t i = 0;
+    auto pos = tmp.find(token);
+
+    while (pos != string::npos && i++ < firstN)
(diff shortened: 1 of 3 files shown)


common/tokenize.cpp
```diff
@@ -16,4 +16,22 @@ vector<string> tokenize(const string &str, const char token)
     return ret;
 }
 
-}
\ No newline at end of file
+vector<string> tokenizeFirst(const string &str, const char token)
```
kcudnik: please add more generic function that takes n as param with default -1 (all)
and also please add overload that takes token as string not only single char that will be more useful

stcheng: so you're suggestion to re-use the tokenize() function with an extra parameter to indicate on what position to split the string into two? or to split the string into n sub-strings? are there any examples, or specific use cases on your side?

stcheng:
@kcudnik updated the pull request. currently i don't have use cases to split the string using another string. i could add it later.

PR #98, opened 2017-05-19
Convert log timer to printf like and use notice level

kcudnik:


stcheng:
could you provide a sample log output?

kcudnik:
: syncdApplyView: apply took 2.490494 sec

kcudnik:
test this please

PR #99, opened 2017-05-28
Extract lua scripts to separate files

kcudnik:


kcudnik:
test this please

kcudnik:
test this please

kcudnik:
test this please

common/redisapi.h
```diff
@@ -29,4 +32,43 @@ static inline std::string encodeLuaArgument(const std::string& arg)
...
+}
+
+static inline std::string loadLuaScript(const std::string& path)
+{
+    SWSS_LOG_ENTER();
+
+    std::vector<std::string> paths = {
+        path,
+        "/usr/share/swss/" + path,
+        "/usr/local/share/swss/" + path
```
qiluo-msft: I know this paths may be handy to you. Is it over design? Is there a user case?

common/redisapi.h
```diff
@@ -29,4 +32,43 @@ static inline std::string encodeLuaArgument(const std::string& arg)
...
+    if (ifs.good())
+    {
+        return std::string((std::istreambuf_iterator<char>(ifs)),
+                (std::istreambuf_iterator<char>()));
+    }
+
+    SWSS_LOG_THROW("failed to read file: '%s': %s", path.c_str(), strerror(errno));
+}
+
+static inline std::string loadLuaScript(const std::string& path)
```
qiluo-msft: >loadLuaScript [](start = 26, length = 13)

Is it possible to reuse loadRedisScript() in this header?

common/redisapi.h
kcudnik: user can sepcify relative or absolute path as needed

common/redisapi.h
kcudnik: not sure what you have in mind, those are 2 different functions

kcudnik:
test this please

kcudnik:
test this please

common/redisapi.h
qiluo-msft: Could you give some detail about "user decide to install those scripts in different locations"? If this is the case, is it possible to get the path programmatically?

---
In reply to: [118832264](https://github.com/Azure/sonic-swss-common/pull/99#discussion_r118832264) [](ancestors = 118832264)

common/redisapi.h
kcudnik: this one is helper since its searches for specific paths where scripts are installed, generic functio is named readTextFile

common/redisapi.h
kcudnik: If you configure swss-common with –prefix=some_non_standard_path then this will guarantee that this path will be checked for those scripts

qiluo-msft: Offline discussed. There is no user case to customize the path by './configure --prefix=XXX' for sonic switch. I suggest make it only one choice and hard-coded.

---
In reply to: [118989570](https://github.com/Azure/sonic-swss-common/pull/99#discussion_r118989570) [](ancestors = 118989570,118832264)

kcudnik:
test this please

kcudnik:
you merged that without passing test for last checkin!

stcheng:
@kcudnik sorry misread the previous test result. i have just manually verified it.

kcudnik:
ok, btw. why build system is not building this pr ?

PR #104, opened 2017-06-09
[Makefile.am]: Use full file names

marian-pritsak:
GNU Automake does not support wildcards, which makes the build fail on
some versions of Automake with error "No rule to make target *.lua".

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -9,10 +9,7 @@ EXTRA_DIST = \
 
 swssdir = $(datadir)/swss
 
-dist_swss_DATA = \
-    consumer_state_table_pops.lua \
-    consumer_table_pops.lua \
-    table_dump.lua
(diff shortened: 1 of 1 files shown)


common/Makefile.am
```diff
@@ -2,11 +2,17 @@ INCLUDES = -I $(top_srcdir) -I/usr/include/libnl3
...
-EXTRA_DIST = *.lua
+EXTRA_DIST = \
+    consumer_state_table_pops.lua \
+    consumer_table_pops.lua \
+    table_dump.lua
 
 swssdir = $(datadir)/swss
 
-dist_swss_DATA = *.lua
+dist_swss_DATA = \
```
kcudnik: Use extra dist here to not repeat twice

marian-pritsak: Done

PR #106, opened 2017-06-27
Increase netlink buffer. Otherwise we're loosing netlink messages

pavel-shirshov:
When we're add many neighbors to our database, we can lose some netlink messages. In this case our app_db database doesn't have all neighbor entries in comparison with Linux kernel.

This pull request introduce a quick fix. It's better to have two threads for the code. One thread for reading data from the socket, another one for writing to database.

### DIFF
--- a/common/netlink.cpp
+++ b/common/netlink.cpp
@@ -34,7 +34,7 @@ NetLink::NetLink() :
     }
 
     /* Set socket buffer size to 256KB */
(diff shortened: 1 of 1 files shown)


common/netlink.cpp
```diff
@@ -34,7 +34,7 @@ NetLink::NetLink() :
     }
 
     /* Set socket buffer size to 256KB */
```
vemulabalaji: Need to change comment to 8192KB.

PR #108, opened 2017-06-29
[netlink]: Check return code from nl_recvmsgs_default to catch lost messages

pavel-shirshov:
When our library is not fast enough in reading netlink messages we're loosing them.
Report about such cases in the log

### DIFF
--- a/common/netlink.cpp
+++ b/common/netlink.cpp
@@ -87,7 +87,14 @@ int NetLink::readCache()
 
 void NetLink::readMe()
 {
-    nl_recvmsgs_default(m_socket);
+    int err = nl_recvmsgs_default(m_socket);
+    if (err < 0)
+    {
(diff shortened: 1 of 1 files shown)


common/netlink.cpp
```diff
@@ -87,7 +87,14 @@ int NetLink::readCache()
 
 void NetLink::readMe()
 {
-    nl_recvmsgs_default(m_socket);
+    int err = nl_recvmsgs_default(m_socket);
+    if (err < 0)
+    {
+        if (err == -NLE_NOMEM)
+            SWSS_LOG_ERROR("netlink reports out of memory on reading a netlink socket. High possiblity of a lost message");
```
stcheng: possibility :)

stcheng:
Please squash and merge the pull request in the future.

PR #110, opened 2017-08-04
Add tables for buffer config update

andriymoroz-mlnx:


andriymoroz-mlnx:
related change https://github.com/Azure/sonic-swss/pull/278

stcheng:
actually i'm thinking about if we do need many tables (or prefixes) related to buffer?
is it better that all the buffer related tables merged to one buffer table and we have a secondary key appended that could be parsed inside the orchagent.

stcheng:
And is it better for us to freeze the new features on the master branch and merging the features to the SAI v1.0 branch?

PR #115, opened 2017-08-22
Refactor Consumer*Table interface, and optimize popping

qiluo-msft:
Introduced TableConsumable interface and let ProducerTable inherits TableBase interface.

### DIFF
--- a/common/consumerstatetable.cpp
+++ b/common/consumerstatetable.cpp
@@ -37,15 +37,14 @@ void ConsumerStateTable::pop(KeyOpFieldsValuesTuple &kco, std::string prefix)
     if (m_buffer.empty())
     {
         pops(m_buffer, prefix);
-    }
-
-    if (m_buffer.empty())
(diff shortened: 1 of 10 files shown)


lguohan:
why 1.0.1 branch, not 1.0.3 branch?

qiluo-msft:
There is no v1.0.3 branch in this repo. I can rename v1.0.1->v1.0.3 after this pull request.
The reason why using this name at the very beginning is because there is only SAI v1.0.1 implementation then.

stcheng:
could you add more information in the commit about the optimization?

qiluo-msft:
That is straightforward. It makes one less empty check for non-empty popping
```
m_buffer.empty()
```

lguohan:
please also add commit description that you have introduced tableconsumable and move producertable to use tablebase directly.

lguohan:
it should be merged into master, right?

PR #117, opened 2017-08-24
Vlan trunk common function support

JipanYanga:
Common functions update for VLAN trunk feature

### DIFF
--- a/common/gettimestamp.h
+++ b/common/gettimestamp.h
@@ -1,0 +1,12 @@
+#ifndef __GETTIMESTAMP__
+#define __GETTIMESTAMP__
+
+#include <string>
+
+namespace swss {
+
+std::string getTimestamp();
+
+}
+
+#endif /* __GETTIMESTAMP__ */
--- a/common/subscriber.cpp
+++ b/common/subscriber.cpp
@@ -1,0 +1,148 @@
+#include <string>
+#include <limits>
+#include <tuple>
+#include <system_error>
+#include <hiredis/hiredis.h>
+#include "logger.h"
+#include "dbconnector.h"
+#include "table.h"
+#include "selectable.h"
+#include "redisselect.h"
+#include "redisapi.h"
+#include "subscriber.h"
+
+namespace swss {
+
+Subscriber::Subscriber(DBConnector *db, std::string tableName)
+    : RedisTransactioner(db)
+    , TableBase(tableName)
+{
+    m_keyspace = "__keyspace@";
+
+    m_keyspace += std::to_string(db->getDB()) + "__:" + tableName + ":*";
+
+    psubscribe(m_db, m_keyspace);
+    /* Don't read existing data implicitly here */
+    setQueueLength(0);
+}
+
+void Subscriber::pop(KeyOpFieldsValuesTuple &kco, std::string prefix)
+{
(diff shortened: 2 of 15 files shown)


common/gettimestamp.h
```diff
@@ -0,0 +1,12 @@
+#ifndef __GETTIMESTAMP__
```
stcheng: rename to `timestamp.h` and use `#pragma once` for all the header files.

common/subscriber.cpp
```diff
@@ -0,0 +1,148 @@
...
+    m_buffer.pop_front();
+}
+
+/*
+ * handle redis keyspace pmessage notification
+ */
+void Subscriber::pops(std::deque<KeyOpFieldsValuesTuple> &vkco, std::string /*prefix*/)
+{
+    RedisReplyAsync r(getSubscribeDBC());
+    // Keep fetching the reply object until all of them are read
```
stcheng: align all the comment format to use `/**/`.

common/subscriber.cpp
```diff
@@ -0,0 +1,148 @@
...
+        ctx = ctx0->element[2];
+        std::string table_key = ctx->str;
+        size_t found = table_key.find(':');
+        // Return if the format of key is wrong
+        if (found == std::string::npos)
+        {
+            SWSS_LOG_ERROR("invalid keyspace notificaton %s", table_key.c_str());
+            continue;
+        }
+        //strip off  the keyspace db string
```
stcheng: extra space

common/subscriber.cpp
```diff
@@ -0,0 +1,148 @@
...
+        auto ctx = ctx0->element[1];
+        if(m_keyspace.compare(ctx->str))
+        {
+            SWSS_LOG_ERROR("invalid pattern %s returned for pmessage of %s", ctx->str, m_keyspace.c_str());
+            continue;
+        }
+
+        ctx = ctx0->element[2];
+        std::string table_key = ctx->str;
+        size_t found = table_key.find(':');
```
stcheng: the `tokenize` function could be used here to check the size of the returned vector.

common/gettimestamp.h
JipanYanga: Ok.

common/subscriber.cpp
JipanYanga: Ok, will remove it.

common/subscriber.cpp
JipanYanga: Ok, will use tokennize.

common/subscriber.cpp
JipanYanga: Ok, will switch to /* */ for all comments.

JipanYanga:
Thanks for the review!
As suggested,  plan to split it into three new pull request.
<1> exec and timestamp changes.
<2> asynchronous redis subscription channel functions for keyspace notification.
<3> non-blocking netlink socket connection update.
Hope this will make it easier to review .  Also will add more detailed comments for each new PR.
Thanks,
Jipan

stcheng:
@jipanyanga Thanks so much! By splitting the commits, the Git history will make more sense for readability. :)

JipanYanga:
target branch not active.

PR #122, opened 2017-09-13
keyspace subscriber implementaion

JipanYanga:
Since keyspace notification is asynchronous, non-blocking redis db connection has been used.
Each subscriber subscribes to specific key pattern and has it own db connection.
There could be multiple key changes in one redis db socket read.
The regular hget operation still uses blocking redis db connection.

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -5,6 +5,7 @@
 #include <errno.h>
 #include <system_error>
 
+#include "common/logger.h"
 #include "common/dbconnector.h"
 #include "common/redisreply.h"
 
@@ -63,6 +64,24 @@ DBConnector::DBConnector(int db, string unixPath, unsigned int timeout) :
     select(this);
 }
 
+/* No timeout argument, non-blocking DB connection */
+DBConnector::DBConnector(int db, string unixPath) :
+    m_db(db)
+{
+    m_conn = redisConnectUnixNonBlock(unixPath.c_str());
+
+    if (m_conn->err)
+        throw system_error(make_error_code(errc::address_not_available),
+                           "Unable to connect to redis (non block unix-socket)");
+    SWSS_LOG_DEBUG("DBConnector db %d, m_conn->fd %d", db, m_conn->fd);
+    m_blocking = false;
+}
+
+bool DBConnector::isBlocking()
+{
+    return m_blocking;
+}
+
 redisContext *DBConnector::getContext()
 {
     return m_conn;
@@ -84,6 +103,16 @@ DBConnector *DBConnector::newConnector(unsigned int timeout)
         return new DBConnector(getDB(),
                                getContext()->unix_sock.path,
                                timeout);
+}
+
+/* Create Non-blocking DB connection */
+DBConnector *DBConnector::newConnector(void)
+{
+    if (getContext()->connection_type != REDIS_CONN_TCP)
+        return new DBConnector(getDB(),
+                               getContext()->unix_sock.path);
+    else
+        return NULL;
 }
 
 }
--- a/common/dbconnector.h
+++ b/common/dbconnector.h
@@ -22,6 +22,7 @@ public:
      */
     DBConnector(int db, std::string hostname, int port, unsigned int timeout);
     DBConnector(int db, std::string unixPath, unsigned int timeout);
+    DBConnector(int db, std::string unixPath);
 
     ~DBConnector();
 
@@ -33,9 +34,15 @@ public:
     /* Create new context to DB */
     DBConnector *newConnector(unsigned int timeout);
 
+    /*  Create new context to DB with Non-blocking connection */
+    DBConnector *newConnector(void);
+
+    bool isBlocking(void);
+
 private:
     redisContext *m_conn;
     int m_db;
+    bool m_blocking = true;
 };
 
 }
--- a/common/redisreplyasync.h
+++ b/common/redisreplyasync.h
@@ -1,0 +1,53 @@
+#pragma once
+
+#include <hiredis/hiredis.h>
+#include <stdexcept>
+#include "dbconnector.h"
+#include "rediscommand.h"
+
+namespace swss {
+
+class RedisReplyAsync
+{
+public:
+    /*
+     * Send a new command to redis and wait for reply
+     * No reply type specified.
+     */
+    RedisReplyAsync(DBConnector *db, const RedisCommand& command);
+    RedisReplyAsync(DBConnector *db, std::string command);
+    /*
+     * Send a new command to redis and waits for reply
+     * The reply must be one of REDIS_REPLY_* format (e.g. REDIS_REPLY_STATUS,
+     * REDIS_REPLY_INTEGER, ...)
+     * isFormatted - Set to true if the command is already formatted in redis
+     *               protocol
+     */
+    RedisReplyAsync(DBConnector *db, const RedisCommand& command, int expectedType);
+    RedisReplyAsync(DBConnector *db, std::string command, int expectedType);
+
+    /*
+     * To read the async reply from subscribing request.
+     */
+    RedisReplyAsync(DBConnector *db);
+
+    /* Free the resources */
+    ~RedisReplyAsync();
+
+    /* Return the actual reply object */
+    redisReply *getContext();
+
+    /* Check that the staus is OK, throw exception otherwise */
+    void checkStatusOK();
+
+private:
+    void checkStatus(char *status);
+    void checkReply();
+    void checkReplyType(int expectedType);
+    void nonBlockingRedisReply(DBConnector *db);
+
+    redisReply *m_reply;
+    DBConnector *m_db;
+};
+
+}
--- a/common/redisselect.h
+++ b/common/redisselect.h
@@ -3,6 +3,7 @@
 #include <string>
 #include <memory>
 #include "selectable.h"
+#include "redisreplyasync.h"
 
 namespace swss {
 
@@ -48,6 +49,12 @@ public:
     {
         redisReply *reply = NULL;
 
+        /*
+         * For non-blocking redis connection, the reply has data
+         * to be processed later.
+         */
+        if (!m_subscribe->isBlocking())
+            return;
         if (redisGetReply(m_subscribe->getContext(), (void**)&reply) != REDIS_OK)
             throw "Unable to read redis reply";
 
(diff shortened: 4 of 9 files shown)


lguohan:
what is the purpose for this one? Is it for getting config db data?

lguohan:
for all swss-common code, google unit test is needed.

lguohan:
@oleksandrivantsiv for awareness

common/dbconnector.h
```diff
@@ -22,6 +22,7 @@ class DBConnector
      */
     DBConnector(int db, std::string hostname, int port, unsigned int timeout);
     DBConnector(int db, std::string unixPath, unsigned int timeout);
+    DBConnector(int db, std::string unixPath);
```
qiluo-msft: >DBConnector [](start = 4, length = 11)

Design question: blocking or not should be independent with connection type (unix socket or net socket). I suggest add parameter to existing constructors instead of creating new one.

common/dbconnector.h
```diff
@@ -33,9 +34,15 @@ class DBConnector
     /* Create new context to DB */
     DBConnector *newConnector(unsigned int timeout);
 
+    /*  Create new context to DB with Non-blocking connection */
+    DBConnector *newConnector(void);
```
qiluo-msft: >newConnector [](start = 17, length = 12)

Suggest reuse existing function.

common/dbconnector.h
```diff
@@ -33,9 +34,15 @@ class DBConnector
...
+
+    bool isBlocking(void);
```
qiluo-msft: const function

common/dbconnector.cpp
```diff
@@ -63,6 +64,24 @@ DBConnector::DBConnector(int db, string unixPath, unsigned int timeout) :
...
+    if (m_conn->err)
+        throw system_error(make_error_code(errc::address_not_available),
+                           "Unable to connect to redis (non block unix-socket)");
+    SWSS_LOG_DEBUG("DBConnector db %d, m_conn->fd %d", db, m_conn->fd);
+    m_blocking = false;
+}
+
+bool DBConnector::isBlocking()
+{
+    return m_blocking;
```
qiluo-msft: >m_blocking [](start = 11, length = 10)

Could you get it from m_conn struct, and remove m_blocking member?

common/subscriber.cpp
```diff
@@ -0,0 +1,151 @@
...
+    kco = m_buffer.front();
+    m_buffer.pop_front();
+}
+
+/*
+ * handle redis keyspace pmessage notification.
+ * Expecting pmessage with 4 elements like:
+ * 0) pmessage
+ * 1) __keyspace@4__*
+ * 2) __keyspace@4__:CFG_VLAN_TABLE:Vlan3000
```
taoyl-ms: We have switched to use "|" instead of ":" for table key separators in order not to conflict with IPv6 addresses.
So the current pmessage will look like:
`__keyspace@4__:VLAN|Vlan3000`.

Also, we'll need to support tables with multi-level keys, such as `__keyspace@4__:INTERFACE|Ethernet0|10.0.0.1`.

common/redisreplyasync.h
```diff
@@ -0,0 +1,53 @@
+#pragma once
+
+#include <hiredis/hiredis.h>
+#include <stdexcept>
+#include "dbconnector.h"
+#include "rediscommand.h"
+
+namespace swss {
+
+class RedisReplyAsync
```
qiluo-msft: >RedisReplyAsync [](start = 6, length = 15)

Design question: why not reuse existing RedisReply class, since all the interfaces are compatible. No need to repeat many code lines.

common/redisselect.h
```diff
@@ -70,6 +77,26 @@ class RedisSelect : public Selectable
         RedisReply r(m_subscribe.get(), s, REDIS_REPLY_ARRAY);
     }
 
+    /* PSUBSCRIBE */
+    void psubscribe(DBConnector* db, std::string channelName)
```
qiluo-msft: >psubscribe [](start = 9, length = 10)

Suggest to modify subscribe() instead of creating new one.

common/schema.h
```diff
@@ -42,6 +43,14 @@ namespace swss {
 
 #define COUNTERS_PORT_NAME_MAP          "COUNTERS_PORT_NAME_MAP"
 
+#define CFG_INTF_TABLE_NAME             "CFG_INTF_TABLE"
```
taoyl-ms: Please refer to https://github.com/taoyl-ms/sonic-py-swsssdk/blob/a9e2a9b99e0f619e820ca9bb45635f0fa3bba313/doc/configdb-schema.md for current in-use configDB table names.

common/subscriber.cpp
```diff
@@ -0,0 +1,151 @@
...
+    while(auto ctx0 = r.getContext())
+    {
+        KeyOpFieldsValuesTuple kco;
+        /* if the Key-space notification is empty, try next one. */
+        if (ctx0->type == REDIS_REPLY_NIL)
+        {
+            continue;
+        }
+
+        assert(ctx0->type == REDIS_REPLY_ARRAY);
```
qiluo-msft: >assert [](start = 8, length = 6)

This should be a runtime error or exception.

common/subscriber.cpp
```diff
@@ -0,0 +1,151 @@
...
+    , TableBase(tableName)
+{
+    m_keyspace = "__keyspace@";
+
+    m_keyspace += std::to_string(db->getDB()) + "__:" + tableName + ":*";
+
+    psubscribe(m_db, m_keyspace);
+    /*
+     * Asynchronous keyspace notification is fire and forget.
+     * For any existing keys, they should be fetched explicitely.
```
qiluo-msft: typo

common/subscriber.cpp
```diff
@@ -0,0 +1,151 @@
...
+    : RedisTransactioner(db)
+    , TableBase(tableName)
+{
+    m_keyspace = "__keyspace@";
+
+    m_keyspace += std::to_string(db->getDB()) + "__:" + tableName + ":*";
+
+    psubscribe(m_db, m_keyspace);
+    /*
+     * Asynchronous keyspace notification is fire and forget.
```
qiluo-msft: >fire and forget [](start = 45, length = 15)

fired and forgotten

common/subscriber.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+    hgetall_key += key;
+    RedisReply r(m_db, hgetall_key, REDIS_REPLY_ARRAY);
+    redisReply *reply = r.getContext();
+    values.clear();
+
+    if (!reply->elements)
+        return false;
+
+    if (reply->elements & 1)
```
qiluo-msft: >reply->elements & 1 [](start = 8, length = 19)

reply->elements % 2 == 1

common/subscriber.cpp
```diff
@@ -0,0 +1,151 @@
...
+ * 0) pmessage
+ * 1) __keyspace@4__*
+ * 2) __keyspace@4__:CFG_VLAN_TABLE:Vlan3000
+ * 3) hset
+ */
+void Subscriber::pops(std::deque<KeyOpFieldsValuesTuple> &vkco, std::string /*prefix*/)
+{
+    RedisReplyAsync r(getSubscribeDBC());
+    /* Keep fetching the reply object until all of them are read from redisReader buf */
+    while(auto ctx0 = r.getContext())
```
qiluo-msft: The blank code style are not consistent. Please check all code.

common/redisreplyasync.h
```diff
@@ -0,0 +1,53 @@
...
+    redisReply *getContext();
+
+    /* Check that the staus is OK, throw exception otherwise */
+    void checkStatusOK();
+
+private:
+    void checkStatus(char *status);
+    void checkReply();
+    void checkReplyType(int expectedType);
+    void nonBlockingRedisReply(DBConnector *db);
```
qiluo-msft: >nonBlockingRedisReply [](start = 9, length = 21)

Name it with a verb or verb phrase.

common/dbconnector.h
JipanYanga: Modifying existing construction will have impact on quite a few users of existing blocking DBConnector. 
No use case for non-blocking net socket yet.  We may add it in library if really needed.

common/dbconnector.h
JipanYanga: Reusing existing function requires parameter update.  It looks to me have a separate new function introduces minimal changes.

common/dbconnector.h
JipanYanga: ok

common/dbconnector.cpp
JipanYanga: ok.

common/subscriber.cpp
JipanYanga: It is interesting. Have you updated APPDB and orchagent so the whole system has consistent processing?

common/redisreplyasync.h
JipanYanga: Synchronous redis reply and Asynchronous redis reply have quite different processing behavior.   Forcing reuse of existing RedisReply class might introduce unnecessary complexity.  But we may have more discussion on this.

common/redisselect.h
JipanYanga: subscribe and psubscribe use different type of db connection, synchronous vs asynchronous. and different commands. subscribe vs psubscribe.  Again, we may have more discussion on the approaches.

common/schema.h
JipanYanga: I prefer to have CFG_ prefix for all tables in CFGDB to avoid confusion with tables in APPDB.  I assume that has not been finalized?

common/subscriber.cpp
JipanYanga: ok

common/subscriber.cpp
JipanYanga: ok.

common/subscriber.cpp
JipanYanga: ok

common/subscriber.cpp
JipanYanga: ok

common/redisreplyasync.h
JipanYanga: ok

common/subscriber.cpp
taoyl-ms: No. Currently it is for configDB only. Historically AppDB use hard-coded logic for multi-level keys to avoid v6 address parsing issue. We intend to fix that and make them consistent but that will take time.

JipanYanga: After checking the those modules which will require the config data,   it looks having different key separators will cause unnecessary translation work in config enforcer or ochagent( if it talks to cfgdb directly).   Really don't think having the transition half done is a good practice.

JipanYanga:
target branch not active.

common/schema.h
oleksandrivantsiv: Agree with Taoyu. Shema should not be changed. Config DB is separated DB and there won't be any conflicts.

qiluo-msft:
You could still follow up on this PR, since v1.0.1 branch is fast-forwarded to master. The benefit is no duplicated reviewing.

PR #123, opened 2017-09-16
Make pops public interface of ConsumerTable

qiluo-msft:


stcheng:
hi, could you describe a little bit about why this change is made?

qiluo-msft:
Follow up with https://github.com/Azure/sonic-swss/pull/312
I need to use pops() to improve orch performance.

Issue #125, opened 2017-09-24
Both PFC_WD_DB and CONFIG_DB use db 4


JipanYanga:
Is this the intended design?

schema.h:
    9  #define LOGLEVEL_DB     3
   10: #define PFC_WD_DB       4

database.json
   23:       "CONFIG_DB": {
   24          "db": 4
   25        }

configdb-load.sh
16: echo -en "SELECT 4\nSET CONFIG_DB_INITIALIZED true" | redis-cli

marian-pritsak:
Why isn't CONFIG_DB declared in the schema?

PR #126, opened 2017-09-25
[schema]: Fix DB indeces

marian-pritsak:
CONFIG_DB and PFC_WD_DB used same DB index
Now they are split to 5 and 4


marian-pritsak:
Addresses https://github.com/Azure/sonic-swss-common/issues/125

PR #129, opened 2017-09-27
[keyspace consumer]: Add implementation of keyspace consumer.

oleksandrivantsiv:


oleksandrivantsiv:
Thanks @JipanYanga for his PR https://github.com/Azure/sonic-swss-common/pull/122 based on which those changes were made.

JipanYanga:
Great progress, thanks for the update!

common/schema.h
```diff
@@ -19,9 +19,7 @@ namespace swss {
 #define APP_NEIGH_TABLE_NAME            "NEIGH_TABLE"
 #define APP_ROUTE_TABLE_NAME            "ROUTE_TABLE"
 #define APP_TUNNEL_DECAP_TABLE_NAME     "TUNNEL_DECAP_TABLE"
-#define APP_MIRROR_SESSION_TABLE_NAME   "MIRROR_SESSION_TABLE"
```
lguohan: suggest to make this a separate PR.

oleksandrivantsiv: Done

common/dbconnector.cpp
```diff
@@ -63,6 +64,22 @@ DBConnector::DBConnector(int db, string unixPath, unsigned int timeout) :
     select(this);
 }
 
+/* No timeout argument, non-blocking DB connection */
+DBConnector::DBConnector(int db, string unixPath) :
+    m_db(db)
+{
+    m_conn = redisConnectUnixNonBlock(unixPath.c_str());
```
lguohan: hi sasha, we do not need to have non-block DB connection here. 

if you look at how selectable works in select::select() function, in line 79, we call select(), once the db connect has data, then we call readme() to get the data. since the select guarantees there is data, so the redisGetReply in the readme() will not be blocked. 

once we we get the reply, then we process the reply to get keyspace notification, and then we store in the queue, and then in pops() function, it will read the queue data and then use the get_values() to get the data.

common/consumerkeyspacetable.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include "table.h"
+#include "selectable.h"
+#include "redisselect.h"
+#include "redisapi.h"
+#include "tokenize.h"
+#include "consumerkeyspacetable.h"
+
+namespace swss {
+
+ConsumerKeySpaceTable::ConsumerKeySpaceTable(DBConnector *db, std::string tableName)
```
lguohan: keyspace is the implementation details, we should rename the table to MultiConsumerTable since the key character is you can have multiple consumer to subscribe the same table and they all get notifications.

common/consumerkeyspacetable.cpp
```diff
@@ -0,0 +1,186 @@
...
+    RedisReply r(m_db, hgetall_key, REDIS_REPLY_ARRAY);
+    redisReply *reply = r.getContext();
+    values.clear();
+
+    if (!reply->elements)
+        return false;
+
+    if (reply->elements & 1)
+        throw std::system_error(make_error_code(std::errc::result_out_of_range),
+                           "Invalid redis reply");
```
lguohan: what is & 1? what is this error?

lguohan:
for all swss-common code, google unit test is needed. 

for this consumer table, we need test various cases

del/set
del/set/del
set/del/set

set/del/set/del/set/del....

common/consumerkeyspacetable.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    for (unsigned int i = 0; i < reply->elements; ++i)
+    {
+        table_entries.push_back(reply->element[i]->str);
+    }
+
+    return true;
+}
+
+bool ConsumerKeySpaceTable::getValues(std::string key, std::vector<FieldValueTuple> &values)
```
JipanYanga: The implementation is almost the same as bool Table::get(string key, vector<FieldValueTuple> &values).
What is the reason to implement it here again,   "|" table separator?

common/consumerkeyspacetable.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include <limits>
+#include <hiredis/hiredis.h>
+#include "dbconnector.h"
+#include "table.h"
+#include "selectable.h"
+#include "redisselect.h"
+#include "redisapi.h"
+#include "tokenize.h"
+#include "consumerkeyspacetable.h"
+
```
stcheng: could you add `using namespace std` in all `*.cpp` files and remove the corresponding `std::`?

common/consumerkeyspacetable.cpp
```diff
@@ -0,0 +1,186 @@
...
+        auto ctx = ctx0->element[1];
+        if (m_keyspace.compare(ctx->str))
+        {
+            SWSS_LOG_ERROR("invalid pattern %s returned for pmessage of %s", ctx->str, m_keyspace.c_str());
+            continue;
+        }
+
+        ctx = ctx0->element[2];
+        std::string msg(ctx->str);
+        size_t pos = msg.find(":");
```
stcheng: could you use the `tokenize` function?

oleksandrivantsiv:
@lguohan Unit tests will be implemented soon.

common/consumerkeyspacetable.cpp
oleksandrivantsiv: Done

common/consumerkeyspacetable.cpp
oleksandrivantsiv: This code is removed from class at all. However, it was taken from: https://github.com/Azure/sonic-swss-common/blob/master/common/table.cpp#L31

common/consumerkeyspacetable.cpp
oleksandrivantsiv: In this case simple find is more comfortable.

common/consumerkeyspacetable.cpp
oleksandrivantsiv: Didn't know that there is an implementation of the same functionality. Thanks for the suggestion.

common/consumerkeyspacetable.cpp
oleksandrivantsiv: done

lguohan:
@qiluo-msft , please review.

lguohan:
@oleksandrivantsiv , can you add unit test to this new class?

common/table.cpp
```diff
@@ -33,8 +33,10 @@ bool Table::get(string key, vector<FieldValueTuple> &values)
                            "Unable to connect netlink socket");
 
     for (unsigned int i = 0; i < reply->elements; i += 2)
-        values.push_back(make_tuple(reply->element[i]->str,
+    {
+        values.push_back(make_tuple(stripSpecialSym(reply->element[i]->str),
```
qiluo-msft: >stripSpecialSym [](start = 36, length = 15)

Could you add some comments what it's input like, what is the expected action on it? It's hard to understand now.

common/multiconsumerstatetable.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+    return Selectable::NODATA;
+}
+
+void MultiConsumerStateTable::readMe()
+{
+    redisReply *reply = NULL;
+
+    if (redisGetReply(m_subscribe->getContext(), (void**)&reply) != REDIS_OK)
+        throw "Unable to read redis reply";
```
qiluo-msft: >throw [](start = 8, length = 5)

throw a runtime_exception, or other specific exception type.

common/multiconsumerstatetable.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+    while (auto ctx0 = getReply())
+    {
+        /* Use RedisReply as auto_ptr to destroy reply object */
+        RedisReply r(ctx0);
+        KeyOpFieldsValuesTuple kco;
+        /* if the Key-space notification is empty, try next one. */
+        if (ctx0->type == REDIS_REPLY_NIL)
+        {
+            continue;
```
qiluo-msft: >continue [](start = 12, length = 8)

Is it burning CPU cycles in a while loop to wait a reply? If yes, we should rethink about the design.

common/multiconsumerstatetable.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+    redisReply *reply = NULL;
+
+    if (redisGetReply(m_subscribe->getContext(), (void**)&reply) != REDIS_OK)
+        throw "Unable to read redis reply";
+
+    m_reply_buffer.push_back(reply);
+}
+
+void MultiConsumerStateTable::pops(deque<KeyOpFieldsValuesTuple> &vkco, string /*prefix*/)
```
qiluo-msft: >vkco [](start = 66, length = 4)

clear vkco first?

common/multiconsumerstatetable.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+    SWSS_LOG_ERROR("MultiConsumerStateTable::pops enter %s", getTableName().c_str());
+
+    if (!m_buffer.empty())
+    {
+        vkco = m_buffer;
```
qiluo-msft: >vkco = m_buffer [](start = 8, length = 15)

vkco.insert(m_buffer.begin(), m_buffer.end())

common/multiconsumerstatetable.cpp
```diff
@@ -0,0 +1,172 @@
...
+        m_buffer.clear();
+        return;
+    }
+
+    while (auto ctx0 = getReply())
```
qiluo-msft: >ctx0 [](start = 16, length = 4)

ctx0 -> reply

common/multiconsumerstatetable.cpp
```diff
@@ -0,0 +1,172 @@
...
+    if (!m_buffer.empty() || !m_reply_buffer.empty())
+    {
+        return Selectable::DATA;
+    }
+
+    if (redisGetReplyFromReader(m_subscribe->getContext(),
+                                (void**)&reply) != REDIS_OK)
+    {
+        return Selectable::ERROR;
+    } else if (reply != NULL)
```
qiluo-msft: >else [](start = 6, length = 4)

keep if-else style consistent.

common/multiconsumerstatetable.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <string>
+#include <deque>
+#include "dbconnector.h"
+#include "consumertablebase.h"
+
+namespace swss {
+
+class MultiConsumerStateTable : public ConsumerTableBase
```
qiluo-msft: >MultiConsumerStateTable [](start = 6, length = 23)

Name is confusing. It has nothing to do with ConsumerStateTable or Multi.

common/multiconsumerstatetable.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+void MultiConsumerStateTable::readMe()
+{
+    redisReply *reply = NULL;
+
+    if (redisGetReply(m_subscribe->getContext(), (void**)&reply) != REDIS_OK)
+        throw "Unable to read redis reply";
+
+    m_reply_buffer.push_back(reply);
```
qiluo-msft: >reply [](start = 29, length = 5)

How about capturing it by RedisReply wrapper, and storing it in a container, so there is no chance of resource leak?

common/multiconsumerstatetable.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include "tokenize.h"
+
+using namespace std;
+
+namespace swss {
+
+MultiConsumerStateTable::MultiConsumerStateTable(DBConnector *db, string tableName)
+    : ConsumerTableBase(db, tableName, "|"), m_table(db, tableName, getSeparator())
+{
+    m_keyspace = "__keyspace@";
```
qiluo-msft: Suggest add some comments at the file header, and unit test, to help others read the code.

common/multiconsumerstatetable.h
JipanYanga: Maybe we could get some hint from NotificationConsumer, which is quite similar to the keyspace consumer implementation.

common/multiconsumerstatetable.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+    m_keyspace += to_string(db->getDB()) + "__:" + tableName + getSeparator() +"*";
+
+    psubscribe(m_db, m_keyspace);
+
+    vector<string> keys;
+    m_table.getTableKeys(keys);
```
JipanYanga: In future implementation, we have the need to get some switch level configuration like switch mac, switch flooding control from configDB, and apply the setting to port, LAG, VLAN and etc.   For settings like mac, It is not easy to dynamically modifying it on port/lag/VLAN, so we want to fetch and save them before all keyspace notification handling in the Select framework.   For other settings in switch table, they may be processed as usual in select framework.

Is it possible to provide option for specifying whether the existing keys should be fetched after psubscribe in this constructor?

common/multiconsumerstatetable.h
oleksandrivantsiv: Name was suggested by @lguohan:
https://github.com/Azure/sonic-swss-common/pull/129#discussion-diff-141996254R13

oleksandrivantsiv: @lguohan Do you agree to change the name?

common/multiconsumerstatetable.cpp
lguohan: >throw "Unable to read redis reply"; [](start = 8, length = 35)

add {}

common/multiconsumerstatetable.cpp
oleksandrivantsiv: RedisReply doesn't have copy constructor implemented. It is not possible to do without modification of RedisReply class. In pops function every reply pointer is put into RedisReply object to free resourses correctly:
https://github.com/Azure/sonic-swss-common/pull/129/files#diff-e21c43ea3f32ce9e0342a44a352e55bbR91

common/multiconsumerstatetable.cpp
oleksandrivantsiv: From the performance point of view it is better to use copy constructor.

common/multiconsumerstatetable.cpp
oleksandrivantsiv: No, we just iterating over replies that we got and do validation. Replies are received only after modification of subscribed table.

common/table.cpp
oleksandrivantsiv: sure

common/multiconsumerstatetable.cpp
oleksandrivantsiv: sure.

common/multiconsumerstatetable.cpp
oleksandrivantsiv: will change the code

common/multiconsumerstatetable.cpp
oleksandrivantsiv: sure

common/multiconsumerstatetable.cpp
oleksandrivantsiv: agree

common/multiconsumerstatetable.cpp
oleksandrivantsiv: In general pops method is called only after event is already received.

lguohan: why not return? 

if there is no notification, the orchagent will be blocked at
select here.
[https://github.com/Azure/sonic-swss/blob/535458432b8bde67829339c8824ac10fa347bc8f/orchagent/orchdaemon.cpp#L135](https://github.com/Azure/sonic-swss/blob/535458432b8bde67829339c8824ac10fa347bc8f/orchagent/orchdaemon.cpp#L135)


only when there is notification, the orchagent will use pops() to get the data.

[https://github.com/Azure/sonic-swss/blob/6334d57dbd121c52a920473493fe2d927f439bd4/orchagent/orch.cpp#L87](https://github.com/Azure/sonic-swss/blob/6334d57dbd121c52a920473493fe2d927f439bd4/orchagent/orch.cpp#L87)



---
In reply to: [143390924](https://github.com/Azure/sonic-swss-common/pull/129#discussion_r143390924) [](ancestors = 143390924)

lguohan: see my other comments.

---
In reply to: [143460751](https://github.com/Azure/sonic-swss-common/pull/129#discussion_r143460751) [](ancestors = 143460751)

common/multiconsumerstatetable.h
lguohan: yes, multiConsumerStateTable is suggested by me. The reason is that it is the state table, since when you get the data, it is the latest state because redis only store the latest value for a key. In this sense, it is the same as consumerstate table. Multi is added because this table supports Mutiple consumer compared with the original consumerstate table.

also, I think we should have a multProducerStateTable is write the the data later. current, we are just using the plain redis set command to do it. I am not sure this is the most optimized way to do it.

JipanYanga: Just curious,  it is a typical  pubsub implementation and it is built on redis pubsub mechanism, why are we so keen to call it *consumer* instead of *subscriber*?  Same for *producer* vs *publisher*. 
https://redis.io/topics/pubsub
https://redis.io/topics/notifications

common/multiconsumerstatetable.cpp
qiluo-msft: If at the end any thing is not popped, then it is leaked.

common/multiconsumerstatetable.cpp
qiluo-msft: It is not right in semantics.

common/multiconsumerstatetable.cpp
oivantsiv: getReply method returns reply message from   cache (from m_replay_buffer deque). It is non blocking. We can't return until messages remains in cache.

oivantsiv: getReply method returns reply message from   cache (from m_replay_buffer deque). It is non blocking. We can't return until messages remains in cache.

oivantsiv: getReply method returns reply message from   cache (from m_replay_buffer deque). It is non blocking. We can't return until messages remains in cache.

common/multiconsumerstatetable.cpp
oivantsiv: I'm not sure that I understand correctly. Can you explain what wrong in copy constructor?

common/multiconsumerstatetable.cpp
oivantsiv: Key space event is not persistent. For each DB operation it is generated only once. To make MuliConsumerStateTable working with Select framework I had to override "readCache" and "readMe" methods. Those methods reads keyspace events and put them into event queue. "pops" method works only with data stored in the cache. If cache is empty it is not called.

common/multiconsumerstatetable.cpp
oivantsiv: This can happen only if application stuck. And leaked memory is not the for think in this case :). 
Will add copy constructor to RedisReply. Make will use it as a wrapper.

common/multiconsumerstatetable.cpp
oivantsiv: getReply inplementation:
https://github.com/Azure/sonic-swss-common/pull/129/files#diff-e21c43ea3f32ce9e0342a44a352e55bbR159
readCache implementation:
https://github.com/Azure/sonic-swss-common/pull/129/files#diff-e21c43ea3f32ce9e0342a44a352e55bbR45
readMe implementation:
https://github.com/Azure/sonic-swss-common/pull/129/files#diff-e21c43ea3f32ce9e0342a44a352e55bbR67

common/multiconsumerstatetable.cpp
qiluo-msft: Leaking is not acceptable in any case. Without deep investigation, I just feel a copy constructor is not necessary for this situation. Let me know if it is wrong.

---
In reply to: [143554061](https://github.com/Azure/sonic-swss-common/pull/129#discussion_r143554061) [](ancestors = 143554061)

common/multiconsumerstatetable.cpp
qiluo-msft: I am trying to read your mind that you intend to insert all the elements in m_buffer into vkco, and then clear m_buffer. The implement just as what you intend.

vkco.insert(m_buffer.begin(), m_buffer.end());
m_buffer.clear()

Trust the compiler to optimize base on context. If the compiler doesn't utilize the opportunity, try optimize without sacrifice readability. If finally it is impossible, you want to use this trick, give intensive comment like // TRICK:


---
In reply to: [143551979](https://github.com/Azure/sonic-swss-common/pull/129#discussion_r143551979) [](ancestors = 143551979)

common/multiconsumerstatetable.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+        vkco.push_back(kco);
+    }
+
+    m_reply_buffer.clear();
+
+    return;
+}
+
+redisReply *MultiConsumerStateTable::getReply()
```
qiluo-msft: >getReply [](start = 37, length = 8)

Actually you mean to popReply?

common/multiconsumerstatetable.h
qiluo-msft: It is still easy to confuse people because the name is pointing against ConsumerStateTable. How about SubscriberStateTable? Just a quick try.

---
In reply to: [143472274](https://github.com/Azure/sonic-swss-common/pull/129#discussion_r143472274) [](ancestors = 143472274)

common/multiconsumerstatetable.cpp
oleksandrivantsiv: Yes, will rename and add comments to make it clear

common/table.cpp
oleksandrivantsiv: done

common/multiconsumerstatetable.cpp
oleksandrivantsiv: done

common/multiconsumerstatetable.cpp
oleksandrivantsiv: done

common/multiconsumerstatetable.cpp
oleksandrivantsiv: done

qiluo-msft:
BTW, could you do us a favor _not_ to force push in a pull request? Leaving the old commits will keep discussion context. We have some internal tools to compare between commits so we can easily locate what you modify incrementally.

oleksandrivantsiv:
@qilimsft Will not do push force in the future.

oleksandrivantsiv:
@qiluo-msft @lguohan Unittests are added. Please review.

oleksandrivantsiv:
@lguohan @qiluo-msft Test stuck because keyspace notification is disabled in test environment. To enable it the following line should be added to /etc/redis/redis.conf file:
https://redis.io/topics/notifications
```
notify-keyspace-events AKE
```
Can you please take a look?

common/table.cpp
qiluo-msft: I cannot find any related comments.

---
In reply to: [143827736](https://github.com/Azure/sonic-swss-common/pull/129#discussion_r143827736) [](ancestors = 143827736)

common/multiconsumerstatetable.h
lguohan: @qiluo-msft, suggests SubscriberStateTable. I agree subscriber is a more accurate name. Consumer consumes the data, and no other consumer can share the same data.

lguohan: Yes, i agree.

---
In reply to: [143455459](https://github.com/Azure/sonic-swss-common/pull/129#discussion_r143455459) [](ancestors = 143455459)

common/multiconsumerstatetable.cpp
lguohan: the dependency issue should be solved by statedb, in this case the port/lag/vlan cfgmgr should not be started until someone configure the switch like mac, flooding controls and write them to the state db.

qiluo-msft:
I fixed. Please go ahead.

---
In reply to: [337655769](https://github.com/Azure/sonic-swss-common/pull/129#issuecomment-337655769) [](ancestors = 337655769)

lguohan:
retest this please

common/multiconsumerstatetable.cpp
oleksandrivantsiv: Compiler won't optimize that. It can't optimize, change or remove methods logic. 
Copy constructor knows elements count it needs to copy. It will allocate memory once and won't do re-allocation after that. 

Insert method inserts elements one by one and based on elements count re-allocation of internal structure may be required. To achieve the same performance with insert method in this case vkco.reserve(m_buffer.size()) should be called additionally.

common/table.cpp
oleksandrivantsiv: @qiluo-msft I've added description to into header file:
https://github.com/Azure/sonic-swss-common/pull/129/files#diff-980cf2c21abcb620e52c9c6dddfa684fR146

common/multiconsumerstatetable.cpp
oleksandrivantsiv: Added description for each function in the header file.

common/table.cpp
oleksandrivantsiv: Description was added in header file.

common/multiconsumerstatetable.h
oleksandrivantsiv: Done

common/multiconsumerstatetable.cpp
oleksandrivantsiv: Done

common/multiconsumerstatetable.cpp
oleksandrivantsiv: Done

tests/redis_multi_state_ut.cpp
```diff
@@ -0,0 +1,382 @@
+#include <common/subscriberstatetable.h>
```
qiluo-msft: #include "common/subscriberstatetable.h"
at the end

oleksandrivantsiv: done

common/subscriberstatetable.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+        vkco.push_back(kco);
+    }
+
+    m_keyspace_event_buffer.clear();
+
+    return;
+}
+
+shared_ptr<RedisReply> SubscriberStateTable::popEventBuffer()
```
qiluo-msft: >RedisReply [](start = 11, length = 10)

Thanks for the improvement. The shared_ptr seems not necessary. If implementing two functions
```
RedisReply& front();
void pop();
```
as STL containers, the shared_ptr could be removed.

oleksandrivantsiv: Can this be changed in separated PR? In other case this one never be closed :)

qiluo-msft: I prefer changing it in this PR. This is the only concern from my side.

oleksandrivantsiv: Can you describe a little bit more? Do you want to remove popEventBuffer and implement those two methods instead? I will be equivalent to just using m_keyspace_event_buffer without wrapper function. What is your concern in using shared_ptr?

qiluo-msft: like this
```
void popEventBuffer();
RedisReply& frontEventBuffer();
```
It's ok to directly use m_keyspace_event_buffer. My arguments are
1. shared_ptr introduces time and space overhead
2. you use shared_ptr for preventing implement copy constructor. However the alternative I proposed (or you proposed) may be much easier, and make code clean.

oleksandrivantsiv: I have the following concerns on this:
1. It won't be possible to do without modification of RedisReply. To emplace RedisReply object into deque *move* constructor should be implemented. RedisReply is not covered with unit tests. It will be hard to verify that modification of this class doesn't have side effects.
2. shared_prt is efficient enough. Time and space overhead is minimal.
3. It will have negative effect on code readability because object should reference should be taken in one place and popped in another. With shared_ptr it is enough to get object in one place and it will be deleted automatically.

oleksandrivantsiv: An article which describes memory and performance overhead given by smart pointers:
http://www.modernescpp.com/index.php/memory-and-performance-overhead-of-smart-pointer

common/subscriberstatetable.cpp
```diff
@@ -0,0 +1,183 @@
...
+    while (auto event = popEventBuffer())
+    {
+        KeyOpFieldsValuesTuple kco;
+        /* if the Key-space notification is empty, try next one. */
+        if (event->getContext()->type == REDIS_REPLY_NIL)
+        {
+            continue;
+        }
+
+        assert(event->type == REDIS_REPLY_ARRAY);
```
qiluo-msft: >event->type [](start = 15, length = 11)

Build break under debug mode.

lguohan:
@oleksandrivantsiv , please fix the debug mode build error in next PR. 

Thanks a lot for adding this.

PR #130, opened 2017-09-27
Change netlink socket to non-blocking mode by default.

JipanYanga:
As recommended at http://www.infradead.org/~tgr/libnl/doc/core.html : If a socket is used to only receive notifications it usually is best to put the socket in non-blocking mode and periodically poll for new notifications.

Also due to the incorrect setting of NLM_F_MULTI flag in some netlink messages, netlink process got stuck in "static int recvmsgs(struct nl_sock *sk, struct nl_cb *cb)" function of libnl if a socket is in blocking mode.

4.x kernel has fix to the wrong use of NLM_F_MULTI problem: github.com/torvalds/linux/commit/46c264daaaa569e24f8aba877d0fd8167c42a9a4 but introducing the fix to 3.16 will break Debian Jessie ABI. Setting netlink socket to non-blocking mode to get around the problem without kernel change.

### DIFF
--- a/common/netlink.cpp
+++ b/common/netlink.cpp
@@ -33,6 +33,7 @@ NetLink::NetLink() :
                            "Unable to connect netlink socket");
     }
 
+    nl_socket_set_nonblocking(m_socket);
     /* Set socket buffer size to 256KB */
     nl_socket_set_buffer_size(m_socket, 2097152, 0);
 }
@@ -92,6 +93,8 @@ void NetLink::readMe()
     {
         if (err == -NLE_NOMEM)
             SWSS_LOG_ERROR("netlink reports out of memory on reading a netlink socket. High possiblity of a lost message");
+        else if (err == -NLE_AGAIN)
+            SWSS_LOG_DEBUG("netlink reports NLE_AGAIN on reading a netlink socket");
         else
             SWSS_LOG_ERROR("netlink reports an error=%d on reading a netlink socket", err);
     }


common/netlink.cpp
```diff
@@ -33,6 +33,7 @@ NetLink::NetLink() :
                            "Unable to connect netlink socket");
     }
 
+    nl_socket_set_nonblocking(m_socket);
```
prsunny: Looks like issue is we received a NLM_F_MULTI message but haven't got a NLMSG_DONE. So instead of waiting, the socket is set to non-blocking. Just a thought, if we provide this partial message to application, will there be any side-effects?

JipanYanga: In sonic, the way we use netlink is to register callback function for netlink messages: 
nl_socket_modify_cb(m_socket, NL_CB_VALID, NL_CB_CUSTOM, onNetlinkMsg, this);


The onNetlinMsg() function will be called in libnl3 static int recvmsgs(struct nl_sock *sk, struct nl_cb *cb):
			/* Valid message (not checking for MULTIPART bit to
			 * get along with broken kernels. NL_SKIP has no
			 * effect on this.  */
			if (cb->cb_set[NL_CB_VALID])
				NL_CB_CALL(cb, NL_CB_VALID, msg);

libnl dosn't concatenate multiple parts into one message. No matter it is a multipart messages or single part message, libnl processes one part each time which means the onNetlinMsg is called for each part (which could be a whole message or a part of a message).     Changing netlink socket from blocking mode to non-blocking mode doesn't affect that behavior.

Based on the comments and implementation for 
	int nl_recv(struct nl_sock *sk, struct sockaddr_nl *nla,
	    unsigned char **buf, struct ucred **creds)
if the the socket buffer is not large enough, then MSG_CTRUNC may be set on the message and nl_recv() will try to read the socket again to complete the message, if no data is immediately available on the socket, non-blocking socket will return without blocking.

In Sonic code, the buffer has been set to 256KB which should be big enough to avoid MSG_CTRUNC.
/* Set socket buffer size to 256KB */
    nl_socket_set_buffer_size(m_socket, 2097152, 0);
    
Please let me know if I missed anything?

prsunny: Agree. Change looks good to me.

PR #131, opened 2017-09-28
Add exec library for executing shell commands

JipanYanga:
Add support for executing shell command in c++ code.

jipan@043cf76fe4fe:/sonic/src/sonic-swss-common/tests$ ./tests --gtest_filter=EXEC*
Running main() from gtest_main.cc
Note: Google Test filter = EXEC*
[==========] Running 2 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 2 tests from EXEC
[ RUN      ] EXEC.result
[       OK ] EXEC.result (6 ms)
[ RUN      ] EXEC.error
[       OK ] EXEC.error (2 ms)
[----------] 2 tests from EXEC (8 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 1 test case ran. (8 ms total)
[  PASSED  ] 2 tests.
jipan@043cf76fe4fe:/sonic/src/sonic-swss-common/tests$ ./tests --gtest_filter=EXEC*
Running main() from gtest_main.cc
Note: Google Test filter = EXEC*
[==========] Running 2 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 2 tests from EXEC
[ RUN      ] EXEC.result
[       OK ] EXEC.result (5 ms)
[ RUN      ] EXEC.error
[       OK ] EXEC.error (1 ms)
[----------] 2 tests from EXEC (6 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 1 test case ran. (6 ms total)
[  PASSED  ] 2 tests.

### DIFF
--- a/common/exec.cpp
+++ b/common/exec.cpp
@@ -1,3 +1,5 @@
+#include <cerrno>
+#include <cstring>
 #include <array>
 #include "exec.h"
 #include "common/logger.h"
@@ -6,11 +8,12 @@ using namespace std;
 
 namespace swss {
 
-string exec(const char* cmd)
+const int buffsz = 128;
+
+int exec(const string &cmd, string &stdout)
 {
-    array<char, 128> buffer;
-    string result;
-    shared_ptr<FILE> pipe(popen(cmd, "r"), pclose);
+    array<char, buffsz> buffer;
+    FILE* pipe = popen(cmd.c_str(), "r");
 
     if (!pipe)
     {
@@ -21,14 +24,22 @@ string exec(const char* cmd)
         throw runtime_error(errmsg);
     }
 
-    while (!feof(pipe.get()))
+    while (!feof(pipe))
     {
-        if (fgets(buffer.data(), 128, pipe.get()) != NULL)
-            result += buffer.data();
+        if (fgets(buffer.data(), buffsz, pipe) != NULL)
+        {
+            stdout += buffer.data();
+        }
     }
 
-    SWSS_LOG_DEBUG("%s : %s", cmd, result.c_str());
-    return result;
+    int ret = pclose(pipe);
+    if (ret != 0)
+    {
+        SWSS_LOG_ERROR("%s: %s", cmd.c_str(), strerror(errno));
+    }
+    SWSS_LOG_DEBUG("%s : %s", cmd.c_str(), stdout.c_str());
+
+    return ret;
 }
 
 }
--- a/common/exec.cpp
+++ b/common/exec.cpp
@@ -1,25 +1,24 @@
-#include <cstdio>
-#include <iostream>
-#include <stdexcept>
 #include <array>
 #include "exec.h"
 #include "common/logger.h"
 
+using namespace std;
+
 namespace swss {
 
-std::string exec(const char* cmd)
+string exec(const char* cmd)
 {
-    std::array<char, 128> buffer;
-    std::string result;
-    std::shared_ptr<FILE> pipe(popen(cmd, "r"), pclose);
+    array<char, 128> buffer;
+    string result;
+    shared_ptr<FILE> pipe(popen(cmd, "r"), pclose);
 
     if (!pipe)
     {
-        std::string errmsg(cmd);
+        string errmsg(cmd);
 
         errmsg = "popen(" + errmsg + ") failed!";
         SWSS_LOG_ERROR("exec: %s", errmsg.c_str());
-        throw std::runtime_error(errmsg);
+        throw runtime_error(errmsg);
     }
 
     while (!feof(pipe.get()))
--- a/common/exec.cpp
+++ b/common/exec.cpp
@@ -1,0 +1,35 @@
+#include <cstdio>
+#include <iostream>
+#include <stdexcept>
+#include <array>
+#include "exec.h"
+#include "common/logger.h"
+
+namespace swss {
+
+std::string exec(const char* cmd)
+{
+    std::array<char, 128> buffer;
+    std::string result;
+    std::shared_ptr<FILE> pipe(popen(cmd, "r"), pclose);
+
+    if (!pipe)
+    {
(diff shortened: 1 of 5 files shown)


common/exec.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "exec.h"
+#include "common/logger.h"
+
+using namespace std;
+
+namespace swss {
+
+std::string exec(const char* cmd)
+{
+    std::array<char, 128> buffer;
```
stcheng: namespace std is already used

JipanYanga: Thanks for the review!  It might be better to remove "using namespace std;"?    It works too.  If it is ok, I could use force push for the change to reduce the number of minor commit log.

common/exec.cpp
```diff
@@ -0,0 +1,35 @@
...
+        errmsg = "popen(" + errmsg + ") failed!";
+        SWSS_LOG_ERROR("exec: %s", errmsg.c_str());
+        throw std::runtime_error(errmsg);
+    }
+
+    while (!feof(pipe.get()))
+    {
+        if (fgets(buffer.data(), 128, pipe.get()) != NULL)
+            result += buffer.data();
+    }
```
qiluo-msft: Try use highlevel file encapsulation instead of handling buffer and trunks directly.

One reference:
[https://stackoverflow.com/questions/2602013/read-whole-ascii-file-into-c-stdstring/2602258#2602258](https://stackoverflow.com/questions/2602013/read-whole-ascii-file-into-c-stdstring/2602258#2602258)

common/exec.cpp
stcheng: hmm. sorry for the late reply. actually I'd like to keep it as most of the files here in this folder following the convention of having `using namespace std` in `*.cpp` files.

common/exec.cpp
JipanYanga: Not an expert in C++, so I did some search trying to make it the way you suggested.
It looks not possible to do it: 
https://stackoverflow.com/questions/10667543/creating-fstream-object-from-a-file-pointer

I do need to remove the unused "#include <iostream>" to avoid further confusion.

common/exec.cpp
JipanYanga: Ok, will add using namespace std and remove the std:: prefix.

common/exec.cpp
```diff
@@ -0,0 +1,34 @@
+#include <array>
+#include "exec.h"
+#include "common/logger.h"
+
+using namespace std;
+
+namespace swss {
+
+string exec(const char* cmd)
```
lguohan: This API needs to provide return value for the command, because we need to use the return value to determine whether the command is successful or not.

the return value of the command can be obtained by pclose().

check here.  http://man7.org/linux/man-pages/man3/popen.3.html

common/exec.cpp
```diff
@@ -0,0 +1,34 @@
...
+        string errmsg(cmd);
+
+        errmsg = "popen(" + errmsg + ") failed!";
+        SWSS_LOG_ERROR("exec: %s", errmsg.c_str());
+        throw runtime_error(errmsg);
+    }
+
+    while (!feof(pipe.get()))
+    {
+        if (fgets(buffer.data(), 128, pipe.get()) != NULL)
```
qiluo-msft: >128 [](start = 33, length = 3)

make 128 a constant variable?

common/exec.cpp
qiluo-msft: Thanks for the info. You may keep it.

---
In reply to: [141997094](https://github.com/Azure/sonic-swss-common/pull/131#discussion_r141997094) [](ancestors = 141997094)

common/exec.cpp
JipanYanga: Yes, It will be good to have pclose() to check the command return status.  
Usually we want the output of a command, maybe just check the return status and log any error as done in _run_command() of bgpcfgd?

Otherwise there are two value to be returned, command output and command return status. Multiple ways to do it if needed, personally I prefer
string exec(const char* cmd, int *ret) or string exec(const char* cmd, int& ret) for not having to worry string memory issue. 

Please give your suggestion, thanks.

lguohan: ret = exec(string &cmd, string &stdout)?

lguohan: we also need to test the failed case in the unit test.

JipanYanga: done.

common/exec.cpp
JipanYanga: changed to "const int buffsz = 128;"

common/exec.cpp
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    int ret = pclose(pipe);
+    if (ret != 0)
+    {
+        SWSS_LOG_ERROR("%s: %s", cmd.c_str(), strerror(errno));
+    }
+    SWSS_LOG_DEBUG("%s : %s", cmd.c_str(), stdout.c_str());
+
+   return ret;
```
lguohan: >    [](start = 0, length = 3)

4 ws please.

common/exec.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+        errmsg = "popen(" + errmsg + ") failed!";
+        SWSS_LOG_ERROR("exec: %s", errmsg.c_str());
+        throw runtime_error(errmsg);
+    }
+
+    while (!feof(pipe))
+    {
+        if (fgets(buffer.data(), buffsz, pipe) != NULL)
+            stdout += buffer.data();
```
lguohan: >stdout += buffer.data(); [](start = 12, length = 24)

use {} please

common/exec.cpp
JipanYanga: done.

common/exec.cpp
JipanYanga: done

PR #134, opened 2017-10-04
[schema.h]: Add plugin table name

marian-pritsak:
PLUGIN_TABLE is used to register plugins in flex counter module


stcheng:
what is the purpose of the 'SAI_OBEJCT_TYPE'?

marian-pritsak:
It is type of plugin. Currently type of plugin is deduced from type of objects to which it is dedicated, but Guohan suggested to create a separate type for plugins, which will be introduced soon in next PRs.

qiluo-msft:
'SAI_OBEJCT_TYPE' is a bad name. There are so many enums in SAI repo that in the pattern 'SAI_OBEJCT_TYPE_*', and this instance has nothing to do with them.

In concept, swss-common should have no knowledge about SAI, if it is possible.

PR #137, opened 2017-10-06
[logger]: Add destructor to detach Logger priority thread

stcheng:
The Logger priority thread was not gracefully detached. This would
cause `terminate called without an active exception` after the main
program exits.

Add related unit test to check if the program exits successfully with
or without the thread exists.

### DIFF
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -15,6 +15,12 @@
 
 namespace swss {
 
+Logger::~Logger() {
+    if (m_prioThread) {
+        m_prioThread->detach();
+    }
+}
+
 const Logger::PriorityStringMap Logger::priorityStringMap = {
     { "EMERG", SWSS_EMERG },
     { "ALERT", SWSS_ALERT },
@@ -177,6 +183,19 @@ void Logger::wthrow(Priority prio, const char *fmt, ...)
     throw std::runtime_error(buffer);
 }
 
+std::string Logger::priorityToString(Logger::Priority prio)
+{
+    for (const auto& i : priorityStringMap)
+    {
+        if (i.second == prio)
+        {
+            return i.first;
+        }
+    }
+
+    return "UNKNOWN";
+}
+
 Logger::ScopeLogger::ScopeLogger(int line, const char *fun) : m_line(line), m_fun(fun)
 {
     swss::Logger::getInstance().write(swss::Logger::SWSS_DEBUG, ":> %s: enter", m_fun);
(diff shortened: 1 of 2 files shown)


stcheng:
In the core dump, the following information is always shown up due to this issue.

```
[New Thread 0x7ffff6504700 (LWP 264)]
terminate called without an active exception

Program received signal SIGABRT, Aborted.
0x00007ffff6bdf067 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56      ../nptl/sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff6bdf067 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff6be0448 in __GI_abort () at abort.c:89
#2  0x00007ffff74ccb3d in __gnu_cxx::__verbose_terminate_handler() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff74cabb6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff74cac01 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff7ba25e7 in swss::Logger::~Logger() () from /usr/lib/x86_64-linux-gnu/libswsscommon.so.0
#6  0x00007ffff6be1b29 in __run_exit_handlers (status=0, listp=0x7ffff6f4f5a8 <__exit_funcs>,
    run_list_atexit=run_list_atexit@entry=true) at exit.c:82
#7  0x00007ffff6be1b75 in __GI_exit (status=<optimized out>) at exit.c:104
#8  0x00007ffff6bcbb4c in __libc_start_main (main=0x404650 <main(int, char**)>, argc=2, argv=0x7fffffffeca8,
    init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffec98)
    at libc-start.c:321
#9  0x0000000000404fac in _start ()
```

lguohan:
where is the test?

stcheng:
@lguohan previously i was able to write tests. However, the current logger only supports UNIX socket to connect to the redis database. This will result in failure to connect to the redis while running the test.

I'll generate a separate pull request so that the Logger class could connect to the database via IP/port combination so that I could add unit test later.

PR #138, opened 2017-10-06
[schema.h]: Add queue map names

marian-pritsak:
Add queue oid to port oid map name
Add queue oid to queue index map name


marian-pritsak:
This change is required for PFC WD. It will allow a quick lookup with queue OID as a key.

PR #139, opened 2017-10-10
[schema.h]: Add PLUGIN_TYPE

marian-pritsak:
This define will replace SAI_OBJECT_TYPE but in order
not to break build, it will be removed after the change in client
code.


marian-pritsak:
@stcheng @lguohan Can you please merge?

stcheng:
actually this name is not intuitive to me. maybe 'LUA_PLUGIN' prefix is better?
also could you move these two 'PLUGIN' lines together?

marian-pritsak:
Done

stcheng:
Later could you also update the `PLUGIN_TABLE` to `LUA_PLUGIN_TABLE`?

lguohan:
is this still needed? any dependency? can i merge?

marian-pritsak:
It's needed. It's part of renaming 'PFC WD' to 'FLEX_COUNTER'.

PR #140, opened 2017-10-10
[schema.h]: Add flex counter defines

marian-pritsak:
These definitions intend to substitute
PFCWD related strings. Deprecated values will
be removed after client code will be changed.


marian-pritsak:
@stcheng @lguohan Can you please merge?

PR #143, opened 2017-10-10
Add stateDB table name for lag and vlan

JipanYanga:

### DIFF
--- a/common/schema.h
+++ b/common/schema.h
@@ -70,6 +70,8 @@ namespace swss {
 /***** STATE DATABASE *****/
 
 #define STATE_PORT_TABLE_NAME           "PORT_TABLE"
+#define STATE_LAG_TABLE_NAME            "LAG_TABLE"
+#define STATE_VLAN_TABLE_NAME           "VLAN_TABLE"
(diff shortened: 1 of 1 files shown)


common/schema.h
```diff
@@ -70,6 +70,8 @@ namespace swss {
 /***** STATE DATABASE *****/
 
 #define STATE_PORT_TABLE_NAME           "PORT_TABLE"
+#define STATE_LAG_TABLE_NAME            "PORTCHANNEL_TABLE"
```
stcheng: I would suggest to change it to LAG_TABLE

JipanYanga: Yes, LAG_TABLE is more aligned with APPL_DB.    How about also have "INTERFACE_TABLE" updated to "INTF_TABLE"?

It looks current python part of code changes had cfgDB taking a quite different table naming style.

JipanYanga:
Changed to #define STATE_LAG_TABLE_NAME            "LAG_TABLE"

common/schema.h
JipanYanga: Changed to #define STATE_LAG_TABLE_NAME "LAG_TABLE"

PR #146, opened 2017-10-21
Add configDB table names to schema.h

JipanYanga:

### DIFF
--- a/common/schema.h
+++ b/common/schema.h
@@ -73,6 +73,16 @@ namespace swss {
 #define PLUGIN_TABLE "PLUGIN_TABLE"
 #define SAI_OBJECT_TYPE "SAI_OBJECT_TYPE"
 
+/***** CONFIGUATION DATABASE *****/
+#define CFG_PORT_TABLE_NAME             "PORT"
+#define CFG_INTF_TABLE_NAME             "INTERFACE"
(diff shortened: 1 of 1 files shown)


lguohan:
@taoyl-ms  to review.

common/schema.h
```diff
@@ -73,6 +73,16 @@ namespace swss {
 #define PLUGIN_TABLE "PLUGIN_TABLE"
 #define SAI_OBJECT_TYPE "SAI_OBJECT_TYPE"
 
+/***** CONFIGUATION DATABASE *****/
+#define CFG_PORT_TABLE_NAME             "PORT"
+#define CFG_INTF_TABLE_NAME             "INTERFACE"
+#define CFG_LAG_INTF_TABLE_NAME         "PORTCHANNEL_INTERFACE"
```
stcheng: why we're using LAG in the TABLE name but PORTCHANNEL in the actual string?

JipanYanga: Using LAG in the TABLE name macro to be consistent with app and state DB, it is less confusing when writing code.

Real table name PORTCHANNEL_INTERFACE is from minigraph, would like to keep it as it is at least for now.

stcheng: ok

PR #152, opened 2017-10-27
[rediscommand]: move class implementation to cpp file for rediscmd class

lguohan:


common/rediscommand.h
```diff
@@ -14,79 +14,32 @@ typedef std::tuple<std::string, std::string, std::vector<FieldValueTuple> > KeyO
 
 class RedisCommand {
 public:
-    RedisCommand() : temp(NULL) { }
-    ~RedisCommand() { redisFreeCommand(temp); }
-    RedisCommand(RedisCommand& that);
-    RedisCommand& operator=(RedisCommand& that);
```
qiluo-msft: FYI: these command disable default copy constructor and default operator=. If they are missing, C++ will provide a bitwise copying implementation, which is wrong.

ref: [https://stackoverflow.com/questions/6077143/disable-copy-constructor](https://stackoverflow.com/questions/6077143/disable-copy-constructor)

suggest according to C++11:
```
    RedisCommand(RedisCommand& that) = delete;
    RedisCommand& operator=(RedisCommand& that) = delete;
```

lguohan: added.

PR #154, opened 2017-10-29
[consumer]: add pop with separate key, op, field values as arguments

lguohan:


common/consumertablebase.cpp
```diff
@@ -0,0 +1,47 @@
...
+            kfvOp(kco).clear();
+            return;
+        }
+    }
+
+    kco = m_buffer.front();
+    m_buffer.pop_front();
+}
+
+void ConsumerTableBase::pop(std::string &key, std::string &op, std::vector<FieldValueTuple> &fvs, std::string prefix)
```
qiluo-msft: >pop [](start = 24, length = 3)

This implementation has unnecessary copy of string/vector objects. Actually, we could implement this one directly, and let previous pop(kco) call it.

lguohan: implemented directly.

qiluo-msft: auto& kco = m_buffer.front();
// copy out
m_buffer.pop_front();

---
In reply to: [147793047](https://github.com/Azure/sonic-swss-common/pull/154#discussion_r147793047) [](ancestors = 147793047)

PR #156, opened 2017-11-02
Add getTimestamp() common function

JipanYanga:

This is to move the getTimestamp() function defined in swss saihelper.cpp and sai_redis_record.cpp to swss-common to avoid duplications.   No change to the function code itself.

Users of getTimestamp() also include Orch class in swss.


lguohan:
@qiluo-msft  to review

PR #158, opened 2017-11-10
Move QoS configuration to config DB

andriymoroz-mlnx:
Signed-off-by: Andriy Moroz <c_andriym@mellanox.com>

See also:
https://github.com/Azure/sonic-buildimage/pull/1162
https://github.com/Azure/sonic-swss/pull/378


yxieca:
Discussed with Shuotian, this review doesn't have to be broken into 2.

However, #1162 needs to include submodule updates to avoid build failure.

PR #159, opened 2017-11-11
add python binding for swss::Table

lguohan:
also rename getTableKeys -> getKeys, getTableContent -> getContent


lguohan:
getTableKeys and getTableContents is not used outside swss-common project yet, so there is no changes in other repo.

PR #160, opened 2017-11-12
Add SelectableTimer class

qiluo-msft:


common/selectabletimer.h
```diff
@@ -0,0 +1,26 @@
...
+class SelectableTimer : public Selectable
+{
+public:
+    SelectableTimer(const timespec& interval);
+    virtual ~SelectableTimer();
+
+    virtual void addFd(fd_set *fd);
+    virtual bool isMe(fd_set *fd);
+    virtual int readCache();
+    virtual void readMe();
```
lguohan: should we add a method to change the interval so that user can change the timer period?

lguohan: I feel we need reset() method so that we can reset the timer right before we call select. currently, your timer begins right after it gets created. there maybe time elapse between your creation and select.

qiluo-msft: Do you actually mean start() and stop() instead of reset()?

common/selectabletimer.cpp
```diff
@@ -0,0 +1,102 @@
...
+namespace swss {
+
+SelectableTimer::SelectableTimer(const timespec& interval)
+    : m_zero({{0, 0}, {0, 0}})
+{
+    // Create the timer
+    m_tfd = timerfd_create(CLOCK_REALTIME, 0);
+    if (m_tfd == -1)
+    {
+        SWSS_LOG_ERROR("failed to create timerfd, errno: %s", strerror(errno));
```
kcudnik: you can use SWSS_LOG_THROW which will log error and throw the same message

common/selectabletimer.h
lguohan: timer.reset() is the combination of stop and start.

PR #171, opened 2018-01-03
[MacAddress]: Optimize to_string() and parseMacString methods

pavel-shirshov:
MacAddress class instantiated many times from string. Also we parse mac addresses a lot. I did a benchmark and I found that our old version of to_string() and parseMacString() methods not so fast.

I've rewritten both of them. Now both methods faster (parseMacString - 76 times faster, almost 10 times faster) and parseMacString() catch more errors.

```
2018-01-03 23:01:53
Run on (16 X 2394.44 MHz CPU s)
CPU Caches:
  L1 Data 32K (x16)
  L1 Instruction 32K (x16)
  L2 Unified 256K (x16)
  L3 Unified 30720K (x16)
------------------------------------------------------------
Benchmark                     Time           CPU Iterations
------------------------------------------------------------
BM_ParseMacOld              840 ns        840 ns     842880
BM_ParseMac                  11 ns         11 ns   62703966
BM_MAC_to_string_old        496 ns        496 ns    1448299
BM_MAC_to_string             50 ns         50 ns   13564922
```

### DIFF
--- a/common/macaddress.cpp
+++ b/common/macaddress.cpp
@@ -24,41 +24,97 @@ MacAddress::MacAddress(const std::string& macStr)
 
 const std::string MacAddress::to_string() const
 {
-    char tmp_mac[32];
-    sprintf(tmp_mac, "%02x:%02x:%02x:%02x:%02x:%02x", m_mac[0], m_mac[1], m_mac[2], m_mac[3], m_mac[4], m_mac[5]);
-    return std::string(tmp_mac);
+    return MacAddress::to_string(m_mac);
 }
 
 std::string MacAddress::to_string(const uint8_t* mac)
 {
-    uint8_t tmp_mac[32];
-    sprintf((char*)tmp_mac, "%02x:%02x:%02x:%02x:%02x:%02x", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
-    return std::string((char*)tmp_mac);
+    std::string str(17, ':');
+    for(int i = 0; i < 6; i++) {
+        int left = i * 3;
+        int right = left + 1;
+        char left_half = static_cast<char>(mac[i] >> 4);
+        char right_half = mac[i] & 0x0f;
+
+        if (left_half >= 0 && left_half <= 9)
+        {
+            str[left] = static_cast<char>(left_half + '0');
+        } else {
+            str[left] = static_cast<char>(left_half + 'a' - 0x0a);
+        }
+
+        if (right_half >= 0 && right_half <= 9)
+        {
+            str[right] = static_cast<char>(right_half + '0');
+        } else {
+            str[right] = static_cast<char>(right_half + 'a' - 0x0a);
+        }
+    }
+
+    return str;
 }
 
-bool MacAddress::parseMacString(const string& macStr, uint8_t* mac)
+bool MacAddress::parseMacString(const string& str_mac, uint8_t* bin_mac)
 {
-    int i;
-    unsigned int value;
-    char ignore;
-    istringstream iss(macStr);
-
-    if (mac == NULL)
+    if (bin_mac == NULL)
     {
         return false;
     }
 
-    iss >> hex;
-
-    for (i = 0; i < 5; i++)
+    if (str_mac.length() != 17)
     {
-        iss >> value >> ignore;
-        if (!iss) return false;
-        if (value >= 256) return false;
-        mac[i] = (uint8_t)value;
+        return false;
     }
-    iss >> value;
-    mac[5] = (uint8_t)value;
+
+    const char* ptr_mac = str_mac.c_str();
+
+    if ((ptr_mac[2]  != ':' || ptr_mac[5]  != ':' || ptr_mac[8] != ':' || ptr_mac[11] != ':' || ptr_mac[14] != ':')
+     && (ptr_mac[2]  != '-' || ptr_mac[5]  != '-' || ptr_mac[8] != '-' || ptr_mac[11] != '-' || ptr_mac[14] != '-'))
+    {
+        return false;
+    }
+
+    for(int i = 0; i < 6; i++)
+    {
+        int left  = i * 3;
+        int right = left + 1;
+
+        if (ptr_mac[left] >= '0' &&  ptr_mac[left] <= '9')
+        {
+            bin_mac[i] = static_cast<uint8_t>(ptr_mac[left] - '0');
+        }
+        else if (ptr_mac[left] >= 'A' &&  ptr_mac[left] <= 'F')
+        {
+            bin_mac[i] = static_cast<uint8_t>(ptr_mac[left] - 'A' + 0x0a);
+        }
+        else if (ptr_mac[left] >= 'a' &&  ptr_mac[left] <= 'f')
+        {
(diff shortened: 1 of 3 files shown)


common/macaddress.cpp
```diff
@@ -24,41 +24,97 @@ MacAddress::MacAddress(const std::string& macStr)
...
     {
         return false;
     }
 
-    iss >> hex;
+    const char* ptr_mac = str_mac.c_str();
 
-    for (i = 0; i < 5; i++)
+    if ((ptr_mac[2] != ':' || ptr_mac[5] != ':' || ptr_mac[8] != ':' || ptr_mac[11] != ':' || ptr_mac[14] != ':')
+     && (ptr_mac[2] != '-' || ptr_mac[5] != '-' || ptr_mac[8] != '-' || ptr_mac[11] != '-' || ptr_mac[14] != '-'))
```
qiluo-msft: Hard to read.

common/macaddress.cpp
```diff
@@ -24,41 +24,97 @@ MacAddress::MacAddress(const std::string& macStr)
...
-    unsigned int value;
-    char ignore;
-    istringstream iss(macStr);
+    if (bin_mac == NULL)
+    {
+        return false;
+    }
 
-    if (mac == NULL)
+    if (str_mac.length() != 17)
```
qiluo-msft: >17 [](start = 28, length = 2)

magic number. 6*2 + 5?

common/macaddress.cpp
```diff
@@ -24,41 +24,97 @@ MacAddress::MacAddress(const std::string& macStr)
...
+     && (ptr_mac[2] != '-' || ptr_mac[5] != '-' || ptr_mac[8] != '-' || ptr_mac[11] != '-' || ptr_mac[14] != '-'))
     {
-        iss >> value >> ignore;
-        if (!iss) return false;
-        if (value >= 256) return false;
-        mac[i] = (uint8_t)value;
+        return false;
+    }
+
+    for(int i = 0; i < 6; i++)
```
qiluo-msft: >for [](start = 4, length = 3)

reuse strtol?

common/macaddress.cpp
```diff
@@ -24,41 +24,97 @@ MacAddress::MacAddress(const std::string& macStr)
...
-    sprintf((char*)tmp_mac, "%02x:%02x:%02x:%02x:%02x:%02x", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
-    return std::string((char*)tmp_mac);
+    std::string str(17, ':');
+    for(int i = 0; i < 6; i++) {
+        int left = i * 3;
+        int right = left + 1;
+        char left_half = static_cast<char>(mac[i] >> 4);
+        char right_half = mac[i] & 0x0f;
+
+        if (left_half >= 0 && left_half <= 9)
```
qiluo-msft: >if [](start = 8, length = 2)

Use a static mapping (char->int) to make it even faster?

common/macaddress.cpp
pavel-shirshov: I don't know how I can make it more readable. Add const del_pos_1 = 2, del_pos_2 = 5 and so on? I doubt it is more readable then now.
What would you suggest?

common/macaddress.cpp
pavel-shirshov: added constant for the number

common/macaddress.cpp
pavel-shirshov: 1. The performance of strtol isn't good.
2. In case strtol read an invalid value it returns 0. So it's impossible to distinct between an error and 0.
Conversion of hexadecimal number from string  to a binary easy. So this is a reason I write everything in the function and don't use standard functions.

common/macaddress.cpp
pavel-shirshov: I don't know how an array could help here. Can you please elaborate?

common/macaddress.cpp
qiluo-msft:
```
// ref: [https://stackoverflow.com/a/8198279/2514803](https://stackoverflow.com/a/8198279/2514803)
template <typename T, typename U>
bool allequal(const T &t, const U &u) {
    return t == u;
}

template <typename T, typename U, typename... Others>
bool allequal(const T &t, const U &u, Others const &... args) {
    return (t == u) && allequal(u, args...);
}

if (!allequal(ptr_mac[2], ptr_mac[5], ptr_mac[8], ptr_mac[11], ptr_mac[14]))
    return false;
if (!(ptr_mac[2] == ':' || ptr_mac[2] == '-'))
    return false;

```

---
In reply to: [159558946](https://github.com/Azure/sonic-swss-common/pull/171#discussion_r159558946) [](ancestors = 159558946)

common/macaddress.cpp
qiluo-msft:
```
const static char map[] = "0123456789abcdef";
str[left] = map[left_half];
```


---
In reply to: [159559465](https://github.com/Azure/sonic-swss-common/pull/171#discussion_r159559465) [](ancestors = 159559465)

pavel-shirshov: Thanks Qi. Now it looks better.
But performance almost the same:
```
2018-01-04 02:26:25
Run on (16 X 2394.44 MHz CPU s)
CPU Caches:
  L1 Data 32K (x16)
  L1 Instruction 32K (x16)
  L2 Unified 256K (x16)
  L3 Unified 30720K (x16)
------------------------------------------------------------
Benchmark                     Time           CPU Iterations
------------------------------------------------------------
BM_ParseMacOld              874 ns        874 ns     798878
BM_ParseMac                  13 ns         13 ns   53531585
BM_MAC_to_string_old        467 ns        467 ns    1498719
BM_MAC_to_string             47 ns         47 ns   14613664
BM_MAC_to_string_map         46 ns         46 ns   15300858
```

common/macaddress.cpp
```diff
@@ -1,64 +1,112 @@
...
 {
-    memcpy(m_mac, mac, 6);
+    memcpy(m_mac, mac, ETHER_ADDR_LEN);
 }
 
 MacAddress::MacAddress(const std::string& macStr)
 {
     bool suc = MacAddress::parseMacString(macStr, m_mac);
-    if (!suc) throw invalid_argument("macStr");
+    if (!suc) throw invalid_argument("can't parse mac address '" + macStr + "'");
```
qiluo-msft: >invalid_argument("can't parse mac address '" + macStr + "'") [](start = 20, length = 60)

Actually invalid_argument("macStr")) is good enough

pavel-shirshov: If you check other classes in "swss-common" you'll see we return more descriptive exceptions there.

pavel-shirshov: Also I think it's better to show the reason of the exception

common/macaddress.cpp
pavel-shirshov: For me this code even harder to read. In my code only numbers are cryptic, here we have the same numbers, but also some additional operation which adds the complexity.

qiluo-msft: Even you don't need template version of allequal, you could do hard-coded allequal once, and reduce the number of comparison 10 -> 6.

pavel-shirshov: I've added the code you provided. The benchmark doesn't show differences between your and my version. I've added comments to make the code clearer.

PR #175, opened 2018-01-12
Replace long long with int64_t

kcudnik:


common/redisreply.cpp
```diff
@@ -139,7 +139,7 @@ void RedisReply::checkStatusQueued()
     checkStatus("QUEUED");
 }
 
-template<> long long int RedisReply::getReply<long long int>()
```
qiluo-msft: Actually hiredis is using 'long long'.
https://github.com/redis/hiredis/blob/29cb95f4f67879a20ee784bdd4f6bd5b2105eb4d/hiredis.h#L114

What is the reason?

kcudnik: yes, already noticed, and fighting with pragmas, i will drop this for now, since swss would need to be updated too

PR #176, opened 2018-01-15
Add RedisClient to pyext

kcudnik:
Required for pyext to get access to VIDTORID table

### DIFF
--- a/common/redisclient.h
+++ b/common/redisclient.h
@@ -32,23 +32,23 @@ class RedisClient
 
         std::vector<std::string> keys(std::string key);
 
-        std::vector<std::string> hkeys(std::string key);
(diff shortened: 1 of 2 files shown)


common/redisclient.h
```diff
@@ -32,23 +32,23 @@ class RedisClient
 
         std::vector<std::string> keys(std::string key);
 
-        std::vector<std::string> hkeys(std::string key);
+        //std::vector<std::string> hkeys(std::string key);
```
lguohan: why do we comment them?

kcudnik: swig is complaining that there is no implementation for those methods since i just left signature for those mthods, but there is no implementation yet

PR #181, opened 2018-01-25
Table class supports redis pipeline

qiluo-msft:


tests/redis_piped_ut.cpp
```diff
@@ -326,7 +326,8 @@ TEST(Table, piped_test)
 {
     string tableName = "TABLE_UT_TEST";
     DBConnector db(TEST_VIEW, "localhost", 6379, 0);
-    Table t(&db, tableName);
+    RedisPipeline pipeline(&db);
+    Table t(&pipeline, tableName, DEFAULT_TABLE_NAME_SEPARATOR, true);
```
lguohan: @sihuihan88 , you need follow this example to enable redispipeline in your scenario.

common/table.h
```diff
@@ -112,10 +113,11 @@ class TableEntryEnumerable {
     void getContent(std::vector<KeyOpFieldsValuesTuple> &tuples);
 };
 
-class Table : public RedisTransactioner, public TableBase, public TableEntryEnumerable {
+class Table : public TableBase, public TableEntryEnumerable {
 public:
     Table(DBConnector *db, std::string tableName, std::string tableSeparator = DEFAULT_TABLE_NAME_SEPARATOR);
-    virtual ~Table() { }
+    Table(RedisPipeline *pipeline, std::string tableName, std::string tableSeparator, bool buffered);
+    virtual ~Table();
```
lguohan: why remove the implementation for the Table deconstructor?

qiluo-msft: Actually not removed. I implemented with non-empty logic in .cpp file

PR #184, opened 2018-01-27
Extending IpPrefix class to return the subnet associated to any given prefix

rodnymolina:
This method is being currently utilized by the new interface's overlapping-detection logic (to be submitted under a different PR within sonic-swss repo). Method's goal is to simply return the subnet associated to any given prefix (see example below). A new pytest testcase will be appended shortly.

Examples:

subnet("10.1.1.1/24") = 10.1.1.0/24
subnet("fc00:0003:0002:0002::2/64") = fc00:0003:0002:0002:0000:0000:0000:0000/64

### DIFF
--- a/common/ipprefix.h
+++ b/common/ipprefix.h
@@ -3,6 +3,7 @@
 
 #include <assert.h>
 #include <stdexcept>
+#include <arpa/inet.h>
 #include "ipaddress.h"
 
 namespace swss {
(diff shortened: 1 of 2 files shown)


lguohan:
please add unit test.

rodnymolina:
UT's created. Please review when have a chance.

PR #186, opened 2018-02-07
[logger] Support logging to stdout/stderr

ishidawataru:
logging to stdout/stderr simplifies deploying sonic daemons as containers.

### DIFF
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -5,6 +5,8 @@
 #include <stdio.h>
 #include <syslog.h>
 #include <fstream>
+#include <sstream>
+#include <iomanip>
 #include <stdexcept>
 #include <schema.h>
 #include <select.h>
@@ -16,8 +18,8 @@
 namespace swss {
 
 Logger::~Logger() {
-    if (m_prioThread) {
-        m_prioThread->detach();
+    if (m_settingThread) {
+        m_settingThread->detach();
     }
 }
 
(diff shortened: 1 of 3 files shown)


common/logger.cpp
```diff
@@ -41,48 +43,96 @@ void Logger::swssNotify(std::string component, std::string prioStr)
         SWSS_LOG_ERROR("Invalid loglevel. Setting to NOTICE.");
         logger.m_minPrio = SWSS_NOTICE;
     }
+    else
+    {
+        logger.m_minPrio = priorityStringMap.at(prioStr);
```
kcudnik: was that a bug ? that this is here now ?

ishidawataru: yes it was a bug.

When `prioStr` is invalid, an `out_of_range` exception will be thrown at the deleted line below.

```diff
-    logger.m_minPrio = priorityStringMap.at(prioStr);
```

common/logger.cpp
```diff
@@ -160,10 +214,29 @@ void Logger::write(Priority prio, const char *fmt, ...)
     // TODO
     // + add thread id using std::thread::id this_id = std::this_thread::get_id();
     // + add timestmap with millisecond resolution
-
     va_list ap;
     va_start(ap, fmt);
-    vsyslog(prio, fmt, ap);
+
+    if (m_output == SWSS_SYSLOG)
```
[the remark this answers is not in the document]
ishidawataru: Hmm, do you think that behavior is intuitive for users?

My use case for stdout/stderr logging is deploying sonic daemons in containers.
In container world, it is common for applications to just log to stdout/stderr.

https://kubernetes.io/docs/concepts/cluster-administration/logging/

In this case, we don't need to worry about losing logs by closing the terminal since the container orchestrator stores it internally.

You can always check logs by

`docker logs <container>` or `kubectl logs <pod> <container>`

ishidawataru:
@kcudnik I answered to your comments. Could you take a look?

ishidawataru:
@kcudnik @lguohan ping

PR #190, opened 2018-03-22
Add map of database ID -> table name separator; No longer pass table name separator when constructing Table objects

jleveque:
Previously, SubscriberStateTable was hardcoded to use `CONFIGDB_TABLE_NAME_SEPARATOR`, which meant that one could only subscribe to notifications from databases which used "|" as a table name separator. If you subscribed to a database which used ":" as a table name separator, you would never receive notifications.

This change creates a constant map of database ID to table name separator, and chooses the appropriate table name separator for a table based on the DB one is connecting to. While fixing the above problem, I also simplified the method calls for instantiating Table objects, as one no longer needs to pass in the table separator. This will also prevent table name separator mismatches in the future.

Also changed name of `DEFAULT_TABLE_NAME_SEPARATOR` to `LEGACY_TABLE_NAME_SEPARATOR` and `CONFIGDB_TABLE_NAME_SEPARATOR` to `NEW_TABLE_NAME_SEPARATOR`

Also define `TEST_DB` at index 15, the largest index supported by default Redis config, for use in unit tests.

NOTE: This PR will require related changes in sonic-swss, sonic-sairedis and sonic-buildimage repos. I will submit those PRs shortly.

### DIFF
--- a/common/dbconnector.h
+++ b/common/dbconnector.h
@@ -20,13 +20,13 @@ public:
      * Timeout - The time in milisecond until exception is been thrown. For
      *           infinite wait, set this value to 0
      */
-    DBConnector(int db, std::string hostname, int port, unsigned int timeout);
-    DBConnector(int db, std::string unixPath, unsigned int timeout);
+    DBConnector(int dbId, std::string hostname, int port, unsigned int timeout);
+    DBConnector(int dbId, std::string unixPath, unsigned int timeout);
 
     ~DBConnector();
 
     redisContext *getContext();
-    int getDB();
+    int getDbId();
 
     static void select(DBConnector *db);
 
@@ -35,7 +35,7 @@ public:
 
 private:
     redisContext *m_conn;
-    int m_db;
+    int m_id;
 };
 
 }
--- a/common/dbconnector.h
+++ b/common/dbconnector.h
@@ -35,7 +35,7 @@ public:
 
 private:
     redisContext *m_conn;
-    int m_id;
+    int m_dbId;
 };
 
 }
--- a/common/redispipeline.h
+++ b/common/redispipeline.h
@@ -94,7 +94,7 @@ public:
 
     int getDbId()
     {
-        return m_db == NULL ? -1 : m_db->getDbId();
+        return m_db->getDbId();
     }
 
 private:
--- a/common/redispipeline.h
+++ b/common/redispipeline.h
@@ -92,6 +92,11 @@ public:
         return m_remaining;
     }
 
+    DBConnector *getDbConnector()
+    {
+        return m_db;
+    }
+
 private:
     DBConnector *m_db;
     std::queue<int> m_expectedTypes;
--- a/common/redispipeline.h
+++ b/common/redispipeline.h
@@ -92,9 +92,9 @@ public:
         return m_remaining;
     }
 
-    DBConnector *getDbConnector()
+    int getDbId()
     {
-        return m_db;
+        return m_db == NULL ? -1 : m_db->getDbId();
     }
 
 private:
--- a/common/schema.h
+++ b/common/schema.h
@@ -13,6 +13,7 @@ namespace swss {
 #define PFC_WD_DB       5
 #define FLEX_COUNTER_DB 5
 #define STATE_DB        6
+#define TEST_DB         15
 
 /***** APPLICATION DATABASE *****/
 
--- a/common/schema.h
+++ b/common/schema.h
@@ -13,7 +13,6 @@ namespace swss {
 #define PFC_WD_DB       5
 #define FLEX_COUNTER_DB 5
 #define STATE_DB        6
-#define TEST_DB         15
 
 /***** APPLICATION DATABASE *****/
 
--- a/common/table.h
+++ b/common/table.h
@@ -18,8 +18,11 @@
 
 namespace swss {
 
-#define DEFAULT_TABLE_NAME_SEPARATOR    ":"
-#define CONFIGDB_TABLE_NAME_SEPARATOR   "|"
+#define LEGACY_TABLE_NAME_SEPARATOR  ":"
+#define NEW_TABLE_NAME_SEPARATOR     "|"
+
+// Mapping of DB ID to table name separator string
+typedef std::map<int, std::string> TableNameSeparatorMap;
 
 typedef std::pair<std::string, std::string> FieldValueTuple;
 #define fvField std::get<0>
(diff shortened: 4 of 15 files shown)


qiluo-msft:
Please check the unit test failure.

common/redispipeline.h
```diff
@@ -92,6 +92,11 @@ class RedisPipeline {
         return m_remaining;
     }
 
+    DBConnector *getDbConnector()
```
qiluo-msft: >getDbConnector [](start = 17, length = 14)

The m_db is for internal use by design. It is carefully appended and popped to make pipeline working. Do not let outside use interleaves with internal use.

common/schema.h
```diff
@@ -13,6 +13,7 @@ namespace swss {
 #define PFC_WD_DB       5
 #define FLEX_COUNTER_DB 5
 #define STATE_DB        6
+#define TEST_DB         15
```
qiluo-msft: >15 [](start = 24, length = 2)

Is there a redis constant we could refer?

common/table.h
```diff
@@ -18,8 +18,11 @@
 
 namespace swss {
 
-#define DEFAULT_TABLE_NAME_SEPARATOR    ":"
-#define CONFIGDB_TABLE_NAME_SEPARATOR   "|"
+#define LEGACY_TABLE_NAME_SEPARATOR  ":"
+#define NEW_TABLE_NAME_SEPARATOR     "|"
```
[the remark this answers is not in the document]
qiluo-msft: Suggest to move into class constant member, and add a getter function.

---
In reply to: [176601615](https://github.com/Azure/sonic-swss-common/pull/190#discussion_r176601615) [](ancestors = 176601615)

common/table.h
```diff
@@ -18,8 +18,11 @@
...
+
+// Mapping of DB ID to table name separator string
+typedef std::map<int, std::string> TableNameSeparatorMap;
```
qiluo-msft: >TableNameSeparatorMap [](start = 35, length = 21)

suggest to keep it as a class private member.

common/redispipeline.h
jleveque: Latest commit now exposes the DB ID of m_db, not m_db itself.

qiluo-msft: m_db should never be NULL. you may use assert, if you want.

---
In reply to: [176623397](https://github.com/Azure/sonic-swss-common/pull/190#discussion_r176623397) [](ancestors = 176623397)

jleveque: Removed NULL check.

common/schema.h
jleveque: No constant that I could find, but I found that you can query the number of databases using the `CONFIG GET databases`:

```
# redis-cli CONFIG GET databases
1) "databases"
2) "16"
```

However, with this info, we can't define a constant like `TEST_DB`. We would need a getter method which returns the test DB id as (num_databases -1).

I still believe defining `TEST_DB` here with the other DB IDs is an improvement over the previous solution of `#define TEST_VIEW            (7)` in tests/redis_ut.cpp. Having to remember to increment that value when adding a new DB is onerous.

common/table.h
```diff
@@ -18,8 +18,11 @@
 
 namespace swss {
 
-#define DEFAULT_TABLE_NAME_SEPARATOR    ":"
-#define CONFIGDB_TABLE_NAME_SEPARATOR   "|"
+#define LEGACY_TABLE_NAME_SEPARATOR  ":"
```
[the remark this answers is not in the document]
jleveque: I agree we should migrate completely to using only one separator (new) everywhere. I'm not sure of the total amount of work required for that. This is a question for @lguohan.

However, the changes I've made in this PR will help make that transition easier, by greatly reducing the number of references that will need to be updated.

common/schema.h
qiluo-msft: Originally TEST_VIEW is only used in unit test. So please leave any related constant there.

---
In reply to: [176849657](https://github.com/Azure/sonic-swss-common/pull/190#discussion_r176849657) [](ancestors = 176849657)

jleveque: It needs to be known here because it needs to be included in the table name separator map. Otherwise, the TableBase constructor will throw an exception, unless we simply use an existing DB for unit tests.

qiluo-msft: As we discussed, other developers may need the freedom to experiment in new DB id without it's allocated formally in swss-common library. So the requirement for out-of-range DB id should be supported.

kcudnik:
This will probably break a lot of stuff in sairedis since sairedis is using hardcoded table separator also, but as string, for example "Table:*" when its lists all the tables, and it may be not easy task to fix all those places
maybe tests will catch some of those which will be used

jleveque:
@kcudnik: This PR shouldn't cause any issues with existing hardcoded strings containing a colon as a separator (like `Table:*`), as I haven't changed which DBs use which separators. However, those _will_ be an issue once we decide to change all separators across all DBs to use the pipe character instead. It sounds like that will be a big job.

FYI, there was only one reference in sairedis that explicitly referenced `DEFAULT_TABLE_NAME_SEPARATOR`, and I have removed that to work with this PR here: https://github.com/Azure/sonic-sairedis/pull/309/files

common/dbconnector.h
```diff
@@ -35,7 +35,7 @@ class DBConnector
 
 private:
     redisContext *m_conn;
-    int m_db;
+    int m_id;
```
lguohan: rename it as m_dbid to align with your function name getDbId()

lguohan:
how do you plan to deal with swsssdk the python version? If we incorporate your changes, do we need any change in swsssdk or we are good?

lguohan:
I think it is better merge this after March release.

lguohan:
in current db, do we have cases that use different separator in one table db? If not, i think this approach is ok.

jleveque:
@lguohan:

> how do you plan to deal with swsssdk the python version? If we incorporate your changes, do we need any change in swsssdk or we are good?

This change should have no impact on swsssdk, as swsssdk doesn't reference swss-common. It uses the Python Redis library directly, and thus is completely independent of swss-common.

However, this change will make it possible to use the new Python bindings of swss-common everywhere we currently use swsssdk, effectively allowing us to deprecate and eventually remove swsssdk altogether.

> in current db, do we have cases that use different separator in one table db? If not, i think this approach is ok.

From my investigation, it appears as though there is no mixing of separators within any DBs. Each database consistently uses only one separator.

PR #192, opened 2018-04-03
Refactor Selectables. Add support for selectable priorities

pavel-shirshov:
**PLEASE DON'T PUSH THIS CHANGE. IT'LL BREAK BOTH SWSS AND SAIREDIS.**

This refactoring introduce support of priorities for Selectables. Selectables with highest priorities will be selected first. If we have two selectables with the same priorities, the selectable which was accessed earliest will be chosen.

The priorities are defined as last argument of type int for following classes:
- ConsumerStateTable (default priority is 0)
- ConsumerTable (default priority is 0)
- Netlink (default priority is 0)
- NotificationConsumer (default priority is 100)
- RedisSelect (default priority is 0)
- SelectableEvent (default priority is 0)
- SelectableTimer (default priority is 50)
- SubscriberStateTable (default priority is 0)

To support this feature Selectable interface was changed significantly.
Now Selectable has to provide following methods:
- getFd() - return file descriptor for the Selectable
- readData() - read all data in Selectable. (The data will be stored internally in selectable)
- hasCachedData() - return true, if there're data for the next read (don't remove Selectable from the queue is there're the data)
- initializedWithData() - return true, if Constructor of selectable read some data inside of it, so we can return it as ready on the next select().
 - updateAfterRead() - this method will be run when select return this Selectable from itself.

**The method Select::select() was changed. int* fd parameter was remove as not used anywhere**

I switched from Unix select() to Linux epoll which is more effective and easy to use.

Also I found that SelectableTimer could lose it's events, if reading rate is greater then timer interval. Probably it's a bug.

Also I made some other small changes and tweaks to improve performance of the swss-common. I'm going to create a separate PR with more performance changes for swss-common.

### DIFF
--- a/common/ipaddress.h
+++ b/common/ipaddress.h
@@ -20,7 +20,7 @@ class IpAddress
 {
 public:
     IpAddress() {}
-    IpAddress(const ip_addr_t &ip) { m_ip = ip; }
+    IpAddress(const ip_addr_t &ip) : m_ip(ip) {}
     IpAddress(uint32_t ip);
     IpAddress(const std::string &ipStr);
 
--- a/common/ipaddress.h
+++ b/common/ipaddress.h
@@ -20,7 +20,7 @@ class IpAddress
 {
 public:
     IpAddress() {}
-    IpAddress(const ip_addr_t &ip) : m_ip(ip) {}
+    IpAddress(const ip_addr_t &ip) { m_ip = ip; }
     IpAddress(uint32_t ip);
     IpAddress(const std::string &ipStr);
 
--- a/common/select.cpp
+++ b/common/select.cpp
@@ -81,7 +81,7 @@ void Select::removeSelectable(Selectable *selectable)
 
 void Select::addSelectables(vector<Selectable *> selectables)
 {
-    for(auto it : selectables)
+    for(const auto& it : selectables)
     {
         addSelectable(it);
     }
--- a/common/select.cpp
+++ b/common/select.cpp
@@ -81,7 +81,7 @@ void Select::removeSelectable(Selectable *selectable)
 
 void Select::addSelectables(vector<Selectable *> selectables)
 {
-    for(const auto& it : selectables)
+    for(auto it : selectables)
     {
         addSelectable(it);
     }
--- a/common/select.cpp
+++ b/common/select.cpp
@@ -5,24 +5,78 @@
 #include <stdio.h>
 #include <sys/time.h>
 #include <sys/types.h>
+#include <sys/epoll.h>
+#include <unistd.h>
+#include <string.h>
 
 using namespace std;
 
 namespace swss {
 
+Select::Select()
+{
+    m_epoll_fd = ::epoll_create1(0);
+    if (m_epoll_fd == -1)
+    {
+        std::string error = std::string("Select::constructor:epoll_create1: error=("
+                          + std::to_string(errno) + "}:"
+                          + strerror(errno));
+        throw std::runtime_error(error);
+    }
+}
+
+Select::~Select()
+{
+    (void)::close(m_epoll_fd);
+}
+
 void Select::addSelectable(Selectable *selectable)
 {
-    if(find(m_objects.begin(), m_objects.end(), selectable) != m_objects.end())
+    const int fd = selectable->getFd();
+
+    if(m_objects.find(fd) != m_objects.end())
     {
         SWSS_LOG_WARN("Selectable is already added to the list, ignoring.");
         return;
     }
-    m_objects.push_back(selectable);
+
+    m_objects[fd] = selectable;
+
+    if (selectable->initializedWithData())
+    {
+        m_ready.insert(selectable);
+    }
+
+    struct epoll_event ev = {
+        .events = EPOLLIN,
+        .data = { .fd = fd, },
+    };
+
+    int res = ::epoll_ctl(m_epoll_fd, EPOLL_CTL_ADD, fd, &ev);
+    if (res == -1)
+    {
+        std::string error = std::string("Select::add_fd:epoll_ctl: error=("
+                          + std::to_string(errno) + "}:"
+                          + strerror(errno));
+        throw std::runtime_error(error);
+    }
 }
 
-void Select::removeSelectable(Selectable *c)
+void Select::removeSelectable(Selectable *selectable)
 {
-    m_objects.erase(remove(m_objects.begin(), m_objects.end(), c), m_objects.end());
+    const int fd = selectable->getFd();
+
+    m_objects.erase(fd);
+    m_ready.erase(selectable);
+
+    int res = ::epoll_ctl(m_epoll_fd, EPOLL_CTL_DEL, fd, NULL);
+    if (res == -1)
+    {
+        std::string error = std::string("Select::del_fd:epoll_ctl: error=("
+                          + std::to_string(errno) + "}:"
+                          + strerror(errno));
+        throw std::runtime_error(error);
+    }
 }
 
 void Select::addSelectables(vector<Selectable *> selectables)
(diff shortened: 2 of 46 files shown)


common/ipaddress.h
```diff
@@ -20,7 +20,7 @@ class IpAddress
 {
 public:
     IpAddress() {}
-    IpAddress(const ip_addr_t &ip) { m_ip = ip; }
+    IpAddress(const ip_addr_t &ip) : m_ip(ip) {}
```
qiluo-msft: >m_ip [](start = 37, length = 4)

Could you separate simple refactoring code into a smaller PR? so we could focus on more difficult one.

common/select.h
```diff
@@ -12,29 +15,32 @@ namespace swss {
 class Select
 {
 public:
+    Select();
+    ~Select();
+
     /* Add object for select */
     void addSelectable(Selectable *selectable);
```
qiluo-msft: >addSelectable [](start = 9, length = 13)

suggest move all priority setting as addSelectable parameter, so no need to modify all the related classes.

qiluo-msft:
You mentioned "To support this feature Selectable interface was changed significantly."
Could you explain the detail? Is it really needed to refactor Selectable interface?

common/ipaddress.h
pavel-shirshov: I can, but it will take more time to extract all the changes and revert them back. Can you please consider to check them? They're tiny and follows just a few easy patterns:
```
func(object obj) -> func(const object& obj)
```
,
```
for(auto obj: object) -> for (const auto& obj: object)
```
'''
Class1::Class1(int var1)
{
  m_var_1 = var1;
}
to
Class1::Class1(int var1) : m_var_1(var1) {}
```

File: common/select.h (hunk 2 above)
Comment (pavel-shirshov) [reply common/select.h]: The classes are already changed. These changes won't affect any client code.
Also, I don't like to add information about priority into Select class.
Following the principle of Single Responsibility I want Select class only work for polling, not for something else.
Also, it's not easy to change Select::addSelectables() method which expects vector of Selectables.
Also, is makes Select class more complicated, we need to store and operate on the priority, which is Selectable attribute.

**Comment by pavel-shirshov** [comment]
I must change Selectable interface because:
Previous interface was:
```
  /* Implements FD_SET */
    virtual void addFd(fd_set *fd) = 0;
    virtual bool isMe(fd_set *fd) = 0;
    /* Read and empty socket caching (if exists) */
    virtual int readCache() = 0;
    /* Read a message from the socket */
    virtual void readMe() = 0;

```
In this interface: readMe() allowed us to read information from Selectable which file descriptor signaled.
readCache() allowed us to read information from the cache a Selectable regardless of the file descriptor.
In the previous implementation we first read all cached from all selectables. As soon as they are empty we start reading from first signaled file descriptors. If the file descriptor populated cache, we read that cache on the next iteration.
If we want to introduce priority we must check the file descriptor on every iteration. We can't read cache value until we have it. We can have higher priority information in the sockets awaiting for us.

Also I changed the previous Selectable interface because it leaked Select class methods into Selectables. 
Selectables should know nothing about how Select is implemented. So isMe() and addFd() is leaked abstractions. Also the old interface had no way to remove selectable from the queue.

File: common/ipaddress.h (hunk 1 above)
Comment (qiluo-msft) [reply common/ipaddress.h]: I am thinking 'selectable improvement' as an experiment feature (not mature now). And other simple one are definitely good contribution to existing codebase. Considering the possibility of future reverting and/or conflicting, it's good to separate. Large amount of files in a PR always harm.

---
In reply to: [178991122](https://github.com/Azure/sonic-swss-common/pull/192#discussion_r178991122) [](ancestors = 178991122)

File: common/select.h (hunk 2 above)
Comment (qiluo-msft) [reply common/select.h]: Got your idea. Let me correct my comment. I suggest move all priority setting as function call to Selectable (not Select), keep original constructor signature, so no need to modify all the related sub classes. Another benefit is that all previous code with default priority will keep previous behavior. If any future user would like to apply different priority assignment, he just is able to do so explicitly.

Since all the sub classes has not used pri at all, keep all the modification in base class.

---
In reply to: [178991671](https://github.com/Azure/sonic-swss-common/pull/192#discussion_r178991671) [](ancestors = 178991671)

File: common/select.h [hunk 3]
```diff
@@ -12,29 +15,32 @@ namespace swss {
...
+
+    /* Remove object from select */
     void removeSelectable(Selectable *selectable);
-    void addSelectables(std::vector<Selectable *> selectables);
 
-    /* Add file-descriptor for select  */
-    void addFd(int fd);
+    /* Add multiple messages for select */
```
qiluo-msft: >messages [](start = 20, length = 8)

objects

common/selectable.h
```diff
@@ -11,23 +11,50 @@ namespace swss {
...
-    virtual void addFd(fd_set *fd) = 0;
-    virtual bool isMe(fd_set *fd) = 0;
-
-    /* Read and empty socket caching (if exists) */
-    virtual int readCache() = 0;
-
-    /* Read a message from the socket */
-    virtual void readMe() = 0;
+private:
+    int m_priority;
```
qiluo-msft: >m_priority [](start = 8, length = 10)

Add some comment about the definition. The larger, the higher priority?

common/selectable.h
```diff
@@ -11,23 +11,50 @@ namespace swss {
...
+    {
+        return m_priority;
+    }
+
+    struct cmp
+    {
+        bool operator()(const Selectable* a, const Selectable* b) const
+        {
+            if (a->getPri() == b->getPri())
+                return a > b;
```
qiluo-msft: >a > b [](start = 23, length = 5)

Why define the order this way? The intuitive understanding is they are equal.

common/select.cpp
```diff
@@ -5,105 +5,153 @@
...
-    m_objects.erase(remove(m_objects.begin(), m_objects.end(), c), m_objects.end());
+    const int fd = selectable->getFd();
+
+    m_objects.erase(fd);
+    m_ready.erase(selectable);
+
+    int res = ::epoll_ctl(m_epoll_fd, EPOLL_CTL_DEL, fd, NULL);
+    if (res == -1)
+    {
+        std::string error = std::string("Select::del_fd:epoll_ctl: error=("
```
qiluo-msft: >std:: [](start = 8, length = 5)

No need to add namespace

common/select.cpp
```diff
@@ -5,105 +5,153 @@
...
+    {
+        int fd = events[i].data.fd;
+        Selectable* sel = m_objects[fd];
+        sel->readData();
+        m_ready.insert(sel);
+    }
 
-    /* Checking caching from reader */
-    for (Selectable *i : m_objects)
+    while (!m_ready.empty())
```
qiluo-msft: >while [](start = 4, length = 5)

Since you return in the block, why not use 'if'?

common/select.cpp
```diff
@@ -5,105 +5,153 @@
...
-        {
-            *fd = f;
-            return Select::FD;
-        }
+    *c = NULL;
+    if (timeout == numeric_limits<unsigned int>::max())
+        timeout = -1;
+
+    /* check if we have some data */
+    ret = select1(c, 0);
```
qiluo-msft: >select1 [](start = 10, length = 7)

select1 is a weird name.

common/ipaddress.h
pavel-shirshov: I've removed the other changes. I kept only one bug with the missed va_end()

common/select.h
pavel-shirshov: I don't agree with you. As soon as we create separate method to change priority it will not work as expected. For example, You create a Selectable, then add it into Select object and then change its priority. It will broke Select object, because it didn't expect anyone changes Selectable objects in the queue.
Also it's bad from design viewpoint. The priority property is 'CREATE_AND_SET' property in SAI terms. We don't want to change it as it created.

pavel-shirshov:
Current master has broken unit tests. As soon as they are fixed, I'll resolve conflicts and make some requested changes.

lguohan:
resolve conflict first?

pavel-shirshov:
I've resolved the conflicts, but I have to check everything, until I'm sure everything is alright.

lguohan:
retest this please

pavel-shirshov:
All tests are done. The PR is ready for the final review

common/selectable.h
```diff
@@ -4,30 +4,83 @@
...
+        {
+            /* Choose Selectable with highest priority first */
+            if (a->getPri() > b->getPri())
+                return true;
+            else if (a->getPri() < b->getPri())
+                return false;
+
+            /* if the priorities are equal */
+            /* use Selectable which was selected later */
+            if (a->getLastTimeUsed() < b->getLastTimeUsed())
```
qiluo-msft: >getLastTimeUsed [](start = 19, length = 15)

The order is dynamic, I am afraid it will break the assumption of std::set

pavel-shirshov:
show logging | egrep 'zebra|fpmsyncd'

Qi,

I share your concern about the Comparator using the mutable state. But in this case I don't have any other simple solution. So what I did: 
1. Move the comparator into private section of the Select class. Only this class using this comparator.
2. The set object which uses the Comparator is already in the private section of the Select class.
3. Put updateClock() method into private section of Selectable, nobody could run this methong except friend class Select.

Also I have a comment about danger of using Selectalbe->updateLastUsedTime() on the Selectable which is inside of the set.

PR #198, opened 2018-05-01
Use references when it's possible and other small enhancements.

pavel-shirshov:
1. Use references as loop arguments
2. Initialize class fields in initializer list
3. Don't use std::move for return value
4. Don't use compare method to compare two strings for equality
5. Fix typo

### DIFF
--- a/common/redisapi.h
+++ b/common/redisapi.h
@@ -101,19 +101,18 @@ static inline std::set<std::string> runRedisScript(DBConnector &db, const std::s
         if (ctx->type == REDIS_REPLY_NIL)
         {
             SWSS_LOG_ERROR("Got EMPTY response type from redis %d", ctx->type);
-            return ret;
         }
-
         else if (ctx->type != REDIS_REPLY_ARRAY)
         {
             SWSS_LOG_ERROR("Got invalid response type from redis %d", ctx->type);
-            return ret;
         }
-
-        for (size_t i = 0; i < ctx->elements; i++)
+        else
         {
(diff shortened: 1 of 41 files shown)


common/redisapi.h
```diff
@@ -101,13 +101,13 @@ static inline std::set<std::string> runRedisScript(DBConnector &db, const std::s
         if (ctx->type == REDIS_REPLY_NIL)
```
qiluo-msft: >if [](start = 8, length = 2)

        if (ctx->type == REDIS_REPLY_NIL)
        {
            SWSS_LOG_ERROR("Got EMPTY response type from redis %d", ctx->type);
        }

        else if (ctx->type != REDIS_REPLY_ARRAY)
        {
            SWSS_LOG_ERROR("Got invalid response type from redis %d", ctx->type);
        }
        else
        {
            for (size_t i = 0; i < ctx->elements; i++)
            {
                SWSS_LOG_DEBUG("Got element %lu %s", i, ctx->element[i]->str);
                ret.insert(ctx->element[i]->str);
            }
        }
        return ret;

pavel-shirshov: I don't see any comments here

pavel-shirshov: https://stackoverflow.com/questions/14856344/when-should-stdmove-be-used-on-a-function-return-value

jleveque: I believe Qi is presenting a refactored, simplified block of code to replace the entire block with.

PR #199, opened 2018-05-03
Make code using more C++11 features.

pavel-shirshov:
Use emplace_back to increase performance.

### DIFF
--- a/common/consumertablebase.h
+++ b/common/consumertablebase.h
@@ -12,7 +12,7 @@ public:
 
     ConsumerTableBase(DBConnector *db, const std::string &tableName, int popBatchSize = DEFAULT_POP_BATCH_SIZE, int pri = 0);
 
-    virtual ~ConsumerTableBase() = default;
+    ~ConsumerTableBase() override = default;
 
     void pop(KeyOpFieldsValuesTuple &kco, const std::string &prefix = EMPTY_PREFIX);
 
--- a/common/consumertablebase.h
+++ b/common/consumertablebase.h
@@ -12,7 +12,7 @@ public:
 
     ConsumerTableBase(DBConnector *db, const std::string &tableName, int popBatchSize = DEFAULT_POP_BATCH_SIZE, int pri = 0);
 
-    virtual ~ConsumerTableBase();
+    virtual ~ConsumerTableBase() = default;
 
     void pop(KeyOpFieldsValuesTuple &kco, const std::string &prefix = EMPTY_PREFIX);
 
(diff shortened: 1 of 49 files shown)


common/consumertablebase.h
```diff
@@ -12,7 +12,7 @@ class ConsumerTableBase: public TableConsumable, public RedisTransactioner
 
     ConsumerTableBase(DBConnector *db, const std::string &tableName, int popBatchSize = DEFAULT_POP_BATCH_SIZE, int pri = 0);
 
-    virtual ~ConsumerTableBase();
```
qiluo-msft: >virtual [](start = 4, length = 7)

keep virtual?

common/logger.h
```diff
@@ -107,7 +107,7 @@ class Logger
     };
 
 private:
-    Logger(){};
+    Logger() = default;
```
qiluo-msft: >default [](start = 15, length = 7)

Just wondering if there is any benefit?

common/netdispatcher.h
```diff
@@ -31,8 +31,8 @@ class NetDispatcher {
     void onNetlinkMessage(struct nl_msg *msg);
 
 private:
-    NetDispatcher();
-    ~NetDispatcher();
+    NetDispatcher() = default;
```
qiluo-msft: >default [](start = 22, length = 7)

delete. The previous purpose is to forbid direct construction.

common/netdispatcher.h
```diff
@@ -31,8 +31,8 @@ class NetDispatcher {
...
+    ~NetDispatcher() = default;
```
qiluo-msft: >~NetDispatcher() = default; [](start = 4, length = 27)

Can we just remove it?

common/notificationconsumer.h
```diff
@@ -23,7 +23,7 @@ class NotificationConsumer : public Selectable
 
     void pop(std::string &op, std::string &data, std::vector<FieldValueTuple> &values);
 
-    virtual ~NotificationConsumer();
```
qiluo-msft: >virtual [](start = 4, length = 7)

Keep virtual?

common/netdispatcher.h
pavel-shirshov: Not in this case. The author of the code put cons/destructor under private: By default both constructor/destructor is defined as public:

common/logger.h
pavel-shirshov: My intentions are:
1. Make the code more C++11 compatible
2. Run C++ static analyzers and linters on PRs.

common/netdispatcher.h
qiluo-msft: I think we can remove destructor.

---
In reply to: [185972680](https://github.com/Azure/sonic-swss-common/pull/199#discussion_r185972680) [](ancestors = 185972680)

PR #203, opened 2018-05-26
Reduce the number of LPUSH/LRANGE/LTRIM operations for producer/consumer

jipanyang:

This is to combine the three separate operations for key, op and value into one operation.

With swssconfig to load route to appDB and trigger route download to ASIC, the overall loading speed increased 5~10% in multiple rounds of test.   Don't have the direct speed improvement number for producer/consumer channel only,  it should be better than that number.
   
Since the change is pretty straight forward and simple, and the improvement is obvious, please consider merging this change.

### DIFF
--- a/common/producertable.cpp
+++ b/common/producertable.cpp
@@ -25,6 +25,14 @@ ProducerTable::ProducerTable(RedisPipeline *pipeline, const string &tableName, b
     , m_pipeowned(false)
     , m_pipe(pipeline)
 {
+    /*
+     * KEYS[1] : tableName + "_KEY_VALUE_OP_QUEUE
+     * ARGV[1] : key
+     * ARGV[2] : value
+     * ARGV[3] : op
+     * KEYS[2] : tableName + "_CHANNEL"
+     * ARGV[4] : "G"
+     */
     string luaEnque =
         "redis.call('LPUSH', KEYS[1], ARGV[1], ARGV[2], ARGV[3]);"
         "redis.call('PUBLISH', KEYS[2], ARGV[4]);";
@@ -64,7 +72,7 @@ void ProducerTable::enqueueDbChange(const string &key, const string &value, cons
     command.format(
         "EVALSHA %s 2 %s %s %s %s %s %s",
         m_shaEnque.c_str(),
-        getKeyQueueTableName().c_str(),
+        getKeyValueOpQueueTableName().c_str(),
         getChannelName().c_str(),
         key.c_str(),
         value.c_str(),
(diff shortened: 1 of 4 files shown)


common/producertable.cpp
```diff
@@ -62,12 +60,11 @@ void ProducerTable::setBuffered(bool buffered)
 void ProducerTable::enqueueDbChange(const string &key, const string &value, const string &op, const string& /* prefix */)
 {
     RedisCommand command;
+
     command.format(
-        "EVALSHA %s 4 %s %s %s %s %s %s %s %s",
+        "EVALSHA %s 2 %s %s %s %s %s %s",
         m_shaEnque.c_str(),
         getKeyQueueTableName().c_str(),
```
qiluo-msft: >getKeyQueueTableName [](start = 8, length = 20)

You can the semantics of lists, so the name make no sense here.

common/producertable.cpp
```diff
@@ -26,10 +26,8 @@ ProducerTable::ProducerTable(RedisPipeline *pipeline, const string &tableName, b
     , m_pipe(pipeline)
 {
     string luaEnque =
-        "redis.call('LPUSH', KEYS[1], ARGV[1]);"
-        "redis.call('LPUSH', KEYS[2], ARGV[2]);"
-        "redis.call('LPUSH', KEYS[3], ARGV[3]);"
-        "redis.call('PUBLISH', KEYS[4], ARGV[4]);";
+        "redis.call('LPUSH', KEYS[1], ARGV[1], ARGV[2], ARGV[3]);"
```
qiluo-msft: >"redis.call('LPUSH', KEYS[1], ARGV[1], ARGV[2], ARGV[3]); [](start = 8, length = 57)

Could you add comment about the arguments?

jipanyang: done.

common/table.h
```diff
@@ -175,17 +175,20 @@ class TableName_KeyValueOpQueues {
...
         : m_key(tableName + "_KEY_QUEUE")
         , m_value(tableName + "_VALUE_QUEUE")
         , m_op(tableName + "_OP_QUEUE")
+        , m_keyvalueop(tableName + "_KEY_VALUE_OP_QUEUE")
     {
     }
 
     std::string getKeyQueueTableName() const { return m_key; }
     std::string getValueQueueTableName() const { return m_value; }
     std::string getOpQueueTableName() const { return m_op; }
```
qiluo-msft: Remove those lines?

PR #204, opened 2018-05-30
Improve ConsumerStateTable: del after consumed, and let consumer writes the final table

qiluo-msft:
Original implementation will have 2 issues:
1. The consumer side will accumulate all the state into redis hashes, and each time a new field/value, the consumer will pop it, and return all the filed/value accumulated.
2. The producer side modify the redis hashes directly. To behavior the same way as Producer/ConsumerTable, let the consumer side modify the redis hashes.

### DIFF
--- a/common/consumer_state_table_pops.lua
+++ b/common/consumer_state_table_pops.lua
@@ -8,8 +8,8 @@ for i = 1, n do
    local key = keys[i]
    local fieldvalues = redis.call('HGETALL', stateprefix..tablename..key)
    table.insert(ret, {key, fieldvalues})
-   for f, v in pairs(fieldvalues) do
-      redis.call('HSET', tablename..key, f, v)
+   for i = 1, #fieldvalues, 2 do
+      redis.call('HSET', tablename..key, fieldvalues[i], fieldvalues[i + 1])
    end
    redis.call('DEL', stateprefix..tablename..key)
 end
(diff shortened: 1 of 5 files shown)


qiluo-msft:
@kcudnik I modified the first comment to answer your question.

common/consumer_state_table_pops.lua
```diff
@@ -1,9 +1,16 @@
...
 for i = 1, n do
    local key = keys[i]
-   local values = redis.call('HGETALL', KEYS[2] .. key)
-   table.insert(ret, {key, values})
+   local fieldvalues = redis.call('HGETALL', stateprefix..tablename..key)
+   table.insert(ret, {key, fieldvalues})
+   for f, v in pairs(fieldvalues) do
+      redis.call('HSET', tablename..key, f, v)
+   end
+   redis.call('DEL', stateprefix..tablename..key)
```
jipanyang: With the extra HSET and DEL processing,  it looks the number of redis operations has been doubled. Is the performance impact a concern here?  

Is it possible to compress all the data in sadd argument and process them after SPOP?

qiluo-msft: Since the requirement changes, performance is not the first concern here. I am open to any optimization here.

Regarding compress, the SADD is used to keep keys, so we have chance to merge the operations on the same key. Compressing key/filed/value into SADD doesn't help.

---
In reply to: [191872386](https://github.com/Azure/sonic-swss-common/pull/204#discussion_r191872386) [](ancestors = 191872386)

jipanyang:
Correct data written to DB table after the fieldvalue set fix. thanks!

PR #208, opened 2018-07-17
redisclient exists, hmset, hgetallordered function, warm restart table names and table getDbId() function

jipanyang:
This is to add some swss common function support for warm reboot development:

<1> redisclient exists(), hmset(), hgetallordered() functions.
<2> warm restart table names in appDB and configDB 
<3> getDbId() function for Table class


common/redisclient.cpp
```diff
@@ -20,6 +20,14 @@ int64_t RedisClient::del(const string &key)
     return r.getContext()->integer;
 }
 
+int64_t RedisClient::exists(const string &key)
```
pavel-shirshov: I think it's better to use bool as return value

common/redisclient.cpp
```diff
@@ -20,6 +20,14 @@ int64_t RedisClient::del(const string &key)
...
+{
+    RedisCommand sexists;
+    sexists.format("EXISTS %s", key.c_str());
```
pavel-shirshov: I think It would be faster to concatenate to strings than use .format her.

common/redisclient.cpp
```diff
@@ -35,6 +43,34 @@ void RedisClient::hset(const string &key, const string &field, const string &val
...
+void RedisClient::hmset(const string &key, const std::map<std::string, std::string> &vmap)
+{
+    if (vmap.size() == 0)
+        return;
+
+    vector<FieldValueTuple> values;
+    auto it = vmap.begin();
+    while (it != vmap.end())
+    {
+        values.push_back(FieldValueTuple(it->first, it->second));
```
pavel-shirshov: ,emplace_back(it->first, it->second)

common/redisclient.cpp
```diff
@@ -35,6 +43,34 @@ void RedisClient::hset(const string &key, const string &field, const string &val
...
+}
+
+void RedisClient::hmset(const string &key, const std::map<std::string, std::string> &vmap)
+{
+    if (vmap.size() == 0)
+        return;
+
+    vector<FieldValueTuple> values;
+    auto it = vmap.begin();
+    while (it != vmap.end())
```
pavel-shirshov: is probably better to use 
for(const auto it: vmap) ?

common/redisclient.cpp
```diff
@@ -57,6 +93,21 @@ unordered_map<string, string> RedisClient::hgetall(const string &key)
     return map;
 }
 
+std::map<std::string, std::string> RedisClient::hgetallordered(const std::string &key)
+{
+    RedisCommand sincr;
+    sincr.format("HGETALL %s", key.c_str());
```
pavel-shirshov: use string concatenation instead of format?

common/redisclient.cpp
```diff
@@ -57,6 +93,21 @@ unordered_map<string, string> RedisClient::hgetall(const string &key)
...
+    RedisReply r(m_db, sincr, REDIS_REPLY_ARRAY);
+
+    auto ctx = r.getContext();
+
+    map<string, string> map;
+    for (unsigned int i = 0; i < ctx->elements; i += 2)
+        map[string(ctx->element[i]->str)] = string(ctx->element[i+1]->str);
```
pavel-shirshov: map.emplace(ctx->element[i]->str, ctx->element[i+1]->str) ?

common/redisclient.cpp
```diff
@@ -35,6 +43,34 @@ void RedisClient::hset(const string &key, const string &field, const string &val
     RedisReply r(m_db, shset, REDIS_REPLY_INTEGER);
 }
 
+void RedisClient::hmset(const string &key, const vector<FieldValueTuple> &values)
+{
+    if (values.size() == 0)
```
pavel-shirshov: if (values.empty()) ?

common/redisclient.cpp
```diff
@@ -35,6 +43,34 @@ void RedisClient::hset(const string &key, const string &field, const string &val
...
+        return;
+
+    RedisCommand shmset;
+    shmset.formatHMSET(key, values);
+    RedisReply r(m_db, shmset, REDIS_REPLY_STATUS);
+}
+
+void RedisClient::hmset(const string &key, const std::map<std::string, std::string> &vmap)
+{
+    if (vmap.size() == 0)
```
pavel-shirshov: if (vmap.empty())?

common/redisclient.cpp
jipanyang: int64_t is the return value type from libhiredis for exists call.  
We are giving one key as parameter here, using bool as return value seems reasonable.  But is it a concern if we deviate from libhiredis as to the return value type?

common/redisclient.cpp
jipanyang: .format is the redis library function, it is being used in almost all the exist swss-common redis related functions.  I assume it should be fast and safe enough?

common/redisclient.cpp
jipanyang: ok

common/redisclient.cpp
jipanyang: ok.

common/redisclient.cpp
jipanyang: ok

common/redisclient.cpp
jipanyang: ok

common/redisclient.cpp
jipanyang: ok

common/schema.h
```diff
@@ -13,6 +13,7 @@ namespace swss {
 #define PFC_WD_DB       5
 #define FLEX_COUNTER_DB 5
 #define STATE_DB        6
+#define RESTORE_DB      7
```
qiluo-msft: >RESTORE_DB [](start = 8, length = 10)

Not needed

common/redisclient.cpp
pavel-shirshov: Hi Jipan,
I got your point, but I think in case of this library we have abstracted hiredis library and use swss-common instead. EXISTS returns 0 or 1 with Boolean semantics. So I think it's better to make it explicitly Boolean.
When I initially saw that your method return int64_t I checked redis documentation, because it was not clear for me why we need int64_t to return exists result.

common/redisclient.cpp
pavel-shirshov: Agree with you. But I think concatenating two strings are faster than parse a format string.
So it's more a speed optimization.

common/redisclient.cpp
jipanyang: Ok, will change to bool return.

common/redisclient.cpp
```diff
@@ -20,6 +20,14 @@ int64_t RedisClient::del(const string &key)
     return r.getContext()->integer;
 }
 
+int64_t RedisClient::exists(const string &key)
+{
+    RedisCommand sexists;
```
qiluo-msft: >sexists [](start = 17, length = 7)

Let's use a different var name.

common/redisclient.cpp
qiluo-msft: [https://redis.io/commands/exists](https://redis.io/commands/exists)
Because redis command return integer. However, we don't need it.

---
In reply to: [203180025](https://github.com/Azure/sonic-swss-common/pull/208#discussion_r203180025) [](ancestors = 203180025)

common/redisclient.cpp
jipanyang: After checking it again, RedisCommand is swss-common class, format is the member function of it which in turn calls redisvFormatCommand() to populate the command data.  

I saw comments like "Build the command string accordingly to protocol" in redisvFormatCommand(),  probably it is safer to use format() here to avoid any potential issue.

common/redisclient.cpp
```diff
@@ -35,6 +43,34 @@ void RedisClient::hset(const string &key, const string &field, const string &val
     RedisReply r(m_db, shset, REDIS_REPLY_INTEGER);
 }
 
+void RedisClient::hmset(const string &key, const vector<FieldValueTuple> &values)
+{
+    if (values.size() == 0)
+        return;
+
```
qiluo-msft: You don't need this check. Already checked in RedisCommand::formatHMSET()

common/redisclient.cpp
```diff
@@ -35,6 +43,34 @@ void RedisClient::hset(const string &key, const string &field, const string &val
...
+    RedisCommand shmset;
+    shmset.formatHMSET(key, values);
+    RedisReply r(m_db, shmset, REDIS_REPLY_STATUS);
+}
+
+void RedisClient::hmset(const string &key, const std::map<std::string, std::string> &vmap)
+{
+    if (vmap.size() == 0)
+        return;
```
qiluo-msft: RedisCommand::formatHMSET

common/redisclient.cpp
```diff
@@ -35,6 +43,34 @@ void RedisClient::hset(const string &key, const string &field, const string &val
...
+{
+    if (values.size() == 0)
+        return;
+
+    RedisCommand shmset;
+    shmset.formatHMSET(key, values);
+    RedisReply r(m_db, shmset, REDIS_REPLY_STATUS);
+}
+
+void RedisClient::hmset(const string &key, const std::map<std::string, std::string> &vmap)
```
qiluo-msft: >RedisClient::hmset [](start = 5, length = 18)

You can overload formatHMSET(), and call it.

common/redisclient.cpp
```diff
@@ -57,6 +93,21 @@ unordered_map<string, string> RedisClient::hgetall(const string &key)
     return map;
 }
 
+std::map<std::string, std::string> RedisClient::hgetallordered(const std::string &key)
```
qiluo-msft: >hgetallordered [](start = 48, length = 14)

Redis hash fields have no order. What is the use case?

common/redisclient.cpp
jipanyang: Ok,  will remote the check in hmset()

common/redisclient.cpp
jipanyang: Not sure what change you mean here?

common/redisclient.cpp
qiluo-msft: I will provide a PR based on your PR.

---
In reply to: [203208567](https://github.com/Azure/sonic-swss-common/pull/208#discussion_r203208567) [](ancestors = 203208567,203201493)

common/redisclient.cpp
qiluo-msft: I will provide a PR.

---
In reply to: [203206068](https://github.com/Azure/sonic-swss-common/pull/208#discussion_r203206068) [](ancestors = 203206068)

common/schema.h
qiluo-msft: I notice you check in again.

---
In reply to: [203177578](https://github.com/Azure/sonic-swss-common/pull/208#discussion_r203177578) [](ancestors = 203177578)

jipanyang: Sorry, wrong push. Should be gone now.

qiluo-msft: Still there.

---
In reply to: [203177578](https://github.com/Azure/sonic-swss-common/pull/208#discussion_r203177578) [](ancestors = 203177578)

common/redisclient.cpp
```diff
@@ -20,6 +20,19 @@ int64_t RedisClient::del(const string &key)
     return r.getContext()->integer;
 }
 
+bool RedisClient::exists(const string &key)
+{
+    RedisCommand rexists;
+    if (key.find_first_of(" \t") != string::npos)
```
pavel-shirshov: Probably it's better to trim spaces and tabs from the key?

common/redisclient.cpp
```diff
@@ -35,6 +48,20 @@ void RedisClient::hset(const string &key, const string &field, const string &val
     RedisReply r(m_db, shset, REDIS_REPLY_INTEGER);
 }
 
+void RedisClient::hmset(const string &key, const vector<FieldValueTuple> &values)
```
pavel-shirshov: implement both functions as a template? The code in them are identical

common/redisclient.cpp
```diff
@@ -43,18 +70,27 @@ void RedisClient::set(const string &key, const string &value)
...
+void RedisClient::hgetall(const std::string &key, OutputIterator result)
 {
     RedisCommand sincr;
     sincr.format("HGETALL %s", key.c_str());
     RedisReply r(m_db, sincr, REDIS_REPLY_ARRAY);
 
     auto ctx = r.getContext();
 
-    unordered_map<string, string> map;
+    map<string, string> map;
```
pavel-shirshov: what it was changed to map? Do we need an order here?

qiluo-msft: It is not used and should be removed. My mistake and I will fix it.

---
In reply to: [204127846](https://github.com/Azure/sonic-swss-common/pull/208#discussion_r204127846) [](ancestors = 204127846)

common/redisclient.cpp
qiluo-msft: Actually this input validation is to prevent any input like "key1 key2". Not for conveniently trimming.

---
In reply to: [204127076](https://github.com/Azure/sonic-swss-common/pull/208#discussion_r204127076) [](ancestors = 204127076)

pavel-shirshov: I'm good. Thank you for the explanations. I'd a comment explaining why do we need this validation here.

PR #214, opened 2018-07-28
Warm reboot: Add support for bool Select::hasCachedSelectable()

jipanyang:

For warm restart state restore,  there is need to check if all cached selectables have been processed.

### DIFF
--- a/common/select.cpp
+++ b/common/select.cpp
@@ -159,4 +159,9 @@ int Select::select(Selectable **c, unsigned int timeout)
 
 }
 
+bool Select::hasCachedSelectable()
+{
+    return !m_ready.empty();
+}
+
 };
--- a/common/select.cpp
+++ b/common/select.cpp
@@ -159,7 +159,7 @@ int Select::select(Selectable **c, unsigned int timeout)
 
 }
 
-bool Select::hasCachedSelectable()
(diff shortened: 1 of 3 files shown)


pavel-shirshov:
hasCachedSelectable() is a misleading name for me.
 Probably isQueueNonEmpty() is better name? I'd better export Boolean status for empty Queue.

What do you think?

jipanyang:
@pavel-shirshov  I don't have strong opinion as to the function name.  Actually it took me several iterations to just get current name.    
Since std::set<Selectable *, Select::cmp> m_ready; is a priority queue implementation(?),  I guess
isQueueNonEmpty() is a better name.

common/select.cpp
```diff
@@ -159,4 +159,9 @@ int Select::select(Selectable **c, unsigned int timeout)
 
 }
 
-};
+bool Select::isQueueNonEmpty()
```
qiluo-msft: >isQueueNonEmpty [](start = 13, length = 15)

Prefer positive adj.
isQueueEmpty

qiluo-msft:
@jipanyang Is this new function used somewhere? I think it just expose some internal implementation details, and the implementation will possible be changed in near future. Why it is useful? Can we remove it?

PR #215, opened 2018-07-30
Add function ConsumerTableBase::getDbConnnector

qiluo-msft:


common/consumertablebase.cpp
```diff
@@ -9,6 +9,11 @@ ConsumerTableBase::ConsumerTableBase(DBConnector *db, const std::string &tableNa
 {
 }
 
+DBConnector* ConsumerTableBase::getDbConnector()
```
prsunny: Should we return this as a const pointer?

pavel-shirshov:
Why do we need this?

PR #218, opened 2018-08-09
swss-common: handle as many entries as possible in each event handle …

dzhangalibaba:
…function

* add empty() in ConsumerTableBase


common/consumertablebase.cpp
```diff
@@ -43,4 +43,9 @@ void ConsumerTableBase::pop(std::string &key, std::string &op, std::vector<Field
     m_buffer.pop_front();
 }
 
+bool ConsumerTableBase::empty()
```
pavel-shirshov: I'd suggest to make it inline and const?

common/consumertablebase.h
```diff
@@ -20,6 +20,7 @@ class ConsumerTableBase: public TableConsumable, public RedisTransactioner
 
     void pop(std::string &key, std::string &op, std::vector<FieldValueTuple> &fvs, const std::string &prefix = EMPTY_PREFIX);
 
+    bool empty();
```
qiluo-msft: >empty [](start = 9, length = 5)

I believe your final goal could be achieved by the existing interface pops(). empty() is not necessary.

common/consumertablebase.cpp
dzhangalibaba: I understand the inline part. But why we need const? Do we use any const Consumer class object today?

common/consumertablebase.h
dzhangalibaba: pops is not enough. Today pops() only return one element. We want to handle as many entries as possible in processEvent(), we need to know if there is any entries in buffer to pop. If any, we continue. So we need an interface to check the buffer.

qiluo-msft: By design, pop() will return one element or none, but pops will return as many as possible. ref: https://github.com/Azure/sonic-swss-common/blob/master/common/consumertablebase.cpp#L22

If it is not true, it is a bug. The logic you need should be already designed there.

dzhangalibaba: i agree pop return one or none. Pops fill the buffer with default max 128 entries. some typo in my previous post, what I want to say is : 
Today, processEvent() calls pop(), pop() calls pops() to fill buffer first if buffer is empty, and then pop() return the first entry or none in the buffer. Then processEvent() start to handle this entry and exit. What I want to do is if the buffer is not empty, we continue to handle the entries left in the buffer in the same processEvent() and not exit. in fact , we can use pop() return value to check if buffer is empty or not without the added empty() interface. But the concern is if there are huge routes, once the buffer is empty, we call pops() to refill the buffer and pop doesn't return none until all the routes are handled, other processXXX func has no chance to be executed until all the routes are handled. The added empty() check is to grantee processEvent() exit after handling max buffer size entries and other  processXXX func has chance to be executed.

dzhangalibaba: look at this together with "sairedis: add while loop to make syncd processEvent() handle as many entries as possible in one event call #335", you may have more context. Let me know your option. If using only pop is OK in the loop, we can do that.

qiluo-msft: Thanks for the details. I believe you are talking about the following PR https://github.com/Azure/sonic-sairedis/pull/335. I think the use case there is quite valid, and can be achieved without new method.

```c++
auto batch = consumer.pops(kco);
for (auto kco : batch)
{
    // original do-while body
}
```

dzhangalibaba: int this way , we need to change the signature of the pops, pops didn't return any values today. Or we need pass some outside buffer as the first parameter, but in this way it is not consistent with pop func, pop is using member m_buffer. Or we need to add some interface to access internal m_buffer member, this actually is the same as adding empty(). anyway we need to change something instead of using pops or pop directly.

common/consumertablebase.cpp
pavel-shirshov: This method is not supposed to change anything in the object. So it's a const method. So const keyword will declare this intention explicitly

jordandong: OK. it is no harm to add const keyword, though in my option it is not necessary. i'll add it if finally we need this interface after QiLuo's suggestion.

dzhangalibaba: Sorry, forgot to signin using company account in the last message.  

OK. it is no harm to add const keyword, though in my option it is not necessary. i'll add it if finally we need this interface after QiLuo's suggestion.

common/consumertablebase.h
dzhangalibaba: As I mentioned in last comment, passing the outside buffer to pops() is Ok to achieve that. But it is not consistent to what it was doing in pop(), in pop() it is calling pops() and fill the internal member m_buffer.
This may cause problem when we mix use pop and pops directly from consumer. Today, probably we only use one of them, but it may cause issue later if we mix pop and pops together.

For example, assume we have a consumer,  and we have data no1 - no20 in redis DB.

First we call consumer.pop(), internally this may fill the m_buffer with size 10 and return one entry saying no1. Now we have 9 left in the m_buffer.
Then  we call consumer.pops(xx), then we may  get another 10 entries(no11 - no20). 

For the user point of view, it is not expected that pop() holds 9 entries.
if we use the interface as below order, it should return the first 11 emelents. But the buffer are different, it performs not as expected. This is my concern. Pop() not justing returning one element, it holds entries in m_buffer as well.  ideally pop() as a interface should only return one entry and does nothing else.
Consumer.pop(). -> expect to return no1
Consumer.pops()->expect to return no2 - no11 (actually it returns no11 -no20 since buffer are different)

qiluo-msft: You are right that "This may cause problem when we mix use pop and pops directly from consumer." and we should definitely fix it.

I will unblock this PR.

qiluo-msft:
Retest this please

common/consumertablebase.cpp
dzhangalibaba: updated the codes, now it should be fine

dzhangalibaba:
could we make some progress on this, it is pending for a long time. Thanks

PR #219, opened 2018-08-09
swss-common: add op 'create' case to avoid flush immediately

dzhangalibaba:
* create operation is used when writing routes into ASIC_DB
* operation create is added to make sure 'create' cases are redis pipelined
* with the changes in sairedis and swss, route performance improved by 200~300 routes/sec

### DIFF
--- a/common/producertable.cpp
+++ b/common/producertable.cpp
@@ -101,8 +101,8 @@ void ProducerTable::set(const string &key, const vector<FieldValueTuple> &values
     }
 
     enqueueDbChange(key, JSon::buildJson(values), "S" + op, prefix);
-    // Only buffer continuous "set/set" or "del" operations
-    if (!m_buffered || (op != "set" && op != "bulkset" ))
+    // Only buffer "set", "bulkset" or "create" operations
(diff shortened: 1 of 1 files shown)


dzhangalibaba:
could we make some progress on this, it is pending for a long time. Thanks

qiluo-msft:
Even you tested with writing routes into ASIC_DB, this is not correct in general. For 'create' op, it should flush pipeline including previous ones and creating one. Consider there are many DB connections, queued 'create' will make more dependency waiting.

dzhangalibaba:
1. when producer g_asicState writing ASIC_DB, it is in order and in the queue, so there should not have any dependency among different events. 

2. If you mean returned OID dependency, that is also fine. OID is  done in swss part which has no dependency on ASIC/sairedis.

Let me know if my understanding is wrong. 


> Even you tested with writing routes into ASIC_DB, this is not correct in general. For 'create' op, it should flush pipeline including previous ones and creating one. Consider there are many DB connections, queued 'create' will make more dependency waiting.

dzhangalibaba:
Thanks for review and approve. Could we merge this , then I can use the master branch to do the next development and don't need to patch it locally every time. Thanks.

PR #221, opened 2018-08-14
[converter]: Add to_upper function to convert string into uppercase

stcheng:
There are many string comparisons in the project. It is good to have
a function like this to compare between strings stored in the database
and those predefined values.

### DIFF
--- a/common/converter.h
+++ b/common/converter.h
@@ -1,6 +1,7 @@
 #ifndef __CONVERTER__
 #define __CONVERTER__
 
+#include <algorithm>
 #include <utility>
 #include <exception>
 #include <stdexcept>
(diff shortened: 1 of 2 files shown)


stcheng:
There are more than one place that need this helper function. https://github.com/Azure/sonic-swss/blob/master/orchagent/aclorch.cpp#L111 is one example.

common/converter.h
```diff
@@ -62,6 +63,13 @@ T to_uint(const std::string &str, T min = std::numeric_limits<T>::min(), T max =
     return static_cast<T>(__to_uint64(str, min, max));
 }
 
+std::string to_upper(const std::string &str)
+{
+    std::string ret = str;
+    transform(ret.begin(), ret.end(), ret.begin(), ::toupper);
+    return ret;
+}
```
qiluo-msft: Actually you can use value type in parameter
```c++
std::string to_upper(std::string str)
{
    transform(str.begin(), str.end(), str.begin(), ::toupper);
    return str;
}
```

PR #227, opened 2018-08-31
Add ProducerStateTable::count()/clear()

jipanyang:
…has been picked up by consumer.

### DIFF
--- a/common/producerstatetable.cpp
+++ b/common/producerstatetable.cpp
@@ -127,7 +127,7 @@ void ProducerStateTable::flush()
     m_pipe->flush();
 }
 
-int64_t ProducerStateTable::pendingCount()
+int64_t ProducerStateTable::count()
 {
     RedisCommand cmd;
     cmd.format("SCARD %s", getKeySetName().c_str());
@@ -139,7 +139,7 @@ int64_t ProducerStateTable::pendingCount()
 
 // Warning: calling this function will cause all data in keyset and the temporary table to be abandoned.
 // ConsumerState may have got the notification from PUBLISH, but will see no data popped.
-void ProducerStateTable::drop()
+void ProducerStateTable::clear()
 {
     // Assembly redis command args into a string vector
     vector<string> args;
--- a/common/producerstatetable.cpp
+++ b/common/producerstatetable.cpp
@@ -122,4 +122,15 @@ void ProducerStateTable::flush()
     m_pipe->flush();
 }
 
+bool ProducerStateTable::isEmpty()
+{
+    RedisCommand cmd;
+    cmd.format("SCARD %s", getKeySetName().c_str());
+    RedisReply r = m_pipe->push(cmd);
+    r.checkReplyType(REDIS_REPLY_INTEGER);
+
+    long long int setSize = r.getReply<long long int>();
+    return setSize == 0;
+}
+
 }
--- a/common/producerstatetable.cpp
+++ b/common/producerstatetable.cpp
@@ -39,6 +39,11 @@ ProducerStateTable::ProducerStateTable(RedisPipeline *pipeline, const string &ta
         "redis.call('DEL', KEYS[3])\n"
         "redis.call('PUBLISH', KEYS[1], ARGV[1])\n";
     m_shaDel = m_pipe->loadRedisScript(luaDel);
+
+    string luaDrop =
+        "redis.call('DEL', KEYS[1])\n"
+        "redis.call('DEL', KEYS[2])\n";
+    m_shaDrop = m_pipe->loadRedisScript(luaDrop);
 }
 
 ProducerStateTable::~ProducerStateTable()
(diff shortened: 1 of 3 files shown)


jipanyang:
@nikos-github suggested  another solution to clear the keyset and temporary table for producerState, it should work too.   I may create another PR for that if needed.

rodnymolina:
@jipanyang thanks for taking care of this one. It looks good to me.

zhenggen-xu:
@jipanyang  As talked, it is probably good to have another function to erase the data in the keyset and temporary table for the particular producerState, so the caller won't need wait for it to be empty.  In case of doing reconciliation, as long as the data is consistent in the consumer and database (plain table), later entries from applications would be passed through and got to the right state.

common/producerstatetable.cpp
```diff
@@ -122,4 +122,15 @@ void ProducerStateTable::flush()
...
 }
 
+bool ProducerStateTable::isEmpty()
+{
+    RedisCommand cmd;
+    cmd.format("SCARD %s", getKeySetName().c_str());
+    RedisReply r = m_pipe->push(cmd);
+    r.checkReplyType(REDIS_REPLY_INTEGER);
+
+    long long int setSize = r.getReply<long long int>();
```
qiluo-msft: >setSize [](start = 18, length = 7)

Maybe it's more useful to rename the function as count, and return the setSize directly.

rodnymolina:
@zhenggen-xu, good idea. However, in that case i don't see a need to call isEmpty() first -- one single call to (e.g) deletePending() would be enough.

common/producerstatetable.cpp
```diff
@@ -122,15 +127,39 @@ void ProducerStateTable::flush()
     m_pipe->flush();
 }
 
-bool ProducerStateTable::isEmpty()
+int64_t ProducerStateTable::setSize()
```
rodnymolina: i think you meant to say: getSize()

rodnymolina: Never mind, I see your point. Disregard.

common/producerstatetable.cpp
```diff
@@ -122,4 +127,39 @@ void ProducerStateTable::flush()
...
+    cmd.format("SCARD %s", getKeySetName().c_str());
+    RedisReply r = m_pipe->push(cmd);
+    r.checkReplyType(REDIS_REPLY_INTEGER);
+
+    return r.getReply<long long int>();
+}
+
+// Warning: calling this function will cause all data in keyset and the temporary table to be abandoned.
+// ConsumerState may have got the notification from PUBLISH, but will see no data popped.
+void ProducerStateTable::drop()
```
qiluo-msft: >drop [](start = 25, length = 4)

How about data pushed in the PUB/SUB channel between ProducerStateQueue and ConsumerStateQueue?

common/producerstatetable.cpp
```diff
@@ -122,4 +127,39 @@ void ProducerStateTable::flush()
...
+{
+    // Assembly redis command args into a string vector
+    vector<string> args;
+    args.emplace_back("EVALSHA");
+    args.emplace_back(m_shaDrop);
+    args.emplace_back("2");
+    args.emplace_back(getKeySetName());
+    args.emplace_back(getStateHashPrefix() + getTableName());
+    args.emplace_back("G");
+    args.emplace_back("''");
```
qiluo-msft: I guess they are copied. necessary here?

common/producerstatetable.h
```diff
@@ -26,12 +26,16 @@ class ProducerStateTable : public TableBase, public TableName_KeySet
 
     void flush();
 
+    int64_t pendingCount();
+
+    void drop();
```
qiluo-msft: By design, ProducerStateTable and ProducerTable have the exact same interfaces. Do you mind also implementing for the other?

common/producerstatetable.cpp
```diff
@@ -122,4 +127,39 @@ void ProducerStateTable::flush()
     m_pipe->flush();
 }
 
+int64_t ProducerStateTable::pendingCount()
```
qiluo-msft: >pendingCount [](start = 28, length = 12)

I think 'count' is a better name. For a queue concept, there is nothing pending.

common/producerstatetable.cpp
qiluo-msft: >drop [](start = 25, length = 4)

Call it clear()? like deque::clear()

jipanyang: If you know how to safely revoke the data already  "PUBLISH", please advise.

common/producerstatetable.cpp
jipanyang: place holder for argv.  If people don't mind script like
```
       "redis.call('DEL', KEYS[1])\n"
        "redis.call('DEL', ARGV[1])\n";
```
I could remove them.  But hope no back and forth.

common/producerstatetable.h
jipanyang: This is not a "solve all your problems" PR,  it should be better do that in separate PR.

common/producerstatetable.cpp
jipanyang: Ok.  Again, hope no back and forth.

common/producerstatetable.cpp
jipanyang: Ok.  Again, hope no back and forth.

common/producerstatetable.cpp
qiluo-msft: I mean 2 lines actually.
```
    args.emplace_back("G");
    args.emplace_back("''");
```

---
In reply to: [214504794](https://github.com/Azure/sonic-swss-common/pull/227#discussion_r214504794) [](ancestors = 214504794)

jipanyang: Yes, both lines are placeholder for argv.  Two keys are specified, putting two argv, no real use for now.    I thinks it is a trivial case, please just indicate what you want,   two place holder and rename "G" to ''  / one place holder/no argv place holder/one key one argv?

qiluo-msft: I prefer no placeholders at all. If not possible, one ore two empty string as placeholders are ok.

PR #228, opened 2018-09-01
Add new function NotificationConsumer::peek, add unit test

qiluo-msft:


common/notificationconsumer.h
```diff
@@ -23,9 +23,19 @@ class NotificationConsumer : public Selectable
...
+    // Note:
+    //    Ensure data ready before popping, either by select or peek
     void pop(std::string &op, std::string &data, std::vector<FieldValueTuple> &values);
     void pops(std::deque<KeyOpFieldsValuesTuple> &vkco);
 
+    // Check the internal queue which fed from redis socket for data ready
+    // Returns:
+    //     1 - data immediately available inside internal queue, may be just fed from redis socket
+    //     0 - no data both in internal queue or redis socket
+    //    -1 - error during peeking redis socket
```
pavel-shirshov: This function could return not only -1, but any other negative value. See above:
```
if (rc <= 0) 
            return rc; 
```

qiluo-msft: Just investigated, the return value is originally from ::poll, which could only be -1 if any error. http://man7.org/linux/man-pages/man2/poll.2.html#RETURN_VALUE
So let's keep the comment there for truth.

PR #229, opened 2018-09-05
sonic-swss-common : moving AISC_DB to this new DB instance

dzhangalibaba:
* moving AISC_DB to this new DB instance
* modify all the places where uses ASIC_DB to run on new DB instance
* all vs tests passed
* sonic-swss : #606, sonic-swss-common : #229, sonic-utilities : #313, sonic-snmpagent : #85, Ingrasys-sonic : #2 and sonic-buildimage :  #2020.  These six PR has dependency on each other since all are related to ASIC_DB modification, we should review and merge them together.

### DIFF
--- a/common/consumerstatetable.cpp
+++ b/common/consumerstatetable.cpp
@@ -15,6 +15,9 @@ ConsumerStateTable::ConsumerStateTable(DBConnector *db, const std::string &table
     : ConsumerTableBase(db, tableName, popBatchSize, pri)
     , TableName_KeySet(tableName)
 {
+    std::string m_luaScript = loadLuaScript("consumer_state_table_pops.lua");
+    m_sha = loadRedisScript(db, m_luaScript);
+
     for (;;)
     {
         RedisReply watch(m_db, "WATCH " + getKeySetName(), REDIS_REPLY_STATUS);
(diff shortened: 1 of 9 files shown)


common/consumerstatetable.cpp
```diff
@@ -15,6 +15,9 @@ ConsumerStateTable::ConsumerStateTable(DBConnector *db, const std::string &table
     : ConsumerTableBase(db, tableName, popBatchSize, pri)
     , TableName_KeySet(tableName)
 {
+    std::string m_luaScript = loadLuaScript("consumer_state_table_pops.lua");
```
qiluo-msft: >loadLuaScript [](start = 30, length = 13)

You load once per object. Actually once per process is enough. Why you change?

common/dbconnector.h
```diff
@@ -12,6 +12,7 @@ class DBConnector
 {
 public:
     static constexpr const char *DEFAULT_UNIXSOCKET = "/var/run/redis/redis.sock";
+    static constexpr const char *SECONDARY_UNIXSOCKET = "/var/run/redis/redis-secondary.sock";
```
qiluo-msft: >SECONDARY_UNIXSOCKET [](start = 33, length = 20)

This pattern of const strings are not scalable in long future. The swss-common should have no responsible or be aware of Redis instance partition.

`DEFAULT_UNIXSOCKET` is just a default value, any other non default usage is the responsible of that component, for example, swss or syncd's.

Suggest remove `SECONDARY_UNIXSOCKET`

tests/ntf_ut.cpp
```diff
@@ -58,7 +58,7 @@ TEST(Notifications, test)
 {
     SWSS_LOG_ENTER();
 
-    swss::DBConnector dbNtf(ASIC_DB, "localhost", 6379, 0);
+    swss::DBConnector dbNtf(ASIC_DB, "localhost", 6380, 0);
```
qiluo-msft: >6380 [](start = 50, length = 4)

I believe it will break unit test.

PR #231, opened 2018-09-13
NotificationProducer.send() returns the number of clients that received the message

qiluo-msft:


stcheng:
is it possible to have the function return > 1? could you add test cases there?

PR #232, opened 2018-09-17
[schema]: Change state table name with suffix STATE_TABLE

stcheng:
In the current schema, the state tabel names have conflicts with the
application table names. e.g. PORT_TABLE. It will cause issues when
one Orch class subscribes to two of these tables and uses the table
name to distinguish which table it is.

It is good to make distinctive names even if tables are in different
databases.


qiluo-msft:
I believe
> uses the table name to distinguish which table it is.

is a bad design. However I agree this is a quick workaround.

stcheng:
yes. the ideal solution is to use <db, table_name> pair to distinguish them

prsunny:
Just a thought, you could identify by the attribute "state" right?

qiluo-msft:
Solve the conflict?

lguohan:
build failure

lguohan:
what do you mean by "uses the table name to distinguish which table it is."? can we use other ways to distinguish which table it is?

PR #241, opened 2018-10-09
mv WarmStart class from swss to swss-common repo

jipanyang:


qiluo-msft:
Can we remove the default value docker_name = "swss"? It is easy to use the wrong one.

jipanyang:
> Can we remove the default value docker_name = "swss"? It is easy to use the wrong one.

Yes, agree with you.

PR #244, opened 2018-11-02
FixProducerStateTable::clear() to clear StateHash properly

taoyl-ms:
Current `FixProducerStateTable::clear()` implementation is calling `DEL` with `getStateHashPrefix() + getTableName()` (something like `DEL _ROUTE_TABLE`), which will not clear StateHash correctly. 

This change fixes it by `DEL` `_ROUTE_TABLE*` instead.

### DIFF
--- a/common/producerstatetable.cpp
+++ b/common/producerstatetable.cpp
@@ -43,7 +43,10 @@ ProducerStateTable::ProducerStateTable(RedisPipeline *pipeline, const string &ta
 
     string luaClear =
         "redis.call('DEL', KEYS[1])\n"
-        "redis.call('DEL', KEYS[2])\n"
+        "local keys = redis.call('KEYS', KEYS[2] .. '*')\n"
+        "for i,k in pairs(keys) do\n"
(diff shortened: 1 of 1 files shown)


taoyl-ms:
@jipanyang could you help review? I somehow could not add you to Reviewers list.

lguohan:
agree with jipan.

taoyl-ms:
Modified unit test to cover this.

tests/redis_piped_state_ut.cpp
```diff
@@ -378,6 +378,9 @@ TEST(ConsumerStateTable, async_singlethread)
     }
     p.clear();
     EXPECT_EQ(p.count(), 0);
+    string queryCommand = "KEYS " + p.getStateHashPrefix() + tableName + ":*";
```
jipanyang: Recommend to use p.getTableNameSeparator() to replace ":",  but it is ok for me.

lguohan: agree with jipan

PR #245, opened 2018-11-05
Remove local googletest submodule and link with system gtest

qiluo-msft:


qiluo-msft:
Build instruction updated https://github.com/Azure/sonic-swss-common#build-with-google-test

PR #246, opened 2018-11-06
Add system warm-reboot check in WarmRestart class

zhenggen-xu:
Add system warm-reboot check in WarmRestart class
UT test is also updated

Add the WarmRestart Class to python binding

### DIFF
--- a/common/warm_restart.cpp
+++ b/common/warm_restart.cpp
@@ -80,9 +80,11 @@ void WarmStart::initialize(const std::string &app_name,
  * capable component. Typically, this function will be called during initialization of
  * SONiC modules; however, method could be invoked at any given point to verify the
  * latest state of Warm-Restart functionality and to update the restore_count value.
+ * A flag of incr_restore_cnt is used to increase the restore cnt or not. Default: true
  */
 bool WarmStart::checkWarmStart(const std::string &app_name,
-                               const std::string &docker_name)
+                               const std::string &docker_name,
+                               const bool incr_restore_cnt)
 {
     std::string value;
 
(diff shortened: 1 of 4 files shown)


jipanyang:
The change looks ok to me.  What is the use case?
One possible issue is that,  for processes in a docker which are experiencing warm restart,  the flag doesn't completely identify whether it is a system warm reboot or a docker warm restart.

Ex.  for swss docker restart,  if the processes see system warm restart flag, it will do warm data reload. But seeing the flag enabled doesn't mean system level reboot is in progress, it may just be a docker warm restart.

zhenggen-xu:
This PR is for the system level warm reboot related processes. E,g, neighbor table restore during the system warm reboot.  We will check the system level reboot flag and make distinction between pure docker warm restart and system level warm reboot.  
C++/Python program will use this new lib function for checking instead of writing own functions.  It will be part of another PR update shortly.

zhenggen-xu:
Also added the WarmRestart class to python binding

common/warm_restart.cpp
```diff
@@ -173,6 +175,13 @@ bool WarmStart::isWarmStart(void)
     return warmStart.m_enabled;
 }
 
+bool WarmStart::isSystemWarmReboot(void)
```
jipanyang: Would it be better renaming it to isSystemWarmRebootEnabled() or something similar?  isSystemWarmReboot() seems  indicating the the system in going through warm reboot which may not be the case.

jipanyang: Would it be better renaming it to isSystemWarmRebootEnabled() or something similar? isSystemWarmReboot() seems indicating the the system in going through warm reboot which may not be the case.

zhenggen-xu:
Done the renaming, also added a flag to increase the restore count or not.

common/warm_restart.cpp
```diff
@@ -122,15 +124,14 @@ bool WarmStart::checkWarmStart(const std::string &app_name,
         warmStart.m_stateWarmRestartTable->hset(app_name, "restore_count", "0");
         return false;
     }
-    else
+
+    if (incr_restore_cnt)
```
jipanyang: I think "else if" should be used here.

PR #247, opened 2018-11-10
Add ProducerStateTable temp view implementation and UT

taoyl-ms:
See detail in design doc: https://github.com/Azure/SONiC/blob/master/doc/warm-reboot/view_switch.md

### DIFF
--- a/common/producer_state_table_apply_view.lua
+++ b/common/producer_state_table_apply_view.lua
@@ -1,0 +1,16 @@
+local arg_start = 2
+for i = 1, ARGV[arg_start] do
+    redis.call('SADD', KEYS[2], ARGV[arg_start + i])
+end
+arg_start = arg_start + ARGV[arg_start] + 1
+for i = 1, ARGV[arg_start] do
+    redis.call('SADD', KEYS[3], ARGV[arg_start + i])
+end
+arg_start = arg_start + ARGV[arg_start] + 1
+for j = 4, #KEYS do
+    for i = 1, ARGV[arg_start] do
+        redis.call('HSET', KEYS[j], ARGV[arg_start + i * 2 - 1], ARGV[arg_start + i * 2])
+    end
+    arg_start = arg_start + 2 * ARGV[arg_start] + 1
+end
+redis.call('PUBLISH', KEYS[1], ARGV[1])
--- a/common/producer_state_table_apply_view.lua
+++ b/common/producer_state_table_apply_view.lua
@@ -1,3 +1,26 @@
+--[[
+Sample args format:
+KEYS:
+   SAMPLE_CHANNEL
+   SAMPLE_KEY_SET
+   SAMPLE_DEL_KEY_SET
+   _SAMPLE:key_0
+   _SAMPLE:key_1
+ ARGV:
+   G   (String to be published to channel)
+   2   (Count of objects to set)
+   key_0
+   key_1
+   0   (Count of objects to del)
+   2   (Count of A/V pair of object 0)
+   attribute_0
+   value_0
+   attribute_1
+   value_1
+   1   (Count of A/V pair of object 1)
+   attribute_0
+   value_0
+]]
 local arg_start = 2
 for i = 1, ARGV[arg_start] do
     redis.call('SADD', KEYS[2], ARGV[arg_start + i])
--- a/common/producerstatetable.cpp
+++ b/common/producerstatetable.cpp
@@ -245,9 +245,19 @@ void ProducerStateTable::apply_temp_view()
         {
             needSet = true;
         }
-        if (needDel) keysToDel.emplace_back(key);
-        if (needSet) keysToSet.emplace_back(key);
-        else m_tempViewState.erase(key);                               // If exactly match, no need to sync new state to StateHash in DB
+
+        if (needDel) 
+        { 
+            keysToDel.emplace_back(key);
+        }
+        if (needSet) 
+        {
+            keysToSet.emplace_back(key);
+        }
+        else  // If exactly match, no need to sync new state to StateHash in DB 
+        {
+            m_tempViewState.erase(key);
+        }            
     }
     // Objects that do not exist currently need to be created
     for (auto const & kfvPair : m_tempViewState)
@@ -260,28 +270,7 @@ void ProducerStateTable::apply_temp_view()
     }
 
     // Assembly redis command args into a string vector
-    //
-    // Sample args format:
-    // KEYS:
-    //   SAMPLE_CHANNEL
-    //   SAMPLE_KEY_SET
-    //   SAMPLE_DEL_KEY_SET
-    //   _SAMPLE:key_0
-    //   _SAMPLE:key_1
-    // ARGV:
-    //   G   (String to be published to channel)
-    //   2   (Count of objects to set)
-    //   key_0
-    //   key_1
-    //   0   (Count of objects to del)
-    //   2   (Count of A/V pair of object 0)
-    //   attribute_0
-    //   value_0
-    //   attribute_1
-    //   value_1
-    //   1   (Count of A/V pair of object 1)
-    //   attribute_0
-    //   value_0
+    // See comment in producer_state_table_apply_view.lua for argument format
     vector<string> args;
     args.emplace_back("EVALSHA");
     args.emplace_back(m_shaApplyView);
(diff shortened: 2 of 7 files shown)


tests/redis_state_ut.cpp
```diff
@@ -29,18 +29,18 @@ static inline int getMaxFields(int i)
 
 static inline string key(int i)
 {
-    return string("key ") + to_string(i);
```
qiluo-msft: >key  [](start = 19, length = 4)

The original test data may be better, it is testing a key/field/value with blank char.

tests/redis_state_ut.cpp
```diff
@@ -502,6 +502,232 @@ TEST(ConsumerStateTable, set_pop_del_set_pop_get)
...
+    EXPECT_EQ(r4.getContext()->elements, (size_t) 0);
+    // Verify number of objects
+    keys.format("KEYS %s:*", tableName.c_str());
+    RedisReply r5(&db, keys, REDIS_REPLY_ARRAY);
+    EXPECT_EQ(r5.getContext()->elements, (size_t) numOfKeysNew);
+    // Verify number of fields
+    hlen.format("HLEN %s", (c.getKeyName(key(0))).c_str());
+    RedisReply r6(&db, hlen, REDIS_REPLY_INTEGER);
+    EXPECT_EQ(r6.getReply<long long int>(), (long long int) maxNumOfFieldsNew);
+}
```
qiluo-msft: Suggestion test cases on corner flows if not covered already.

Double create:
1. create
2. create
3. apply

Double apply:
1. create
2. apply
3. apply

Delete all:
1. create
2. set many
3. delete all
4. apply

Empty view (ie. clear)
1. create
2. immediately apply

common/producer_state_table_apply_view.lua
```diff
@@ -0,0 +1,16 @@
+local arg_start = 2
```
qiluo-msft: Suggest add file header comments to describe the usage, such as parameters.

taoyl-ms: I put them in apply_view() function in producerstatetable.cpp. Do you think here will be a better place?

---
In reply to: [232432763](https://github.com/Azure/sonic-swss-common/pull/247#discussion_r232432763) [](ancestors = 232432763)

common/producerstatetable.cpp
```diff
@@ -168,4 +189,142 @@ void ProducerStateTable::clear()
...
+
+void ProducerStateTable::apply_temp_view()
+{
+    // Drop all pending operation first
+    clear();
+
+    TableDump currentState;
+    {
+        Table mainTable(m_pipe, getTableName(), false);
+        mainTable.dump(currentState);
```
qiluo-msft: >dump [](start = 18, length = 4)

There is a comment "note that this function is not efficient"...

common/producerstatetable.cpp
```diff
@@ -168,4 +189,142 @@ void ProducerStateTable::clear()
...
+                needSet = true;
+            }
+        }
+        if (newFieldValueMap.size() > fieldValueMap.size())             // New field added
+        {
+            needSet = true;
+        }
+        if (needDel) keysToDel.emplace_back(key);
+        if (needSet) keysToSet.emplace_back(key);
+        else m_tempViewState.erase(key);                               // If exactly match, no need to sync new state to StateHash in DB
```
qiluo-msft: follow existing coding style?

common/producerstatetable.cpp
```diff
@@ -168,4 +189,142 @@ void ProducerStateTable::clear()
...
+    TableDump currentState;
+    {
+        Table mainTable(m_pipe, getTableName(), false);
+        mainTable.dump(currentState);
+    }
+
+    std::vector<std::string> keysToSet;
+    std::vector<std::string> keysToDel;
+
+    // Compare based on existing objects
```
jipanyang: It could be helpful if some debug level log is available to dump the data in current view and tmp view.

common/producerstatetable.cpp
```diff
@@ -168,4 +189,142 @@ void ProducerStateTable::clear()
...
+    //   SAMPLE_CHANNEL
+    //   SAMPLE_KEY_SET
+    //   SAMPLE_DEL_KEY_SET
+    //   _SAMPLE:key_0
+    //   _SAMPLE:key_1
+    // ARGV:
+    //   G   (String to be published to channel)
+    //   2   (Count of objects to set)
+    //   key_0
+    //   key_1
```
jipanyang: Excluding keys for del operation,  It looks duplicate key data provided in KEYS and ARGV?

taoyl-ms: Technically yes, since all KEYS for StateHash has a corresponding key to set in KeySet. However, One of them has format '_SAMPLE:key_0' and is actual key of redis object, and another one has format 'key_0' and is not object key (therefore should not be passed in KEYS), I'll prefer to pass them as KEYS and ARGV seperatedly. 

---
In reply to: [232443018](https://github.com/Azure/sonic-swss-common/pull/247#discussion_r232443018) [](ancestors = 232443018)

common/producerstatetable.cpp
taoyl-ms: We're calling it only once here as well.

---
In reply to: [232433590](https://github.com/Azure/sonic-swss-common/pull/247#discussion_r232433590) [](ancestors = 232433590)

common/producerstatetable.cpp
taoyl-ms: This mean to dump the entire table? Wouldn't there be too many logs?

---
In reply to: [232442252](https://github.com/Azure/sonic-swss-common/pull/247#discussion_r232442252) [](ancestors = 232442252)

jipanyang: it is one time thing, usually warm restart only.  For debugging purpose, I think it is ok and helpful.

common/producerstatetable.cpp
qiluo-msft: Even once, it is still not efficient.

---
In reply to: [232849738](https://github.com/Azure/sonic-swss-common/pull/247#discussion_r232849738) [](ancestors = 232849738,232433590)

taoyl-ms: OK. I'll refine the implementation for better performance.

---
In reply to: [233584343](https://github.com/Azure/sonic-swss-common/pull/247#discussion_r233584343) [](ancestors = 233584343,232849738,232433590)

tests/redis_state_ut.cpp
taoyl-ms: Test cases added according to CR comments.

---
In reply to: [232432588](https://github.com/Azure/sonic-swss-common/pull/247#discussion_r232432588) [](ancestors = 232432588)

common/table_dump.lua
```diff
@@ -4,10 +4,11 @@ local res = {}
 for i,k in pairs(keys) do
 
    local skeys = redis.call("HKEYS", k)
```
qiluo-msft: >skeys [](start = 9, length = 5)

skeys is not used and can be removed.

common/producerstatetable.cpp
```diff
@@ -168,4 +189,164 @@ void ProducerStateTable::clear()
...
+
+    TableDump currentState;
+    {
+        Table mainTable(m_pipe, getTableName(), false);
+        mainTable.dump(currentState);
+    }
+
+    // Print content of current view and temp view as debug log
+    SWSS_LOG_DEBUG("View switch of table %s required.", getTableName().c_str());
+    SWSS_LOG_DEBUG("Objects in current view:");
```
jipanyang: Not so sure,  would INFO level be better considering this is kind of critical data when debugging possible reconciliation issue,  there is so much routine debug level messages in different modules as of today.

common/producerstatetable.cpp
```diff
@@ -67,6 +71,16 @@ void ProducerStateTable::setBuffered(bool buffered)
 void ProducerStateTable::set(const string &key, const vector<FieldValueTuple> &values,
                  const string &op /*= SET_COMMAND*/, const string &prefix)
 {
+    if (m_tempViewActive)
+    {
+        // Write to temp view instead of DB
+        for (const auto& iv: values)
+        {
+            m_tempViewState[key][fvField(iv)] = fvValue(iv);
```
jipanyang: Is case like below also ok?  is that covered in unit test? 
```
127.0.0.1:6379[4]> hgetall hello
1) ""
2) ""
```

taoyl-ms: Yes. Unit test already covered empty values. 

---
In reply to: [234321630](https://github.com/Azure/sonic-swss-common/pull/247#discussion_r234321630) [](ancestors = 234321630)

PR #249, opened 2018-11-16
Add NEIGH_TABLE to configDB for neighbor configuration

prsunny:
**Note: `CFG_VRF_TABLE_NAME`  is already defined in line 125**


prsunny:
@jipanyang , updated the schema as part of https://github.com/Azure/sonic-swss/pull/693

PR #252, opened 2018-11-17
ASIC temperature sensors support

padmanarayana:
**Please do not merge yet.**

Recently (https://github.com/opencomputeproject/SAI/pull/880), new switch attributes were added to retrieve the temperature readings from the ASIC's internal sensors. 

This is a preliminary commit (pending SAI support from vendors) for temperature monitoring. The max, average and the entire list of temperatures are now added to the flex counters DB so that platform sensors scripts may query and use these values in thermal control algorithms.


stcheng:
is this still needed?

PR #253, opened 2018-12-17
[selectable]: Update throw message

stcheng:
Update SelectableEvent and SelectableTimer throw message when
read failed.

### DIFF
--- a/common/selectableevent.cpp
+++ b/common/selectableevent.cpp
@@ -53,9 +53,8 @@ void SelectableEvent::readData()
 
     if (s != sizeof(uint64_t))
     {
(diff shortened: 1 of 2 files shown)


common/selectableevent.cpp
```diff
@@ -53,9 +53,8 @@ void SelectableEvent::readData()
 
     if (s != sizeof(uint64_t))
     {
-        SWSS_LOG_ERROR("read failed, errno: %s", strerror(errno));
-
-        throw std::runtime_error("read failed");
+        SWSS_LOG_THROW("SelectableEvent read failed, errno: %s",
```
qiluo-msft: Please also add 's' into  the message. Same below.

PR #254, opened 2018-12-27
Fix *::readData() which uses read(2)

pavel-shirshov:
[read(2)](http://man7.org/linux/man-pages/man2/read.2.html) could read less bytes than requested.
`It is not an error if this number is smaller than the number of bytes requested`
Read until we read everything we want


common/selectableevent.cpp
```diff
@@ -43,18 +43,25 @@ int SelectableEvent::getFd()
...
-        s = read(m_efd, &r, sizeof(uint64_t));
+        s = read(m_efd, r_ptr, bytes_to_read);
+        if (s > 0)
+        {
+            r_ptr += s;
+            bytes_to_read -= s;
+        }
     }
-    while(s == -1 && errno == EINTR);
+    while ((s == -1 && errno == EINTR) || (bytes_to_read > 0));
```
qiluo-msft: >bytes_to_read [](start = 43, length = 13)

If (bytes_to_read > 0) && (s == -1 && errno != EINTR), do you want to loop or throw?

common/selectabletimer.cpp
```diff
@@ -75,19 +75,25 @@ int SelectableTimer::getFd()
 void SelectableTimer::readData()
```
qiluo-msft: >readData [](start = 22, length = 8)

could you reuse code?

pavel-shirshov: I think it doesn't worth it in this case. We have this code in two methods only.

common/selectableevent.cpp
qiluo-msft: s > 0 will be more accurate. If this is correct, consider more readable code:
```
    do
    {
        s = read(m_efd, r_ptr, bytes_to_read);
        if (s > 0)
        {
            r_ptr += s;
            bytes_to_read -= s;
            if (bytes_to_read > 0)
            {
                continue;
            }
        }
    }
    while (s == -1 && errno == EINTR);
```

---
In reply to: [244072039](https://github.com/Azure/sonic-swss-common/pull/254#discussion_r244072039) [](ancestors = 244072039)

pavel-shirshov: s could be equal to zero which is valid result. s == -1 indicate and error.
So we try to read again in two cases:
1. We got an error and the error is EINTR
2. We didn't get an error and we still have something to read.

pavel-shirshov: I'm going to use glibc functions to read from event_fd which have simpler interface
```
           typedef uint64_t eventfd_t;

           int eventfd_read(int fd, eventfd_t *value);
           int eventfd_write(int fd, eventfd_t value);
```

PR #261, opened 2019-02-13
Force only supported commands on consumer table

kcudnik:


common/consumer_table_pops.lua
```diff
@@ -46,7 +46,14 @@ for i = n, 1, -3 do
...
        end
 
-   elseif op ~= 'flush' and op ~= 'flushresponse' and op ~= 'get' and op ~= 'getresponse' and op ~= 'notify' then
+   elseif
+       op == 'set' or
+       op == 'SET' or
+       op == 'create' or
+       op == 'remove' or
+       op == 'DEL' then
+   -- put entries into REDIS hash only when operations aare this types
```
lguohan: aare -> are

kcudnik: fixed

common/consumer_table_pops.lua
```diff
@@ -46,7 +46,14 @@ for i = n, 1, -3 do
            st = st + 2
        end
 
-   elseif op ~= 'flush' and op ~= 'flushresponse' and op ~= 'get' and op ~= 'getresponse' and op ~= 'notify' then
+   elseif
+       op == 'set' or
+       op == 'SET' or
```
qiluo-msft: Why SET/DEL are upcases? Should we make them case insensitive?

kcudnik: thats a good question, maybe reducing to just lower case everything is good idea

PR #264, opened 2019-02-26
Fix MIRROR_SESSION table macro name

jipanyang:

"MIRROR_SESSION" table is in stateDB.


yxieca:
This change will cause warm-reboot issue. DB migration code is required to be in place.

lguohan:
that is in a different PR.

stcheng:
ah this is also causing the current unit tests failed due to the naming of the tables get changed 😿

PR #270, opened 2019-03-06
Make class Select support batch read from selectables

pavel-shirshov:
Previously Select class supported only one-by-one read from Selectables.
It was impossible to get notification from Select, read many entries from a selectable and get another right notification from Select.
This patch introduce possibility to read whole Selectable cache after one signal from Select. An application will not receive any notification from the selectable with empty cache.

Currently RedisSelect doesn't support batch reads directly. Because of RedisSelect class design, we must update m_queueLength after every read. Currently we update it in Select::poll_descriptors(), but we should run the update after each read.


common/select.cpp
```diff
@@ -110,17 +110,22 @@ int Select::poll_descriptors(Selectable **c, unsigned int timeout)
...
         auto sel = *m_ready.begin();
 
-        *c = sel;
-
         m_ready.erase(sel);
         // we must update clock only when the selector out of the m_ready
         // otherwise we break invariant of the m_ready
         sel->updateLastUsedTime();
 
+        if (!sel->hasData())
```
qiluo-msft: >hasData [](start = 18, length = 7)

I think the interface of hasCachedData is redundant if we have hasData already.

I propose code change
```
while (!m_ready.empty())
{
    auto sel = *m_ready.begin();

    m_ready.erase(sel);
    // we must update clock only when the selector out of the m_ready
    // otherwise we break invariant of the m_ready
    sel->updateLastUsedTime();

    if (!sel->hasData())
    {
        continue;
    }

    *c = sel;

    // reinsert Selectable back to the m_ready set, when there're more messages in the cache
    m_ready.insert(sel);

    sel->updateAfterRead();

    return Select::OBJECT;
}
```

common/selectable.h
```diff
@@ -23,6 +23,16 @@ class Selectable
     /* Read all data from the fd assicaited with Selectable */
     virtual void readData() = 0;
 
+    /* true if Selectable has data in it for immediate read
+       this is supposed to be implemented in Selectibles, which implements hasCachedData().
+       it's possible to have a case, when an application read all cached data after 1 signal,
+       and we have stale selectible with no data in the m_ready queue in Select class.
+       The class without hasCachedData never is going to be in m_ready state without the data */
```
qiluo-msft: >*/ [](start = 94, length = 2)

The comment block formatting is weird. How about
```
/* something
    something more
*/
```

common/select.cpp
pavel-shirshov: It's not redundant. 
Look at situation when you have Data to read from a selectable, but its cache is empty.
In this case you should return Select::OBJECT, because you have data for reading, but you shouldn't put it back into m_ready, because there is no data to read in the future.

qiluo-msft: I saw your point and it is a valid case. However, putting it back doesn't heart much, because future poll_descriptors() will still check hasData(), and return TIMEOUT.

---
In reply to: [263169747](https://github.com/Azure/sonic-swss-common/pull/270#discussion_r263169747) [](ancestors = 263169747)

qiluo-msft: Offline discussed, we would like to revisit this comment later.

---
In reply to: [263171154](https://github.com/Azure/sonic-swss-common/pull/270#discussion_r263171154) [](ancestors = 263171154,263169747)

PR #271, opened 2019-03-10
Qiluo/producer state publish2

qiluo-msft:
This is a competing PR to https://github.com/Azure/sonic-swss-common/pull/258

I reverted https://github.com/Azure/sonic-swss-common/pull/214 since it exposed unnecessary implementation details.

Problem statement is same as https://github.com/Azure/sonic-swss-common/pull/258, so I shamelessly copied it here.

> The typical use case that this PR is trying to deal with is related to route flapping.
> 
> Assuming such a scenario:
> <1> 4K routes learned on an interface.
> 
> <2> interface down, 4k routes withdraw.
> <3> fpmsyncd inserts 4k route keys into producerStateTabk keyset and generates 4k redis publish notifications  
> 
> <4> Interface goes up
> 
> <4.A> 
> Interface may go up and learn the 4k routes back  before orchagent has chance to pick up any notification and process the keyset.  That is fine, change in https://github.com/Azure/sonic-swss-common/pull/257 handled this case, no new redis publish notification will be created since the route keys already exist in producerStateTabk keyset and are to be processed. 
> 
> <4.B> 
> Interface may go up and learn the 4k routes back after orchagent has processed a few notifications but not all the redis notifications (say, 1k notification processed),  with the first notification,  consumer in orchagent batch popped all the 4k route keys from producerStateTabk keyset,  (orchagent has set popBatchSize as 8192),  for the other 999 notifications processed, orchagent just call consumer_state_table_pops.lua 999 times without seeing any real data.
> 4k - 1k = 3k more redis notifications pending in redis queue now.
> 
> <5> Interface goes down 
> <5.A>  If <4.A> was hit during previous interface up, good, no new redis publish notification will be generated
> 
> <5.B>  If <4.B> was hit during previous interface up. 4K more redis notifications put into the redis queue.
> <6> Interface goes up, loop back to <4>
> 
> The exact scenarios vary, but from <4> to <6>, number of pending notifications keeps increasing.   
> 
> **The issue is that orchagent may be abled to handle all the route changes with the batch processing (popBatchSize = 8192), redis publish notification was generated for each key and orchagent has to respond to each of the notifications with current select framework though most of them bring no value.**

The implementation proposed here follows some design principles.
1. Generate one and only one Redis publish notification (channel message) when there is one data in the key set. So the number of messages and keys are the same, and we can ensure to decrease the same amount during pop/pops()
2. The only exception to above is when producer calls clear(). The channel messages could not be recalled.
3. Some users will prefer pop(), and others may prefer pops() to pop a batch. But no one will use both for the same consumer table
4. Consumer will never publish to channel, ie. create channel message


qiluo-msft:
@jipanyang @zhenggen-xu As we offline discussed I submit this PR. Please help review. Thanks for your solution and inspiration.

common/redisselect.h
```diff
@@ -32,6 +32,11 @@ class RedisSelect : public Selectable
 protected:
     std::unique_ptr<DBConnector> m_subscribe;
     long long int m_queueLength;
+
+    void discard(long long int n);
```
kcudnik: size_t maybe ?

kcudnik (dismissed):

common/redisselect.h
qiluo-msft: It is following m_queueLength, which is following redis api convention.

---
In reply to: [264058130](https://github.com/Azure/sonic-swss-common/pull/271#discussion_r264058130) [](ancestors = 264058130)

kcudnik: ok

common/redisselect.cpp
```diff
@@ -66,6 +72,35 @@ void RedisSelect::updateAfterRead()
...
+    if (n <= 0)
+        n = 1;
+
+    if (n > m_queueLength)
+    {
+        /* If we have less messages, discard them all
+         * This is no big harm since all the late messages will be selected and nothing popped
+         * when the channel is not completely busy.
+         */
+        n -= m_queueLength;
```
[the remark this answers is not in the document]
qiluo-msft: No. It will be optimized out by compiler. It may help readability a little bit.

---
In reply to: [276422184](https://github.com/Azure/sonic-swss-common/pull/271#discussion_r276422184) [](ancestors = 276422184)

common/redisselect.cpp
```diff
@@ -66,6 +72,35 @@ void RedisSelect::updateAfterRead()
...
+void RedisSelect::discard(long long int n)
+{
+    readRemainingData();
+
+    /*
+     * We will discard at least one message, to prevent any mistakenly infinite loop
+     * in the select-pop(s) pattern
+     */
+    if (n <= 0)
+        n = 1;
```
jipanyang: In what scenario n <= 0?

common/redisselect.cpp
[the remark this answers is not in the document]
jipanyang: > No. It will be optimized out by compiler. It may help readability a little bit.
> 
> In reply to: [276422184](https://github.com/Azure/sonic-swss-common/pull/271#discussion_r276422184) [](ancestors = 276422184)

I think it causes more confusion than helping readability.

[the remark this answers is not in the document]
qiluo-msft: Removed.

---
In reply to: [276515514](https://github.com/Azure/sonic-swss-common/pull/271#discussion_r276515514) [](ancestors = 276515514)

common/redisselect.cpp
qiluo-msft: I am trying to prevent any logic error outside the scope of RedisSelect. Considering the case if there are redundant notifications but there is no data popped, caller may call discard(0), and here it consumer one message in the queue if there is any. discard(negative) is not for protection purpose and should be not in a true case.

---
In reply to: [276515380](https://github.com/Azure/sonic-swss-common/pull/271#discussion_r276515380) [](ancestors = 276515380)

lguohan:
retest this please

common/consumertablebase.cpp
```diff
@@ -43,4 +43,22 @@ void ConsumerTableBase::pop(std::string &key, std::string &op, std::vector<Field
...
+// ref: Select::selecat() implementation
+bool ConsumerTableBase::hasCachedData()
+{
+    return true;
+}
+
+// Override with an empty body, and subclasses will explicitly discard messages in the channel
+void ConsumerTableBase::updateAfterRead()
+{
+}
```
pavel-shirshov: I don't understand why do we need this?
The base class already has an empty body

tests/redis_subscriber_state_ut.cpp
```diff
@@ -431,92 +430,3 @@ TEST(SubscriberStateTable, one_producer_multiple_subscriber)
...
-            c1.pop(kco);
-        }
-
-        EXPECT_EQ(kfvKey(kco), key);
-        EXPECT_EQ(kfvOp(kco), "SET");
-        /* No cached selectable left */
-        r = cs.isQueueEmpty();
-        EXPECT_TRUE(r);
-    }
-}
```
pavel-shirshov: Why this test was removed?

common/consumertablebase.cpp
```diff
@@ -43,4 +43,22 @@ void ConsumerTableBase::pop(std::string &key, std::string &op, std::vector<Field
...
+    return RedisSelect::hasData() || !m_buffer.empty();
+}
+
+// hasCachedData is depreciated, and this is a safe placeholder
+// We will use hasData() to indicate that there is pending data
+// ref: Select::selecat() implementation
+bool ConsumerTableBase::hasCachedData()
+{
+    return true;
+}
```
pavel-shirshov: The base class has the same implementation. It doesn't need to be implemented here.

common/subscriberstatetable.cpp
```diff
@@ -88,7 +88,7 @@ bool SubscriberStateTable::hasData()
 
 bool SubscriberStateTable::hasCachedData()
 {
-    return m_buffer.size() > 1 || m_keyspace_event_buffer.size() > 1;
+    return true;
```
pavel-shirshov: It's default implementation in the base class. We could remove whole method.

PR #286, opened 2019-06-19
Add an assert to logger, which will log a message and abort.

renukamanavalan:
Upon failure to read timer_fd, do an assert instead of throw.
This way, the core can be obtained at the point of failure, instead of silent process termination.

### DIFF
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -17,6 +17,22 @@
 
 namespace swss {
 
+void
+err_exit(const char *fn, int ln, int e, const char *fmt, ...)
+{
+    va_list ap;
+    char buff[1024];
+    size_t len;
+
+    va_start(ap, fmt);
+    snprintf(buff, sizeof(buff), "%s::%d err(%d/%s): ", fn, ln, e, strerror(e));
+    len = strlen(buff);
+    vsnprintf(buff+len, sizeof(buff)-len, fmt, ap);
+    va_end(ap);
+    SWSS_LOG_ERROR("Aborting: %s", buff);
+    abort();
+}
+
 Logger::~Logger() {
     if (m_settingThread) {
         m_settingThread->detach();
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -17,8 +17,7 @@
 
 namespace swss {
 
-void
-err_exit(const char *fn, int ln, int e, const char *fmt, ...)
+void err_exit(const char *fn, int ln, int e, const char *fmt, ...)
 {
     va_list ap;
     char buff[1024];
--- a/common/logger.h
+++ b/common/logger.h
@@ -31,7 +31,7 @@ void err_exit(const char *fn, int ln, int e, const char *fmt, ...)
         ;
 
 
-#define ASSERT(x, fmt, args...)                             \
+#define ABORT_IF_NOT(x, fmt, args...)                      \
     if (!(x)) {                                             \
         int e = errno;                                      \
         err_exit(__FUNCTION__, __LINE__, e, (fmt), ##args); \
--- a/common/logger.h
+++ b/common/logger.h
@@ -23,6 +23,21 @@ namespace swss {
 
 #define SWSS_LOG_THROW(MSG, ...)       swss::Logger::getInstance().wthrow(swss::Logger::SWSS_ERROR,  ":- %s: " MSG, __FUNCTION__, ##__VA_ARGS__)
 
+extern void err_exit(const char *fn, int ln, int e, const char *fmt, ...)
+#ifdef __GNUC__
+        __attribute__ ((format (printf, 4, 5)))
+        __attribute__ ((noreturn))
+#endif
+        ;
+
+
+#define ASSERT(x, fmt, args...)                             \
+    if (!(x)) {                                             \
+        int _e = errno;                                     \
+        err_exit(__FUNCTION__, __LINE__, _e, fmt, ##args);  \
+    }
+
+
 class Logger
 {
 public:
--- a/common/logger.h
+++ b/common/logger.h
@@ -23,7 +23,7 @@ namespace swss {
 
 #define SWSS_LOG_THROW(MSG, ...)       swss::Logger::getInstance().wthrow(swss::Logger::SWSS_ERROR,  ":- %s: " MSG, __FUNCTION__, ##__VA_ARGS__)
 
-extern void err_exit(const char *fn, int ln, int e, const char *fmt, ...)
+void err_exit(const char *fn, int ln, int e, const char *fmt, ...)
 #ifdef __GNUC__
         __attribute__ ((format (printf, 4, 5)))
         __attribute__ ((noreturn))
@@ -33,8 +33,8 @@ extern void err_exit(const char *fn, int ln, int e, const char *fmt, ...)
 
 #define ASSERT(x, fmt, args...)                             \
     if (!(x)) {                                             \
-        int _e = errno;                                     \
-        err_exit(__FUNCTION__, __LINE__, _e, fmt, ##args);  \
+        int e = errno;                                      \
+        err_exit(__FUNCTION__, __LINE__, e, (fmt), ##args); \
     }
 
 
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -77,20 +77,16 @@ void SelectableTimer::readData()
     uint64_t r;
 
     ssize_t s;
+    errno = 0;
     do
     {
-        // Read the timefd so it will be reset
         s = read(m_tfd, &r, sizeof(uint64_t));
     }
     while(s == -1 && errno == EINTR);
 
-    if (s != sizeof(uint64_t))
-    {
-        SWSS_LOG_THROW("SelectableTimer read failed, s:%zd errno: %s",
-                s, strerror(errno));
-    }
-}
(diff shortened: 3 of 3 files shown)


common/logger.cpp
```diff
@@ -17,6 +17,22 @@
...
+void
+err_exit(const char *fn, int ln, int e, const char *fmt, ...)
+{
+    va_list ap;
+    char buff[1024];
+    size_t len;
+
+    va_start(ap, fmt);
+    snprintf(buff, sizeof(buff), "%s::%d err(%d/%s): ", fn, ln, e, strerror(e));
+    len = strlen(buff);
```
pavel-shirshov: snprintf will return this len for you:
http://man7.org/linux/man-pages/man3/fprintf.3p.html#RETURN_VALUE

common/selectabletimer.cpp
```diff
@@ -77,20 +77,16 @@ void SelectableTimer::readData()
...
-    if (s != sizeof(uint64_t))
-    {
-        SWSS_LOG_THROW("SelectableTimer read failed, s:%zd errno: %s",
-                s, strerror(errno));
-    }
-}
+    ASSERT(s == sizeof(uint64_t), "Failed to read timerfd. s=%ld", s)
 
-// FIXME: if timer events are read slower than timer frequency we will lost time events
+    // s = count of timer events happened since last read.
```
pavel-shirshov: Is it typo? Should it be `r = count of timer events happened since last read`
I think s must be `sizeof(uint64_t)`?

common/logger.cpp
renukamanavalan: True. Technically it says, 'what it had written or could have written'. With a buffer of 1024, it should be 'what it had written'. 

With that kind of if/but, I try to go on the solid side. Moreover, we are on the way to abort, so an extra call to strlen for a memory in stack/cache, should not cost much.

common/selectabletimer.cpp
renukamanavalan: Good catch. It should be "r = ..."

renukamanavalan: Fixed.

common/logger.h
```diff
@@ -23,6 +23,21 @@ namespace swss {
 
 #define SWSS_LOG_THROW(MSG, ...)       swss::Logger::getInstance().wthrow(swss::Logger::SWSS_ERROR,  ":- %s: " MSG, __FUNCTION__, ##__VA_ARGS__)
 
+extern void err_exit(const char *fn, int ln, int e, const char *fmt, ...)
```
qiluo-msft: >extern [](start = 0, length = 6)

extern is redundant

common/logger.h
```diff
@@ -23,6 +23,21 @@ namespace swss {
...
+        __attribute__ ((format (printf, 4, 5)))
+        __attribute__ ((noreturn))
+#endif
+        ;
+
+
+#define ASSERT(x, fmt, args...)                             \
+    if (!(x)) {                                             \
+        int _e = errno;                                     \
+        err_exit(__FUNCTION__, __LINE__, _e, fmt, ##args);  \
```
qiluo-msft: >fmt [](start = 45, length = 3)

(fmt)

common/logger.h
```diff
@@ -23,6 +23,21 @@ namespace swss {
...
+#ifdef __GNUC__
+        __attribute__ ((format (printf, 4, 5)))
+        __attribute__ ((noreturn))
+#endif
+        ;
+
+
+#define ASSERT(x, fmt, args...)                             \
+    if (!(x)) {                                             \
+        int _e = errno;                                     \
```
qiluo-msft: >_e [](start = 12, length = 2)

why not just use errno in next statement? We prefer no underscore for local var.

common/logger.cpp
```diff
@@ -17,6 +17,22 @@
 
 namespace swss {
 
+void
+err_exit(const char *fn, int ln, int e, const char *fmt, ...)
```
qiluo-msft: Could you follow the formatting of similar pattern in the same file?

common/logger.cpp
```diff
@@ -17,6 +17,22 @@
...
+{
+    va_list ap;
+    char buff[1024];
+    size_t len;
+
+    va_start(ap, fmt);
+    snprintf(buff, sizeof(buff), "%s::%d err(%d/%s): ", fn, ln, e, strerror(e));
+    len = strlen(buff);
+    vsnprintf(buff+len, sizeof(buff)-len, fmt, ap);
```
qiluo-msft: >vsnprintf [](start = 4, length = 9)

Check return value for any encoding error by user input.

common/logger.cpp
qiluo-msft: >strlen [](start = 10, length = 6)

You can get len from `snprintf` return value

common/logger.h
renukamanavalan: dropped.

common/logger.h
renukamanavalan: good catch

common/logger.h
renukamanavalan: I can drop '_'. I still would like to make a copy, before calling macro, as processing the args passed in has a theoretical potential to affect the global errno.

common/logger.cpp
renukamanavalan: Guess, "void err_exit ...". Done

common/logger.cpp
renukamanavalan: There is nothing much I can do by checking the return value, as I don't have any further choice, other than logging, what it managed to write,

All I would care is that there is no buffer overflow, which the API assures.

common/logger.cpp
renukamanavalan: True. Technically it returns, 'what it had written or could have written'. With a buffer of 1024, it should be 'what it had written'.

"return value is  the  number of characters (excluding the terminating null byte) which would have been written "

With that kind of if/but, I try to go on the solid side. Moreover, we are on the way to abort, so an extra call to strlen for a memory in stack/cache, should not cost much.

common/logger.cpp
```diff
@@ -17,6 +17,21 @@
...
+    va_end(ap);
+    SWSS_LOG_ERROR("Aborting: %s", buff);
```
kcudnik: this should be higher even ? do we have critical ?

common/selectabletimer.cpp
```diff
@@ -77,20 +77,16 @@ void SelectableTimer::readData()
...
     }
     while(s == -1 && errno == EINTR);
 
-    if (s != sizeof(uint64_t))
-    {
-        SWSS_LOG_THROW("SelectableTimer read failed, s:%zd errno: %s",
-                s, strerror(errno));
-    }
-}
+    ASSERT(s == sizeof(uint64_t), "Failed to read timerfd. s=%ld", s)
```
kcudnik: why this abort here is preffered than throw ?

renukamanavalan: The "throw" causes orchagent to exit silently.
As we do not have RCA for this issue yet, I would like a core dump at the point of failure.

In either case, orchagent exits. With this change, it gives a dump too.

common/logger.h
```diff
@@ -23,6 +23,21 @@ namespace swss {
...
 
+void err_exit(const char *fn, int ln, int e, const char *fmt, ...)
+#ifdef __GNUC__
+        __attribute__ ((format (printf, 4, 5)))
+        __attribute__ ((noreturn))
+#endif
+        ;
+
+
+#define ASSERT(x, fmt, args...)                             \
```
qiluo-msft: >ASSERT [](start = 8, length = 6)

ASSERT is not a good name. Normally asserts do nothing in NDEBUG.

common/logger.cpp
renukamanavalan: Good point. We do have critical and above.

        SWSS_EMERG,
        SWSS_ALERT,
        SWSS_CRIT,

Which one would you suggest?

common/logger.h
renukamanavalan: Do you have a suggestion ?

common/logger.cpp
renukamanavalan: I do lean towards raising log level to Critical. But doing here would make it look CRITICAL for exit through this path (ASSERT) only, but not all.

I prefer to take it as another job and tackle it, as a whole.

common/logger.h
renukamanavalan: Please suggest, as I don't have another good choice.

qiluo-msft: Something like ABORT_IF

---
In reply to: [295890565](https://github.com/Azure/sonic-swss-common/pull/286#discussion_r295890565) [](ancestors = 295890565)

renukamanavalan: Good one. But more like, ABORT_IF_NOT as the expression written for evaluation, is for good case and we abort if not the good case.

common/selectabletimer.cpp
```diff
@@ -74,7 +74,7 @@ int SelectableTimer::getFd()
 
 void SelectableTimer::readData()
 {
-    uint64_t r;
+    uint64_t r = UINTPTR_MAX;
```
pavel-shirshov: Probably  `UINT64_MAX` is better here?
Or even
`std::numeric_limits<uint64_t>::max()`

renukamanavalan: Agree.

PR #288, opened 2019-06-24
Add an assert to logger, which will log a message and abort. (#286)

renukamanavalan:
* Add an assert to logger, which will log a message and abort.
Upon failure to read timer_fd, do an assert instead of throw.
This way, the core can be obtained at the point of failure, instead of silent process termination.

* Comments corrected per review comments.

* Changes per comments; No logical code changes.

* s/ASSERT/ABORT_IF_NOT/ per review comments.

* Initialize counter, so when we fail and get core, we could assess, if it was written to or not.

* Made it better per review comments. No logical change in code.

### DIFF
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -17,6 +17,21 @@
 
 namespace swss {
 
+void err_exit(const char *fn, int ln, int e, const char *fmt, ...)
+{
+    va_list ap;
+    char buff[1024];
+    size_t len;
+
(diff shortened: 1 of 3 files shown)


yxieca:
Can I cherry-pick #286 directly? Or is there merge conflict?

renukamanavalan:
Closing in favor of cherry pick #286

PR #290, opened 2019-06-27
fix bug that Executor does not get correct priority saved in m_select…

jihaix:
1、What I did
changes in sonic-swss-common: make member function `Select::getPri() `to be virtual
changes in sonic-swss: override `getPri()` in class Executor to return priority of m_selectables

2、Why I do it
`Select:cmp` using priority got by function `getPri()` to sort order of event. But class Select's member function `getPri() `is not virtual，so that class Executor can't not override getPri()  to return priority saved in m_selectable. Class Executor  is just a container of pointer m_selectable with type `Select *`.  We should  get priority saved in m_selectable rather than 0, which is default structed by Executor's base class Select.  Thinking of below situation ,  pointer `Select * s` point to object of type Executor,` s -> getPri()` will always return priority of 0

3、How I verify it 
Debug & Test， show priority of Executor Object

### DIFF
--- a/common/selectable.h
+++ b/common/selectable.h
@@ -53,7 +53,7 @@ public:
     {
     }
 
-    int getPri() const
+    virtual int getPri() const
     {
         return m_priority;
     }


jihaix:
pull request in sonic-swss  
https://github.com/Azure/sonic-swss/pull/957/files

common/selectable.h
```diff
@@ -53,7 +53,7 @@ class Selectable
     {
     }
 
-    int getPri() const
+    virtual int getPri() const
```
pavel-shirshov: Why do you need to make this method virtual?
Are you going to change implementation of the method in child classes?

jihaix: yes, I'm going to change implementation of the method in class Executor. m_selectable is the Selectable pointer that make sense， but Executor can't get priority of m_selectable, it can only get priority of its base class, which is always 0.
Below is definition of Executor
```
class Executor : public Selectable
{
public:
    Executor(Selectable *selectable, Orch *orch, const string &name)
        : m_selectable(selectable)
        , m_orch(orch)
        , m_name(name)
    {
    }
    ......
    int getFd() override { return m_selectable->getFd(); }
    void readData() override { m_selectable->readData(); }
    bool hasCachedData() override { return m_selectable->hasCachedData(); }
    bool initializedWithData() override { return m_selectable->initializedWithData(); }
    void updateAfterRead() override { m_selectable->updateAfterRead(); }
  ```

Comment (pavel-shirshov) [reply common/selectable.h]: in the definition of Executor I don't see and getPri() method.
How it could be possible to executor to read 0 using Selectable getPri() method?

Comment (jihaix) [reply common/selectable.h]: getPri() is defined in its base clase Selectable, it return the value of m_priority, which was default set as 0
```
class Selectable
{
public:
    Selectable(int pri = 0) : m_priority(pri),
                              m_last_used_time(std::chrono::steady_clock::now()) {}
......
    int getPri() const
    {
        return m_priority;
    }
    ......
```

PR #295, opened 2019-07-09
timerfd:read failure - Record in logs as error.

renukamanavalan:
The return value of 0, with errno=0 is treated as benign, as this appears to be a Hardware related kernel bug.
NOTE: This bug is found to occur in S6100 platforms only.

### DIFF
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -107,6 +107,11 @@ void SelectableTimer::readData()
     ABORT_IF_NOT((s == sizeof(uint64_t)) || ((s == 0) && (errno == 0)),
             "Failed to read timerfd. s=%ld", s)
 
+    if (s != sizeof(uint64_t)) {
+        SWSS_LOG_ERROR("Benign failure to read from timerfd return=%d error=%s",
+                s, strerror(errno));
+    }
+
     // r = count of timer events happened since last read.
 }
 
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -108,8 +108,7 @@ void SelectableTimer::readData()
             "Failed to read timerfd. s=%ld", s)
 
     if (s != sizeof(uint64_t)) {
-        SWSS_LOG_ERROR("Benign failure to read from timerfd return=%zd error=%s",
-                s, strerror(errno));
+        SWSS_LOG_ERROR("Benign failure to read from timerfd return=%zd Expect: 8", s);
     }
 
     // r = count of timer events happened since last read.
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -108,7 +108,7 @@ void SelectableTimer::readData()
             "Failed to read timerfd. s=%ld", s)
 
     if (s != sizeof(uint64_t)) {
-        SWSS_LOG_ERROR("Benign failure to read from timerfd return=%zd Expect: %d",
+        SWSS_LOG_ERROR("Benign failure to read from timerfd return=%zd Expect: %zd",
                     s, sizeof(uint64_t));
     }
 
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -86,7 +86,7 @@ void SelectableTimer::readData()
 
     /*
      * By right or most likely s = 8 here.
-     * Else in failure case, s = -1 with errno != 0, b
+     * Else in failure case, s = -1 with errno != 0
      * But due to a (HW influenced) Kernel bug, it could be s=0 & errno=0
      *
      * This bug has been observed in S6100 only.
@@ -104,8 +104,8 @@ void SelectableTimer::readData()
      *  
      *  The behavior of read returning 0, with errno=0 is an unexpected behavior.
      */
-    ABORT_IF_NOT((s == sizeof(uint64_t)) || ((s == 0) && (errno == 0)),
-            "Failed to read timerfd. s=%ld", s)
+    ABORT_IF_NOT((s == sizeof(uint64_t)) || (s == 0),
+            "Failed to read timerfd. s=%ld", s);
 
     if (s != sizeof(uint64_t)) {
         SWSS_LOG_ERROR("Benign failure to read from timerfd return=%zd Expect: %zd",
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -108,7 +108,7 @@ void SelectableTimer::readData()
             "Failed to read timerfd. s=%ld", s)
 
     if (s != sizeof(uint64_t)) {
-        SWSS_LOG_ERROR("Benign failure to read from timerfd return=%d error=%s",
+        SWSS_LOG_ERROR("Benign failure to read from timerfd return=%zd error=%s",
                 s, strerror(errno));
     }
 
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -108,7 +108,8 @@ void SelectableTimer::readData()
             "Failed to read timerfd. s=%ld", s)
 
     if (s != sizeof(uint64_t)) {
-        SWSS_LOG_ERROR("Benign failure to read from timerfd return=%zd Expect: 8", s);
+        SWSS_LOG_ERROR("Benign failure to read from timerfd return=%zd Expect: %d",
+                    s, sizeof(uint64_t));
     }
 
     // r = count of timer events happened since last read.
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -84,32 +84,31 @@ void SelectableTimer::readData()
     }
     while(s == -1 && errno == EINTR);
 
-    /*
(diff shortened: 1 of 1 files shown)


common/selectabletimer.cpp
```diff
@@ -107,6 +107,11 @@ void SelectableTimer::readData()
     ABORT_IF_NOT((s == sizeof(uint64_t)) || ((s == 0) && (errno == 0)),
             "Failed to read timerfd. s=%ld", s)
 
+    if (s != sizeof(uint64_t)) {
```
jleveque: Will execution ever reach this? I believe your new `ABORT_IF_NOT` macro will cause the program to exit if `s != sizeof(uint64_t)` above, correct?

renukamanavalan: Added one more condition to pass as "s==0 && errno==0"

renukamanavalan: As we see it as kernel+HW bug, specific to S6100

common/selectabletimer.cpp
```diff
@@ -107,6 +107,11 @@ void SelectableTimer::readData()
...
+        SWSS_LOG_ERROR("Benign failure to read from timerfd return=%zd error=%s",
+                s, strerror(errno));
+    }
+
```
pavel-shirshov: errno contains correct information only when s == -1
http://man7.org/linux/man-pages/man2/read.2.html#RETURN_VALUE

common/selectabletimer.cpp
jleveque: I see. I guess it's just that the logic is a bit unclear at first glance.

common/selectabletimer.cpp
jleveque: True, and we know that we will only execute this line if `errno == 0`, so logging errno here isn't really helpful anyway.

renukamanavalan: You are right. For that matter, even s value need not be printed, as it has to be 0. But this is explicit only when you read code. 

For some one reading the logs only, try to give them complete data.

renukamanavalan: Agree, errno does not make sense with s >= 0. Let me drop it.

jleveque: Maybe also add something to the log message about this being an apparent kernel bug?

renukamanavalan: I have an elaborate comments in the code. Do you prefer in log message too?

renukamanavalan: Again this is not really kernel bug. It is an HW influenced bug, found in S6100 only,

jleveque: Understood. I just feel like something like this should be caught in the kernel. :)

common/selectabletimer.cpp
```diff
@@ -107,6 +107,10 @@ void SelectableTimer::readData()
     ABORT_IF_NOT((s == sizeof(uint64_t)) || ((s == 0) && (errno == 0)),
             "Failed to read timerfd. s=%ld", s)
 
+    if (s != sizeof(uint64_t)) {
+        SWSS_LOG_ERROR("Benign failure to read from timerfd return=%zd Expect: 8", s);
```
jleveque: Instead of hardcoding 8, might be better to log `sizeof(uint64_t)`

common/selectabletimer.cpp
renukamanavalan: There is a trade off. Being timer related, with high frequency hit, being too careful can be expensive.

common/selectabletimer.cpp
renukamanavalan: Sure. I was kind of debating :-)

common/selectabletimer.cpp
```diff
@@ -108,8 +108,7 @@ void SelectableTimer::readData()
             "Failed to read timerfd. s=%ld", s)
```
pavel-shirshov: (s == 0) && (errno == 0)
errno doesn't have any sense with s== 0

jleveque: This is the case of the hardware bug that this log message is catching.

renukamanavalan: dropped.

renukamanavalan: Sorry, I misread your comments. You are correct. But as Joe points, we are dealing with a bug here. 

In the same line, it does not make sense to get a return value of 0, for a blocking-fd.

So I prefer to leave it as such.

pavel-shirshov: > RETURN VALUE         top
       On success, the number of bytes read is returned (zero indicates end
       of file), and the file position is advanced by this number.  It is
       not an error if this number is smaller than the number of bytes
       requested; this may happen for example because fewer bytes are
       actually available right now (maybe because we were close to end-of-
       file, or because we are reading from a pipe, or from a terminal), or
       because read() was interrupted by a signal.  See also NOTES.

       On error, -1 is returned, and errno is set appropriately.  In this
       case, it is left unspecified whether the file position (if any)
       changes.

errno is set only when s == -1.
Otherwise errno is set to something else, the result of previous operation which is not defined here

pavel-shirshov: > This is the case of the hardware bug that this log message is catching.

How do the hardware bug could change the read API call?

common/selectabletimer.cpp
```diff
@@ -104,8 +104,13 @@ void SelectableTimer::readData()
      *  
      *  The behavior of read returning 0, with errno=0 is an unexpected behavior.
      */
-    ABORT_IF_NOT((s == sizeof(uint64_t)) || ((s == 0) && (errno == 0)),
-            "Failed to read timerfd. s=%ld", s)
+    ABORT_IF_NOT((s == sizeof(uint64_t)) || (s == 0),
```
jleveque: I feel like this could be made more understandable as follows:

```
    if (s == 0)) {
        SWSS_LOG_ERROR("Benign failure to read from timerfd return=%zd Expect: %zd",
                    s, sizeof(uint64_t));
        return;
    }

    ABORT_IF_NOT((s == sizeof(uint64_t)), "Failed to read timerfd. s=%ld", s);

```

pavel-shirshov:
It's not "expected". That is clearly defined in API:
http://man7.org/linux/man-pages/man2/read.2.html#RETURN_VALUE
```
       On success, the number of bytes read is returned (zero indicates end
       of file), and the file position is advanced by this number.  It is
       not an error if this number is smaller than the number of bytes
       requested; this may happen for example because fewer bytes are
       actually available right now (maybe because we were close to end-of-
       file, or because we are reading from a pipe, or from a terminal), or
       because read() was interrupted by a signal.  See also NOTES.

       On error, -1 is returned, and errno is set appropriately.  In this
       case, it is left unspecified whether the file position (if any)
       changes.
```

jleveque:
@pavel-shirshov: Understood that read can return >= 0 on success. The problem is with the timerfd, which should only ever return `sizeof(uint64_t)`. If read returns 0, when we can treat it as if it was an EOF, and ignore it. However, if it returns 1 <= retval < sizeof(uint64_t), instead of aborting as Renuka is doing, should we assume that the read() call was interrupted and attempt to read the remaining (sizeof(uint64_t) - retval) bytes?

From http://man7.org/linux/man-pages/man2/timerfd_create.2.html:
```
Operating on a timer file descriptor
       The file descriptor returned by timerfd_create() supports the follow‐
       ing operations:

       read(2)
              If the timer has already expired one or more times since its
              settings were last modified using timerfd_settime(), or since
              the last successful read(2), then the buffer given to read(2)
              returns an unsigned 8-byte integer (uint64_t) containing the
              number of expirations that have occurred.  (The returned value
              is in host byte order—that is, the native byte order for inte‐
              gers on the host machine.)

              If no timer expirations have occurred at the time of the
              read(2), then the call either blocks until the next timer
              expiration, or fails with the error EAGAIN if the file
              descriptor has been made nonblocking (via the use of the
              fcntl(2) F_SETFL operation to set the O_NONBLOCK flag).

              A read(2) fails with the error EINVAL if the size of the sup‐
              plied buffer is less than 8 bytes.

              If the associated clock is either CLOCK_REALTIME or
              CLOCK_REALTIME_ALARM, the timer is absolute
              (TFD_TIMER_ABSTIME), and the flag TFD_TIMER_CANCEL_ON_SET was
              specified when calling timerfd_settime(), then read(2) fails
              with the error ECANCELED if the real-time clock undergoes a
              discontinuous change.  (This allows the reading application to
              discover such discontinuous changes to the clock.)
```

pavel-shirshov:
@Joe LeVeque<mailto:jolevequ@microsoft.com>
From the linux source code I see that it’s possible for timerfd return 0:
https://elixir.bootlin.com/linux/v3.16.69/source/fs/timerfd.c#L255
The macro returns 0
And then all other conditions are false.

Initially I wanted to check errno iff read() returned -1 as API states

pavel-shirshov:
@jleveque The Linux v3.16 can return 0 as return value.
I don't think we need to work with partially returned value. How I see from the code it's always some error value, 0 or sizeof(uint64_t).

PR #300, opened 2019-08-26
Debug Framework changes for dump routines and custom assert

anilsadineni:
These changes along with changes in sonic-utilities are needed to test the framework

The code is inline with design from:
https://github.com/Azure/SONiC/pull/398


common/debugframework.h
```diff
@@ -0,0 +1,116 @@
...
+    enum AssertAction
+    {
+        SYSLOG,
+        DUMP,
+        BTRACE,
+        ABORT,
+        UNSPECIFIED
+    };
+    static void setAssertAction(std::string val);
+    static void custom_assert(bool exp, AssertAction act, const char *func, const unsigned line);
```
qiluo-msft: >custom_assert [](start = 16, length = 13)

customAssert

common/debugframework.h
```diff
@@ -0,0 +1,116 @@
...
+#include <thread>
+#include <assert.h>
+#include <execinfo.h>
+#include "dbconnector.h"
+#include "table.h"
+
+
+#ifdef assert
+#undef assert
+#define assert(exp) Debugframework::custom_assert(exp, Debugframework::AssertAction::UNSPECIFIED, __PRETTY_FUNCTION__, __LINE__)
```
qiluo-msft: >assert [](start = 8, length = 6)

Why re-define standard assert macro? Originally assert will only take effect when build in DEBUG mode, and totally optimized in RELEASE mode. Your use case is different. You may choose another name.

common/debugframework.h
```diff
@@ -0,0 +1,116 @@
...
+
+namespace swss {
+
+#define SWSS_DEBUG_PRINT          swss::Debugframework::getInstance().dumpWrite
+#define SWSS_DEBUG_PRINT_BEGIN    swss::Debugframework::getInstance().dumpBegin
+#define SWSS_DEBUG_PRINT_END      swss::Debugframework::getInstance().dumpEnd
+
+typedef void (*DumpInfoFunc)(std::string, KeyOpFieldsValuesTuple);
+
+class Debugframework
```
qiluo-msft: >Debugframework [](start = 6, length = 14)

DebugFramework

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+
+Debugframework &Debugframework::getInstance()
+{
+    static Debugframework m_debugframework;
+    return m_debugframework;
+}
+
+Debugframework::Debugframework() {
+    // Read .ini configuration file to init
```
qiluo-msft: >Read .ini configuration file to init [](start = 7, length = 36)

Is the comment related to code?

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+    DBConnector db(CONFIG_DB, DBConnector::DEFAULT_UNIXSOCKET, 0);
+    m_configParams = std::unique_ptr<Table>(new Table(&db, APP_DEBUGFM_CONFIG_TABLE_NAME));
+
+}
+
+Debugframework::~Debugframework() {
+    //m_registeredComps.clear();
+    //m_configParams.clear();
```
qiluo-msft: remove dead code

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+}
+
+void Debugframework::linkWithFrameworkNoThread(std::string &componentName)
```
qiluo-msft: >std::string & [](start = 47, length = 13)

`const std::string &` or `std::string`

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+    }
+    
+    // Add the component to the registry
+    updateRegisteredComponents(componentName);
+    
+    // Now create a thread and let the rest be handled there
+    std::thread t(Debugframework::runnableThread, componentName, funcPtr);
+    t.detach();
+
+}
```
qiluo-msft: remove extra empty line

common/schema.h
```diff
@@ -44,6 +44,11 @@ namespace swss {
 #define APP_SFLOW_TABLE_NAME              "SFLOW_TABLE"
 #define APP_SFLOW_SESSION_TABLE_NAME      "SFLOW_SESSION_TABLE"
 #define APP_SFLOW_SAMPLE_RATE_TABLE_NAME  "SFLOW_SAMPLE_RATE_TABLE"
+#define APP_DEBUGFM_CONFIG_TABLE_NAME     "DEBUGFM_CONFIG_TABLE"
```
qiluo-msft: >APP [](start = 8, length = 3)

Maybe the tables should be moved to ConfigDB.

common/debugframework.h
```diff
@@ -0,0 +1,116 @@
...
+    
+    std::vector<std::string> m_registeredComps;
+    std::map<std::string, std::fstream*> m_compFds;
+    std::unique_ptr<Table>  m_configParams;
+    
+    TargetLocation     m_targetLocation;
+    PostAction         m_postAction;
+
+#ifndef NO_RET_TYPE
+#define NO_RET_TYPE [[noreturn]]
```
qiluo-msft: >[[noreturn]] [](start = 20, length = 12)

The original format is good enough. No need to define a macro.

common/debugframework.h
qiluo-msft: >  [](start = 20, length = 1)

an extra trailing blank

common/debugframework.h
```diff
@@ -0,0 +1,116 @@
...
+#define SWSS_DEBUG_PRINT          swss::Debugframework::getInstance().dumpWrite
+#define SWSS_DEBUG_PRINT_BEGIN    swss::Debugframework::getInstance().dumpBegin
+#define SWSS_DEBUG_PRINT_END      swss::Debugframework::getInstance().dumpEnd
+
+typedef void (*DumpInfoFunc)(std::string, KeyOpFieldsValuesTuple);
+
+class Debugframework 
+{
+public:
+    static Debugframework  &getInstance();       /* To have only one instance aka singleton */
```
qiluo-msft: >  [](start = 26, length = 1)

an extra blank

common/debugframework.h
```diff
@@ -0,0 +1,116 @@
...
+
+    //typedef std::function< void (std::string componentName, KeyOpFieldsValuesTuple args)> DumpInfoFunc;
+
```
qiluo-msft: remove dead code

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+#include "schema.h"
+#include "select.h"
+#include "dbconnector.h"
+#include "redisclient.h"
+#include "producerstatetable.h"
+#include "subscriberstatetable.h"
+#include "notificationconsumer.h"
+#include "notificationproducer.h"
+
+#define ACTIONS_CONFIG_FILE "/etc/sonic/dbgFm.ini"
```
qiluo-msft: >ACTIONS_CONFIG_FILE [](start = 8, length = 19)

How is the macro used?

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+            for (auto entry: entries)
+            {
+               std::string key = kfvKey(entry);
+               if (key == componentName)
+               {
+                   try{
+                       (*funcPtr)(componentName, entry);
+                   }catch(...){
+                       throw "Registered dump routine returned error";
```
qiluo-msft: >returned error [](start = 54, length = 14)

throw an exception

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+                   }
+
+                   /* Inform done with same key */
+                   FieldValueTuple fvResp("RESPONSE", "SUCCESS");
```
qiluo-msft: >SUCCESS [](start = 55, length = 7)

When exception thrown, still SUCCESS?

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+void Debugframework::dumpBegin(std::string &component)
+{
+    auto& dbgfm = getInstance();
+    Debugframework::TargetLocation t = dbgfm.getTarget();
+
+    if (t == SWSS_FILE)
+    {
+        // open a file in append mode and keep it open
+        std::fstream *outfile;
+        std::string options("/var/log/" + component + "_debug.log");
```
qiluo-msft: >/var/log/ [](start = 29, length = 9)

Don't use fixed path, you should have a configuration for it.

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+    va_start(ap, fmt);
+
+    if (dbgfm.getTarget() == SWSS_SYSLOG)
+    {
+        vsyslog(LOG_DEBUG, fmt, ap);
+        va_end(ap);
+    }
+    else
+    {
+        char buffer[0x1000];
```
qiluo-msft: >0x1000 [](start = 20, length = 6)

Better to allocate on heap

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+        vsnprintf(buffer, 0x1000, fmt, ap);
+        va_end(ap);
+
+        std::fstream *f;
+        auto itr = dbgfm.m_compFds.find(component);
+        if (itr != dbgfm.m_compFds.end())
+        {
+            f = itr->second;
+            *f << buffer << "\n";
```
qiluo-msft: >"\n" [](start = 28, length = 4)

If you need flush, use `endl`

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+        case SYSLOG:
+            break;
+
+        // break not added intentionally
+        case DUMP:
+            {
+                std::string comp("all");
+                std::string args("DETAIL:FULL;");
+                invokeTrigger(comp, args);
+            }
```
qiluo-msft: `break` ?

common/debugframework.h
anilsadineni: The basic idea here is that without having to change the code where assert has been used, bring customAssert and have this enabled even in production builds. Based on the configuration either log or collect more info or abort shall be action when assert condition is met and hence standard assert is re-defined.

common/debugframework.cpp
anilsadineni: It is now deadcode.  The original plan was to have a init config file but since this is library and not daemon it did not make lot of sense to have a config file.  Shall remove.

common/schema.h
anilsadineni: This was discussed in the design review meeting. This is really only on a trigger and hence considered to place it in the APP_DB.

common/debugframework.cpp
anilsadineni: Same as above. With library doesnt make lot of sense. Shall remove this deadcode.

common/debugframework.cpp
anilsadineni: Oops! response failure in the catch is missed. Shall correct it

common/debugframework.cpp
anilsadineni: This is short lived and called several times and trying to get from heap might be really costly.

common/debugframework.cpp
anilsadineni: As you see comment in line 427 break is not added intentionally, dump is exhaustive info and that should include backtrace as well

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+}
+
+void Debugframework::linkWithFrameworkNoThread(std::string &componentName)
+{
+    updateRegisteredComponents(componentName);
+}
+
+void Debugframework::linkWithFramework(std::string &componentName, const DumpInfoFunc funcPtr)
+{
+    if (funcPtr == NULL)
```
kcudnik: missing SWSS_LOG_ENTER()

common/debugframework.cpp
```diff
@@ -0,0 +1,455 @@
...
+    auto& dbgfm = getInstance();
+    Debugframework::TargetLocation t = dbgfm.getTarget();
+
+    if (t == SWSS_FILE)
+    {
+        // open a file in append mode and keep it open
+        std::fstream *outfile;
+        std::string options("/var/log/" + component + "_debug.log");
+        try{
+            outfile = new std::fstream();
```
kcudnik: why this is a pointer ? not actual object? on s tack ?

common/debugframework.h
```diff
@@ -0,0 +1,116 @@
...
+    std::map<std::string, std::fstream*> m_compFds;
+    std::unique_ptr<Table>  m_configParams;
+    
+    TargetLocation     m_targetLocation;
+    PostAction         m_postAction;
+
+#ifndef NO_RET_TYPE
+#define NO_RET_TYPE [[noreturn]]
+#endif
+    NO_RET_TYPE static void runnableThread(std::string componentName, const DumpInfoFunc funcPtr);
```
kcudnik: if this is thread function, threads should be able to quit loop in nice way, for example in dtr there should be join thread and break run flag in thread

common/debugframework.h
```diff
@@ -0,0 +1,116 @@
...
+    static void custom_assert(bool exp, AssertAction act, const char *func, const unsigned line);
+
+private: 
+    Debugframework(); 
+    ~Debugframework();
+    Debugframework(const Debugframework&);
+    Debugframework &operator=(const Debugframework&);
+
+    static void listenDoneEvents(std::string componentName);
+    static void relayTrigger(std::string &componentName, KeyOpFieldsValuesTuple args);
```
kcudnik: all paramters in functions should be defined 1 per line and with prefixes _In_ _Inout or _Out_

bhaveshdell (dismissed):

common/debugframework.cpp
anilsadineni: done

common/debugframework.cpp
anilsadineni: done

common/debugframework.cpp
anilsadineni: Added for all relevant functions except in small utility functions that get called repeatedly

common/debugframework.cpp
anilsadineni: done

common/debugframework.cpp
anilsadineni: This is already in the context of debug routines and additional exception handling is an overhead and hence not changing this for now.

common/debugframework.cpp
anilsadineni: The post scripts are TBD and for now there is nothing in configDB. When the post scripts (TBD)  are in place this needs to allign with those scripts.  Shall defer this till that point.

common/debugframework.cpp
anilsadineni: This is created during begin and kept open while the components dumps info using  multiple writes and in the end this is closed and freed.

common/debugframework.cpp
anilsadineni: done

common/debugframework.h
anilsadineni: changed this in lot of places..

anilsadineni: done

common/debugframework.h
anilsadineni: done

common/debugframework.h
anilsadineni: done

common/debugframework.h
anilsadineni: done

common/debugframework.h
anilsadineni: This comes into picture if the component registers with a thread option and the thread stays until the component exits so this really never returns and hence the noreturn flag.

common/debugframework.h
anilsadineni: I am new to project and tried to stay consistent with other files in the dir.

common/debugframework.cpp
```diff
@@ -0,0 +1,454 @@
...
+            {
+                std::string key = kfvKey(entry);
+                if (key == componentName)
+                {
+                    FieldValueTuple fvResp("RESPONSE", "SUCCESS");
+                    try{
+                        (*funcPtr)(componentName, entry);
+                    }catch(...){
+                        throw "Registered dump routine returned error";
+                        fvResp.second = "FAILURE";
```
kcudnik: will be never executed,
also please log error message if catch was catched

common/debugframework.cpp
```diff
@@ -0,0 +1,454 @@
...
+            std::fstream *out = itr->second;
+            out->close();
+            delete out;
+            dbgfm.m_compFds.erase(itr);
+        }
+    }
+    return;
+}
+
+void DebugFramework::dumpWrite(const std::string &component, const char *fmt, ...)
```
kcudnik: one param per line, also add _In_ /_Out_ etc prefixes

common/debugframework.cpp
```diff
@@ -0,0 +1,454 @@
...
+    DebugFramework::TargetLocation t = dbgfm.getTarget();
+    
+    if (t == SWSS_FILE)
+    {
+        auto itr = dbgfm.m_compFds.find(component);
+        if (itr != dbgfm.m_compFds.end())
+        {
+            std::fstream *out = itr->second;
+            out->close();
+            delete out;
```
kcudnik: use shared pointers in map/set, to not explicitly delete them

common/debugframework.cpp
```diff
@@ -0,0 +1,454 @@
...
+}
+
+void DebugFramework::setTarget(const std::string &val)
+{
+    std::string key("TARGET");
+    m_configParams->hset(key, key, val);
+}
+
+DebugFramework::PostAction DebugFramework::getPostAction()
+{
```
kcudnik: missing SWSS_LOG_ENTER(); on every function

common/debugframework.cpp
```diff
@@ -0,0 +1,454 @@
...
+            {
+                std::string comp("all");
+                std::string args("DETAIL:FULL;");
+                invokeTrigger(comp, args);
+            }
+
+        case BTRACE:
+            {
+                std::FILE *fp = fopen(ASSERT_FILE, "w");
+#define SIZE 100
```
kcudnik: please define at the file beginning and add more meaningfull name

common/debugframework.cpp
```diff
@@ -0,0 +1,454 @@
...
+{
+    DBConnector db(APPL_DB, DBConnector::DEFAULT_UNIXSOCKET, 0);
+
+    Table table(&db, APP_DEBUG_RCOMPONENT_TABLE_NAME);
+    FieldValueTuple fv("THREAD", "noOp");
+    std::vector<FieldValueTuple>fieldValues = { fv };
+    table.set(component, fieldValues);
+}
+
+NO_RET_TYPE void DebugFramework::runnableThread(const std::string componentName, const DumpInfoFunc funcPtr)
```
qiluo-msft: >const std::string [](start = 48, length = 17)

const std::string &

common/debugframework.cpp
```diff
@@ -0,0 +1,454 @@
...
+                    /* Inform done with same key */
+                    std::vector<FieldValueTuple>fieldValues = { fvResp };
+                    p.send(componentName, componentName, fieldValues);
+                }
+            }
+        }
+    }  // end while loop
+}
+
+void DebugFramework::invokeTrigger(const std::string componentName, std::string argList)
```
qiluo-msft: >const std::string [](start = 35, length = 17)

const std::string &

common/debugframework.cpp
```diff
@@ -0,0 +1,454 @@
...
+                break;
+            }
+        }
+    }
+    // Now create a thread and wait for done notifications
+    std::thread t(DebugFramework::listenDoneEvents, componentName);
+    t.detach();
+}
+
+void DebugFramework::listenDoneEvents(const std::string componentName)
```
qiluo-msft: >const std::string [](start = 38, length = 17)

const std::string &

common/debugframework.cpp
qiluo-msft: Sorry, I gave the wrong comment. In your second commit, any code after `throw` will not run, so it is dead code. Original code should be good.

common/debugframework.h
qiluo-msft: I am worried about the side-effect. Assume there is a third party header file, which uses `assert` intensively for DEBUG build. If someone include this header before third party one, then there will be overhead to RELEASE build.

---
In reply to: [328907805](https://github.com/Azure/sonic-swss-common/pull/300#discussion_r328907805) [](ancestors = 328907805)

anilsadineni: I am not sure of the overhead to release build you mention.  What would otherwise have no effect in release build will now have some outcome when assert condition is hit. Is that what you refer?  

The outcome of assert is now controlled using a default or configuration command.  The idea is to not miss out on assert but have some insight into it and the level of overhead is controlled.  So I am not sure if we can term that overhead or side-effect, this is by design the desired effect for better debug ability.

common/debugframework.cpp
anilsadineni: I have added to all the functions which does meaningful work other than helper functions and utility functions which will be called multiple times.  Please check my comment above when I changed to address your review comments

common/debugframework.cpp
anilsadineni: Are you referring to SAL annotation style of doxygen style?  As mentioned in response to your earlier comment, I tried to stay consistent with the other files in the directory and elsewhere in the project.  Can you please point me to any other file which followed your suggestion for me to be clear. Thanks,

common/debugframework.cpp
anilsadineni: Good suggestion and optimization. However for now there are no components using this framework. Once the usage picks up, we need to revisit the file naming conventions, post scripts, uploads and based on that these changes might be moot.  For now this is not in usual operations and only during debug and hence this might not be a lot of overhead even without the suggested optimization.

common/debugframework.cpp
anilsadineni: done

common/debugframework.cpp
anilsadineni: done

common/debugframework.cpp
anilsadineni: done

common/debugframework.cpp
anilsadineni: done

common/debugframework.cpp
anilsadineni: done

PR #302, opened 2019-09-05
timerfd return 0 with errno =0 -  handle as False alarm.

renukamanavalan:
timerfd read is expected to return 8 or for any other value appropriate
non-zero errno is expected. But due to a kernel bug, that is found to
occur in S6100 platform, there could be a return value of 0, with errno=0.

Per DELL, this implies timer fired but not forwarded, so take it as
false alarm, and next immediate read is expected to succeed.

Made the fix to meet it.

### DIFF
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -87,7 +87,7 @@ void SelectableTimer::readData()
              * case errno should be EAGAIN.
              *
              * Due to a kernel bug, exposed only in S6100 HW platform,
-             * this could be true.
+             * we could see s == 0 and errno == 0.
              *
              * In this case, the timer is fired, but not adanced, hence
              * another read is expected to fix it per DELL.
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -72,38 +72,22 @@ int SelectableTimer::getFd()
     return m_tfd;
 }
 
-void SelectableTimer::readData()
+int SelectableTimer::readData()
 {
-    uint64_t r = UINT64_MAX;
+    uint64_t r = 0;
 
     ssize_t s;
     errno = 0;
     do
     {
         s = read(m_tfd, &r, sizeof(uint64_t));
-        if ((s == 0) && (errno == 0)) {
-            /*
-             * s == 0 is expected only for non-blocking fd, in which
-             * case errno should be EAGAIN.
-             *
-             * Due to a kernel bug, exposed only in S6100 HW platform,
-             * we could see s == 0 and errno == 0.
-             *
-             * In this case, the timer is fired, but not advanced, hence
-             * another read is expected to fix it per DELL.
-             *
-             * Log an error and continue to read.
-             *
-             */
-            SWSS_LOG_ERROR("Benign failure to read from timerfd return=0 && errno=0. Expect: return=%zd",
-                    sizeof(uint64_t));
-        }
     }
-    while((s == -1 && errno == EINTR) || ((s == 0) && (errno == 0)));
+    while(s == -1 && errno == EINTR);
 
-    ABORT_IF_NOT(s == sizeof(uint64_t), "Failed to read timerfd. s=%zd", s);
+    ABORT_IF_NOT((s == 0) || (s == sizeof(uint64_t), "Failed to read timerfd. s=%zd", s);
 
     // r = count of timer events happened since last read.
+    return r;
 }
 
 }
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -89,7 +89,7 @@ void SelectableTimer::readData()
              * Due to a kernel bug, exposed only in S6100 HW platform,
              * we could see s == 0 and errno == 0.
              *
-             * In this case, the timer is fired, but not adanced, hence
+             * In this case, the timer is fired, but not advanced, hence
              * another read is expected to fix it per DELL.
              *
              * Log an error and continue to read.
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -84,7 +84,7 @@ uint64_t SelectableTimer::readData()
     }
     while(ret == -1 && errno == EINTR);
 
-    ABORT_IF_NOT((ret == 0) || (ret == sizeof(uint64_t), "Failed to read timerfd. ret=%zd", ret);
+    ABORT_IF_NOT((ret == 0) || (ret == sizeof(uint64_t)), "Failed to read timerfd. ret=%zd", ret);
 
     // cnt = count of timer events happened since last read.
     return cnt;
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -81,35 +81,27 @@ void SelectableTimer::readData()
     do
     {
         s = read(m_tfd, &r, sizeof(uint64_t));
+        if ((s == 0) && (errno == 0)) {
+            /*
+             * s == 0 is expected only for non-blocking fd, in which
+             * case errno should be EAGAIN.
+             *
+             * Due to a kernel bug, exposed only in S6100 HW platform,
+             * this could be true.
+             *
+             * In this case, the timer is fired, but not adanced, hence
+             * another read is expected to fix it per DELL.
+             *
+             * Log an error and continue to read.
+             *
+             */
+            SWSS_LOG_ERROR("Benign failure to read from timerfd return=0 && errno=0. Expect: return=%zd",
+                    sizeof(uint64_t));
+        }
     }
-    while(s == -1 && errno == EINTR);
+    while((s == -1 && errno == EINTR) || ((s == 0) && (errno == 0)));
 
-    if (s != sizeof(uint64_t)) {
-        /*
-         * By right or most likely s = 8 here.
-         * Else in failure case, s = -1 with errno != 0
-         * But due to a (HW influenced) Kernel bug, it could be s=0 & errno=0
-         *
-         * This bug has been observed in S6100 only.
-         * The bug incidence is pretty seldom.
-         *
-         * A short note on kernel bug:
-         *  timerfd_read, upon asserting that underlying hrtimer has triggered once,
-         *  calls hrtimer_forward_now, to get total count of expiries since timer start
-         *  to now, as read gets called asynchronously at a time later than the time
-         *  point of trigger.
-         *  The hrtimer_forward_now is expected to return >= 1, as it is certain
-         *  that one trigger/expiry is confirmed to have occurred.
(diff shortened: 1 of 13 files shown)


common/selectabletimer.cpp
```diff
@@ -81,35 +81,27 @@ void SelectableTimer::readData()
     do
     {
         s = read(m_tfd, &r, sizeof(uint64_t));
+        if ((s == 0) && (errno == 0)) {
+            /*
+             * s == 0 is expected only for non-blocking fd, in which
+             * case errno should be EAGAIN.
+             *
+             * Due to a kernel bug, exposed only in S6100 HW platform,
+             * this could be true.
```
jleveque: This comment is a bit confusing. I believe you mean to say something like "Due to a kernel bug, exposed only in S6100 HW platform, we could see s == 0 and errno == 0 with a blocking fd." Please make this comment more clear for posterity.

common/selectabletimer.cpp
```diff
@@ -81,35 +81,27 @@ void SelectableTimer::readData()
...
         s = read(m_tfd, &r, sizeof(uint64_t));
+        if ((s == 0) && (errno == 0)) {
+            /*
+             * s == 0 is expected only for non-blocking fd, in which
+             * case errno should be EAGAIN.
+             *
+             * Due to a kernel bug, exposed only in S6100 HW platform,
+             * we could see s == 0 and errno == 0.
+             *
+             * In this case, the timer is fired, but not adanced, hence
```
stcheng: advanced

stcheng:
check one typo

renukamanavalan:
Thanks. Fixed.

common/selectabletimer.cpp
```diff
@@ -81,35 +81,27 @@ void SelectableTimer::readData()
     do
     {
         s = read(m_tfd, &r, sizeof(uint64_t));
+        if ((s == 0) && (errno == 0)) {
```
pavel-shirshov: errno value is valid only when read returns -1.
```
On error, -1 is returned, and errno is set appropriately.  In this case, it is left unspecified whether the file position (if any) changes.
```
Please check `man read`

common/selectabletimer.cpp
```diff
@@ -81,35 +81,27 @@ void SelectableTimer::readData()
...
+             *
+             * Log an error and continue to read.
+             *
+             */
+            SWSS_LOG_ERROR("Benign failure to read from timerfd return=0 && errno=0. Expect: return=%zd",
+                    sizeof(uint64_t));
+        }
     }
-    while(s == -1 && errno == EINTR);
+    while((s == -1 && errno == EINTR) || ((s == 0) && (errno == 0)));
```
pavel-shirshov: How do we know that next read wouldn't be blocked after read returned 0?
If read is going to blocked, it means swss global loop will be blocked by the read here.

common/selectabletimer.cpp
renukamanavalan: We are dealing with kernel bug here. Prefer it to be explicit. 
I would leave it as such, unless there is an issue with this check.

jleveque: I see Pavel's point. If `read()` returns 0, we should probably enter this block regardless of what `errno` is set to. `read()` should only return 0 upon encountering EOF, which shouldn't occur with a timerfd. Only if `read()` returns -1 should we check to see if `errno` is set to `EAGAIN`.

In your comment you state, "s == 0 is expected only for non-blocking fd, in which case errno should be EAGAIN." Have you confirmed this? According to the man page, this is not the case; errno is only valid if `read()` returns -1.

renukamanavalan: You are correct.  s == 0, does not set errno. Or not supposed to set errno.
As you said, a blocking timerfd should never return 0, either. But it does.
As I mentioned above, being a kernel bug, I prefer to be explicit. 
Let me check with DELL guys too and drop it, if they too agree.

common/selectabletimer.cpp
aruljeniston: As we discussed here, timerfd on monotonic clock should not return 0. But due to some reason kernel returns 0. Code study shows when timerfd read() returns 0, kernel fails to move the expiry time to the next interval and restarts the timer with old expiry time. So the timer is very likely to be fired immediately. This could cause the next read to return immediately but we can't 100% affirmatively say the next read would be always non blocking. Adding to this, if the same problem happens on real-time clock due to clock changes, the next read would very likely block. Our recommendation is simply ignore if read return 0. But this API alone may not be able to  do it.

Note:- we have already confirmed the possibility of returning 0 for real-time clock. Linux community is working on it.

aruljeniston: - On error case, SelectableTimer::readData() loops over timerfd read() till EINTR error gets cleared. Wouldn't the second read block in this case? 
- do we see a possibility to timerfd read() to set EINTR error. I don't see it man page.

renukamanavalan: Hi all,

**Facts:**
1) We are facing **a kernel bug,** which means, we see a behavior that is **different** from what is stated in **'man pages'**.

2) Yes, man page for [timerfd](http://man7.org/linux/man-pages/man2/timerfd_create.2.html), does not mention about EINTR.  

**EINTR:**
If EINTR is not expected in this call, the loop that checks for EINTR would never happen. The man page does not explicitly say that EINTR is *not* expected, either. So I prefer to leave it there, which would be a no-op, if we never get one. As we check "s==-1" first, we use the value in cache first and only then fetch errno, as needed, so may not have any additional cost either.

**Question**:
Currently the code identifies this kernel bug as "(s == 0) && (errno == 0)".
As suggested by reviewer comments, would it be  OK to check **"(s == 0)"** only.
If I don't get any comments, let me take it as "yes", so **please comment, if "no"**

Ways to handle kernel error in SelectableTimer::readData, **when this bug occurs**:
1) Go back to read() as in this PR.
    **pro**: Simple fix in one place that is transparent to callers of this API
    **con**: If our code reading is incorrect, we could block for another N seconds. 
            By theory N can be any long. I see some code calling for 5 Mins and future code
            updates could call for even longer. If this would happen, this could be
            one nightmare for debugging, as the orchagent freezing would be
            transient and seldom.

2) Change the API signature from 'void' return value to 'int' return value and return 0.
    **Pro**: API is clean and convey the false alarm out
    **Con**: Touch every code that uses this timer to handle the false alarm.
             But the higher probability is that majority of the callers (if not all)
             of this SelectableTimer class do not care about being called more than
             once within the N seconds. Just the fact that the API has been void
             implies that.
             This means, that only few code would need to be adjusted,
             Said that, still someone has to vet all the callers.

3) No code change; Declare in header the possibility of false alarm.
    Pro: Simple, if all callers are impervious to the effect of being called more than once within N.
    Con: Future code updates could be prone, if devs don't read the header file thoroughly.

I am tending towards 2.
Please speak up, if you disagree. Would like to wind this up soon, as we would like this fix in 201811 image.

pavel-shirshov:
@aruljeniston
Do we know the reasons why the kernel returns 0? Until we know this, we can't fix anything and be sure everything will work.
I checked the kernel code and I saw that it's impossible to get 0.
But we getting 0, which means either our compiler generated bad code or the platform works incorrectly. Can we investigate why it's possible to get 0?
I can provide results I found in the kernel code.

pavel-shirshov:
@renukamanavalan Do you know what kernel bug exactly we're facing? Can we fix it?

renukamanavalan:
@pavel-shirshov 
You have already stated the bug, "_Do we know the reasons why the kernel returns 0_".

timerfd::read is expected to return non-zero value or error (ret = -1 && errno != 0). But we do get return value of 0. 
Q: How do we handle it (as there is no kernel fix) ?
Provided a couple of ways above. 

If you have some suggestions on other ways of working around it, please provide.

aruljeniston:
"Do we know the reasons why the kernel returns 0".
- Kernel could return 0 for real-time TFD_TIMER_ABSTIME timers. It has been communicated to Linux community and they are planning to change the man page of timerfd for real-time timers. 
you may read the last two posts of this mail thread https://lists.gt.net/linux/kernel/3435187 . 
- We do see a possibility in the code where timerfd::read() could return 0 on monotonic timers but Linux community says it is only possible if hardware clocks (tsc in this case) are not in sync. We have been running several tests but haven't seen any problem in the hardware so far.
- Our preference is to pick option 2 as it doesn't have any side effect.

pavel-shirshov:
@aruljeniston 
Thank you for the information.
We have two things here:
1. The poll() returned us that a timer object fired, but when we read, we got 0.
2. read() returned 0.
I read the thread and found that you didn't discuss 1., but only 2.
Do we know when poll could return us timer event?
1. When tsc returned older value then expected?
2. When the timer expired?
3. In any other time?

Thanks

aruljeniston:
>Do we know when poll could return us timer event?
Yes. poll() returns timer event when timer expired or cancelled.

1. After timer expired, poll() returns EPOLLIN event.
2. Eventually application calls read(). 
3. During read(), due to unknown reason, timekeeper monotonic clock returns value lesser than the expiry time. This causes read() to return 0.

pavel-shirshov:
How do we know we got a false alarm?
I think how the things are going:
1. Our timer expired. The event is fired
2. Our tsc goes back for some reason.
3. When we read the event, the event wasn't fired yet. But it's going to be fired again.
So I think our first alarm with read() = 0 is correct alarm, but the second one is false. :)
But I can't be wrong here.

pavel-shirshov:
Why I don't like the second option: It will change our API significantly. Currently we have interface for Selectable which has void readData(). and we have many implementations for the interface. As soon as we change it, we should:
1. Implement this interface for every selectable in swss-common and swss
2. Change tests for -common
3. Change all usage for selectables. It's swss and possibly a lot of python utilities.

It's ok for changes for our master, but not for stable branch.
And we need the fixed for the stable branch

common/selectabletimer.cpp
```diff
@@ -72,38 +72,22 @@ int SelectableTimer::getFd()
     return m_tfd;
 }
 
-void SelectableTimer::readData()
+int SelectableTimer::readData()
 {
-    uint64_t r = UINT64_MAX;
+    uint64_t r = 0;
```
jleveque: Since you're refactoring, I suggest you replace `r` and `s` with more descriptive variable names.

renukamanavalan: Just realized little more work to do, as the base class declares the readData. Need to see all users of base class Selectable

lguohan:
build failure can you check?

renukamanavalan:
Sure. Soon

renukamanavalan:
Corresponding swss changes will be soon in a PR for sonic-swss.

PR #310, opened 2019-09-21
Threshold feature schema changes.

shdasari:
The threshold feature allows users to configure a threshold on specific port's priority-group or queue buffers and be notified when a breach is seen on the buffer usage. Please refer to threshold feature at https://github.com/Azure/SONiC/pull/429/commits/396207404f37f4684606592d097abbf15e2ad0a7#diff-8c1cef396dfe5dd465889eb11be4106c. 

These changes bring in schema changes for the threshold feature.


common/schema.h
```diff
@@ -83,6 +83,7 @@ namespace swss {
 #define COUNTERS_SWITCH_NAME_MAP            "COUNTERS_SWITCH_NAME_MAP"
 #define COUNTERS_DEBUG_NAME_PORT_STAT_MAP   "COUNTERS_DEBUG_NAME_PORT_STAT_MAP"
 #define COUNTERS_DEBUG_NAME_SWITCH_STAT_MAP "COUNTERS_DEBUG_NAME_SWITCH_STAT_MAP"
+#define COUNTERS_THRESHOLD_BREACH_TABLE     "THRESHOLD_BREACH_TABLE"
```
lguohan: remove the TABLE suffix

common/schema.h
```diff
@@ -164,6 +165,7 @@ namespace swss {
 #define CFG_PFC_WD_TABLE_NAME         "PFC_WD"
 #define CFG_FLEX_COUNTER_TABLE_NAME "FLEX_COUNTER_TABLE"
 #define CFG_WATERMARK_TABLE_NAME "WATERMARK_TABLE"
+#define CFG_THRESHOLD_TABLE_NAME      "THRESHOLD_TABLE"
```
lguohan: for CFG db, we'd like to remove TABLE suffix as well.

common/schema.h
shdasari: Uploaded diffs with review comment addressed.

common/schema.h
shdasari: Uploaded diffs with review comment addressed.

shdasari: Also, have update sonic-utilities code changes based upon these new defines.

shdasari:
Changes being tracked via https://github.com/Azure/sonic-swss-common/pull/433.

PR #312, opened 2019-10-11
Add VRF object table in state_db

tylerlinp:
What I did
Add VRF object table in state_db

Why I did it
To record vrforch entry status for synchronize

How I verified it
Run vs tests and ansible tests

Details if related


tylerlinp:
[[VRF]: submit vrf feature  #943
](https://github.com/Azure/sonic-swss/pull/943)

PR #315, opened 2019-11-13
[MultiDB] libswsscommon.deb install default database_config.json if it doesn't exist

dzhangalibaba:
* copy database_config.json into /var/run/redis/sonic-db/ when  install libswsscommon.deb  in postinst script
* this is used when the environment has no database docker running which means NO database_config,json, mostly it is used in testing environment
* after this changes, we can do the C++ API and redis-cli replacement without affect VS test since the config file is there.


dzhangalibaba:
@qiluo-msft

dzhangalibaba:
retest this please

debian/postinst
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# This `DEBIAN/postinst` script is run post-installation
+DST_PATH="/var/run/redis/sonic-db"
+DST_FILE="database_config.json"
+SRC_PATH="/usr/share/swss"
+# if there is no $DST_FILE, it is for testing, it is needed to copy one default $DST_FILE
```
qiluo-msft: >if there is no $DST_FILE, it is for testing, it is needed to copy one default $DST_FILE [](start = 2, length = 87)

For vs test, libswsscommon is installed on the host, and python code will call it. Is your PR for this case? If yes, it does not make sense to `cp` to host DST_PATH.

dzhangalibaba: yes . it is for vs test cases. cp file  on host:DST_PATH make sense. 

when conftest.py run vs, it mount host:/var/run/redis-vs/{sw_name} to vs:var/run/redis/, so we need to copy the host:/var/run/redis/sonic-db/database_config.json to host:/var/run/redis-vs/{sw_name}/sonic-db/ and the vs can use this database_config.json file. thsi is done at https://github.com/Azure/sonic-swss/pull/1131

I am not sure why the mount name is different from /var/run/redis/, if it is the same like /var/run/redis/:/var/run/redis/, then we don't need the copy and just this PR is enough.

dzhangalibaba:
will do changes in docker-sonic-vs images itself instead of using libswsscommon

PR #316, opened 2019-11-21
[MultiDB]: apply new APIs in swss-common

dzhangalibaba:
* replacing old APIs with new APIs in swsscommon

### DIFF
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -91,7 +91,7 @@ void Logger::linkToDbWithOutput(const std::string &dbName, const PriorityChangeN
 
     // Initialize internal DB with observer
     logger.m_settingChangeObservers.insert(std::make_pair(dbName, std::make_pair(prioNotify, outputNotify)));
-    DBConnector db(LOGLEVEL_DB, DBConnector::DEFAULT_UNIXSOCKET, 0);
+    DBConnector db("LOGLEVEL_DB", 0);
     RedisClient redisClient(&db);
     auto keys = redisClient.keys("*");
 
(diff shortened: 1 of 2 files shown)


dzhangalibaba:
@qiluo-msft  FYI

qiluo-msft:
I will approve this after https://github.com/Azure/sonic-swss/pull/1138 passes vs test

dzhangalibaba:
> I will approve this after [Azure/sonic-swss#1138](https://github.com/Azure/sonic-swss/pull/1138) passes vs test

@qiluo-msft https://github.com/Azure/sonic-swss/pull/1138 mergered, we can start on this now

qiluo-msft:
Another comment from @kcudnik 
https://github.com/Azure/sonic-swss-common/pull/318#issuecomment-560337480
> adding "/var/run/redis/sonic-db/database_config.json" to build docker should solve the issue, but also ".at()" should be used on map to get database instances to force those will exists in map, and also throw exception should be added if database config json files not exists

dzhangalibaba:
> Another comment from @kcudnik
> [#318 (comment)](https://github.com/Azure/sonic-swss-common/pull/318#issuecomment-560337480)
> 
> > adding "/var/run/redis/sonic-db/database_config.json" to build docker should solve the issue, but also ".at()" should be used on map to get database instances to force those will exists in map, and also throw exception should be added if database config json files not exists

@qiluo-msft @kcudnik @lguohan 
adding db config file into build docker definitely will solve the issue. Now the question is "do we really want to add it into build docker ?"  What's the difference between build docker having a default db config file at /var/run/redis/sonic-db/ and submodule Makefile coping(mkdir/mkdir) a db config file into /var/run/redis/sonic-db/  at build time ? They all changed the building system. The feedback at sairedis PR suggest NOT copy db config file into build system.  

BTW , I am happy to have a db config by default in build docker, in this way, we don't need to worry for all the submodules who might miss db config file when doing some tests while building time. It will save my life. Can we go ahead with this approach ? Let me know and I can go head to adding it into build docker OR change each individual test case to have this db config file while building.

kcudnik:
seems fine to me, that config needs to be copied to build docker anyway
but you can still address at() methods and file open check if file exists

dzhangalibaba:
> seems fine to me, that config needs to be copied to build docker anyway
> but you can still address at() methods and file open check if file exists

yes. I can start address the at() and file checking first. 

Let's see what's the opinion on adding db config file into build docker from Guohan and Qi. Then we can what's to do on this next.

PR #317, opened 2019-12-01
Embed log level info in every log line

kcudnik:
Some syslog configurations may not put log level in log file


qiluo-msft:
What is the motivation? For your case, how about fixing syslog configurations?

kcudnik:
on systems with "sysklogd" instead of syslog/rsyslog, format template of log lines can't be changed therefore log priority is not included in syslog lines, that's why i propose here to add error level here

qiluo-msft:
I feel it is annoying for normal user just using default rsyslogd. I am ok if you change it only on a system with sksklogd, but don't change behavior for others.

PR #318, opened 2019-12-02
Revert "[MultiDB]: apply new APIs in swss-common"

qiluo-msft:
Reverts Azure/sonic-swss-common#316
sairedis is using logger.cpp, and it was broken in runtime.


kcudnik:
adding "/var/run/redis/sonic-db/database_config.json" to build docker should solve the issue, but also ".at()" should be used on map to get database instances to force those will exists in map, and also throw exception should be added if database config json files not exists

PR #321, opened 2019-12-04
[MultiDB]: install default db config file when installing libswsscommon if there is no such a file

dzhangalibaba:
* distribute  database_config.json into /var/run/redis/sonic-db/ when installing libswsscommon.deb in postinst script

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -1,0 +1,11 @@
+#!/bin/bash
+# This `DEBIAN/postinst` script is run post-installation
+DST_PATH="/var/run/redis/sonic-db"
+DST_FILE="database_config.json"
+SRC_PATH="/usr/share/swss"
+
+# if there is no $DST_FILE, it is needed to copy one default to $DST_FILE
+if [ ! -e $DST_PATH/$DST_FILE ]; then
+    mkdir -p $DST_PATH
+    cp $SRC_PATH/$DST_FILE $DST_PATH/$DST_FILE
+fi
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -7,7 +7,8 @@ EXTRA_DIST = \
     consumer_table_pops.lua \
     producer_state_table_apply_view.lua \
     table_dump.lua \
-    fdb_flush.lua
+    fdb_flush.lua \
+    database_config.json
 
 swssdir = $(datadir)/swss
 
--- a/common/database_config.json
+++ b/common/database_config.json
@@ -1,0 +1,57 @@
+{
+    "INSTANCES": {
+        "redis":{
+            "hostname" : "127.0.0.1",
+            "port" : 6379,
+            "unix_socket_path" : "/var/run/redis/redis.sock"
+        }
+    },
+    "DATABASES" : {
+        "APPL_DB" : {
+            "id" : 0,
+            "separator": ":",
+            "instance" : "redis"
+        },
+        "ASIC_DB" : {
+            "id" : 1,
+            "separator": ":",
+            "instance" : "redis"
+        },
+        "COUNTERS_DB" : {
+            "id" : 2,
+            "separator": ":",
+            "instance" : "redis"
+        },
+        "LOGLEVEL_DB" : {
+            "id" : 3,
+            "separator": ":",
+            "instance" : "redis"
+        },
+        "CONFIG_DB" : {
+            "id" : 4,
+            "separator": "|",
+            "instance" : "redis"
+        },
+        "PFC_WD_DB" : {
+            "id" : 5,
+            "separator": ":",
+            "instance" : "redis"
+        },
+        "FLEX_COUNTER_DB" : {
+            "id" : 5,
+            "separator": ":",
+            "instance" : "redis"
+        },
+        "STATE_DB" : {
+            "id" : 6,
+            "separator": "|",
+            "instance" : "redis"
+        },
+        "SNMP_OVERLAY_DB" : {
+            "id" : 7,
+            "separator": "|",
+            "instance" : "redis"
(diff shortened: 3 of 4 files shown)


dzhangalibaba:
@qiluo-msft installing default db config file if there is no one in expected location.

qiluo-msft:
> 

Thanks for implement the logic manually. I guess [http://man7.org/linux/man-pages/man5/deb-conffiles.5.html](http://man7.org/linux/man-pages/man5/deb-conffiles.5.html) is just for that, could you help explorer?

---
Refers to: debian/postinst:12 in ef01c58. [](commit_id = ef01c58e8793c4dcee4bef4b6402fc52963654bf, deletion_comment = False)

dzhangalibaba:
> > 
> 
> Thanks for implement the logic manually. I guess http://man7.org/linux/man-pages/man5/deb-conffiles.5.html is just for that, could you help explorer?
> 
> Refers to: debian/postinst:12 in [ef01c58](https://github.com/Azure/sonic-swss-common/commit/ef01c58e8793c4dcee4bef4b6402fc52963654bf). [](commit_id = [ef01c58](https://github.com/Azure/sonic-swss-common/commit/ef01c58e8793c4dcee4bef4b6402fc52963654bf), deletion_comment = False)

Hi Qi, conffiles works, but there is one concern. When we use conffiles, if the config file already exists, installing deb pkg will give a prompt as below. The user needs to decide what to do next.
Did somne search, looks there is no better to skip this prompt in deb pkg installing. If we use this way,  we need to add options when issusing "sudo dpkg -i --force-confold libswsscommon_1.0.0_amd64.deb" with --force-confold option to skip the prompt. Which means we need to modify all the build script which is using "dpkg -i libswsscommon_1.0.0_amd64.deb". This seems not good to me, it introduce more changes and did not get any big benefits. It also force installing to not overwrite existing files, which may be not good in future when there is a need to overwrite. Using manually copy in postinst can avoid this problem and didn't introduce extra changes.


```
Configuration file '/var/run/redis/sonic-db/database_config.json'
 ==> File on system created by you or by a script.
 ==> File also in package provided by package maintainer.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** database_config.json (Y/I/N/O/D/Z) [default=N] ? Y
```

qiluo-msft:
Thanks @dzhangalibaba !

Could you please share your code about "conffiles works" in another PR? I would like to explorer further. To suppress dpkg prompt is the right thing to do in future.

dzhangalibaba:
> Thanks @dzhangalibaba !
> 
> Could you please share your code about "conffiles works" in another PR? I would like to explorer further. To suppress dpkg prompt is the right thing to do in future.

FYI, https://github.com/Azure/sonic-swss-common/pull/323

debian/postinst
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# This `DEBIAN/postinst` script is run post-installation
+DST_PATH="/var/run/redis/sonic-db"
+DST_FILE="database_config.json"
+SRC_PATH="/usr/share/swss"
+
+# if there is no $DST_FILE, it is needed to copy one default to $DST_FILE
```
kcudnik: if file exists we should copy this one one wih suffix ".new" to the same directory to indicate that there is new configuration, btw. i though that there is mechanism already in debian packages to do that, instead manually do that for each file, here is some nice explanation: https://raphaelhertzog.com/2010/09/21/debian-conffile-configuration-file-managed-by-dpkg/

kcudnik:
Just as a question, our future goal is  to support multiple syncd's running across system, controlled by single OA agent, some requirements i posted in my long email:
* each syncd must have global unique id (0..255 etc)
* each syncd must have database assigned (redis db index)
* each syncd must support multiple switches, and each switch must have defined HWINFO for it (SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO (and maybe also SAI_SWITCH_ATTR_SWITCH_PROFILE_ID)  this information is needed to uniquely identify every switch instance globally

my question here is whether we want this information in this file or do we need other file ? this switch/syncd info is something more than db config, but it strictly depends on database configuration which is defined in this file

also different syncd instances can ha same database server used, but asic_db index must be different for each syncd instance

debian/postinst
dzhangalibaba: We can add suffix to save this conf, but we won’t use confines as explained under Qi’s comment, we don’t want prompt.

dzhangalibaba:
> Just as a question, our future goal is to support multiple syncd's running across system, controlled by single OA agent, some requirements i posted in my long email:
> 
> * each syncd must have global unique id (0..255 etc)
> * each syncd must have database assigned (redis db index)
> * each syncd must support multiple switches, and each switch must have defined HWINFO for it (SAI_SWITCH_ATTR_SWITCH_HARDWARE_INFO (and maybe also SAI_SWITCH_ATTR_SWITCH_PROFILE_ID)  this information is needed to uniquely identify every switch instance globally
> 
> **my question here is whether we want this information in this file or do we need other file?** this switch/syncd info is something more than db config, but it strictly depends on database configuration which is defined in this file
> 
> note: different syncd instances can ha same database server used, but asic_db index must be different for each syncd instance

I think we need extend current conf file to support the case you mention. And then add the backend logic to parse conf file to finish the work what you case want

kcudnik:
seems good to me

debian/postinst
kcudnik: ok

PR #323, opened 2019-12-05
[Enhancement] debian/conffiles will give prompt if file existing when installing , need a way to supress prompt

dzhangalibaba:
* if /var/run/redis/sonic-db/database_config.json doesn't exist
    * "sudo dpkg -i libswsscommon_1.0.0_amd64.deb" work as expected.
* if /var/run/redis/sonic-db/database_config.json does exist
    * "sudo dpkg -i --force-confold libswsscommon_1.0.0_amd64.deb" will skip the prompt and use the existing file without overwrite.
    * "sudo dpkg -i libswsscommon_1.0.0_amd64.deb" gives prompt, user need to type
```
    Configuration file '/var/run/redis/sonic-db/database_config.json'
 ==> File on system created by you or by a script.
 ==> File also in package provided by package maintainer.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** database_config.json (Y/I/N/O/D/Z) [default=N] ? Y
```
    
* We need to find a way to suppress prompt if we only using "dpkg -i ".


dzhangalibaba:
@qiluo-msft FYI

debian/libswsscommon.install
```diff
@@ -1,3 +1,4 @@
 usr/lib/*/lib*.so.*
 usr/share/swss/*.lua
+var/run/redis/sonic-db/database_config.json
```
lguohan: var run usually hold running states, not configuration. why not /etc/sonic/redis/sonic-db/database_config.json.

qiluo-msft: Let me try to clarify the reason:
It is the database docker's responsibility to read startup configuration, and generate running configuration. All the downstream components read the "running configuration", and this one is an example.

dzhangalibaba:
@qiluo-msft  ready for review since  https://github.com/Azure/sonic-buildimage/pull/3879 is already merged.

dzhangalibaba:
I suggest we merged this change  until the latest slave-docker is built and used for building, otherwise it may break the build cause of the prompt.

PR #325, opened 2019-12-15
Add resultToString method for Select class

kcudnik:


common/select.cpp
```diff
@@ -167,4 +167,25 @@ bool Select::isQueueEmpty()
...
+            return "OBJECT";
+
+        case swss::Select::ERROR:
+            return "ERROR";
+
+        case swss::Select::TIMEOUT:
+            return "TIMEOUT";
+
+        default:
+            SWSS_LOG_WARN("unknown select result: %d", result);
```
qiluo-msft: >WARN [](start = 21, length = 4)

Treat it as an error?

common/select.h
```diff
@@ -34,6 +34,14 @@ class Select
 
     int select(Selectable **c, int timeout = -1);
     bool isQueueEmpty();
+
+    /**
+     * @brief Result to string.
+     *
+     * Convert select operation result to string representation.
+     */
+    static std::string resultToString(int result);
```
qiluo-msft: >resultToString [](start = 23, length = 14)

What is the use case? Could you add a unit test for this use case?

kcudnik: use case is to print nice string in log if select operation fails for any reason, i will add unittests

common/select.cpp
kcudnik: i set this as warning since it's not a critical issue imo

PR #326, opened 2020-01-13
Add unregister method to NetDispatcher

kcudnik:


common/netdispatcher.h
```diff
@@ -30,13 +33,32 @@ class NetDispatcher {
...
+
 private:
+
     NetDispatcher() = default;
 
+    NetDispatcher(const NetDispatcher&) = delete;
+
+    NetMsg* getCallback(int nlmsg_type);
+
+private:
```
qiluo-msft: Not needed. Is it on purpose?

kcudnik: yes, on purpose, look that its under mutex, so all operations on m_map are synchronized, also get must be protected, so hence t his function, i could do it lika { MUTEX; callbacl = m_map.get(); } but having function is much cleaner

common/netdispatcher.cpp
```diff
@@ -4,34 +4,64 @@
 
 using namespace swss;
 
+#define MUTEX std::lock_guard<std::mutex> _lock(m_mutex);
```
qiluo-msft: >mutex [](start = 35, length = 5)

Use spinlock to improve runtime performance, maybe atomic_flag?

kcudnik: this is not performance critical ?

PR #330, opened 2020-02-19
[schema] update schema with rates and utilization table names

mykolaf:
According to [HLD](https://github.com/Azure/SONiC/blob/ec6d35dd2c28491bfade19cfee990fe612f1e5e9/doc/rates-and-utilization/Rates_and_utilization_HLD.md)

Signed-off-by: Mykola Faryma <mykolaf@mellanox.com>


liat-grozovik:
@abdosi can you please merge so we can update submodule pointers and use it for the next PRs on the list?

PR #332, opened 2020-03-12
Clean netlink includes

kcudnik:


common/netlink.h
```diff
@@ -1,29 +1,30 @@
...
 #include "selectable.h"
 
-namespace swss {
+#include <netlink/netlink.h>
+#include <netlink/route/rtnl.h>
 
-class NetLink : public Selectable {
-public:
-    NetLink(int pri = 0);
-    ~NetLink() override;
```
qiluo-msft: >override [](start = 15, length = 8)

Is there a reason to remove `override`?

kcudnik: its add virtual wchich auto add override

PR #335, opened 2020-03-25
[DBConnector] Add methods to set/get Redis client name

jleveque:
Add `setClientName()` and `getClientName()` methods to DBConnector class to allow for assigning a name to the client in Redis, which is helpful when debugging Redis clients using `redis-cli client list`.

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -224,8 +224,20 @@ string DBConnector::getClientName()
 {
     string command("CLIENT GETNAME");
 
-    RedisReply r(this, command, REDIS_REPLY_STATUS);
-    return r.getReply<std::string>();
+    RedisReply r(this, command);
+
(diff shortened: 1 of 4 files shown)


jleveque:
@kcudnik: Your review was dismissed, as I hadn't fully tested the PR. I am finished making changes and have taken the PR out of the "draft" state. Please feel free to review again when you have a moment.

keboliu:
@jleveque do you have plan to add the "unsubscribe" interface?

jleveque:
@keboliu: I agree that we could look into adding an interface to remove/close DB connections. However, a to subscribe to a table, one must call `addSelectable()`. There already exists an API, `removeSelectable()` which should stop subscriptions on specific tables. You could look into if `removeSelectable()` is sufficient in the interim?

PR #336, opened 2020-04-06
[MultiDB]: Use table name separator from database_config.json, instead of hardcoded in a map

qiluo-msft:

PR #340, opened 2020-04-09
Fix backward compatible for a DBConnector without a dbName

qiluo-msft:
If DBConnector is constructed by a dbId, there is no dbName at all. We need to get a separator from the dbId. And the mapping is get from database_config.json.

The bug was introduced by https://github.com/Azure/sonic-swss-common/pull/336


qiluo-msft:
@dzhangalibaba please help review.

PR #342, opened 2020-05-19
The pclose() function returns -1 if error is detected for executing s…

irene-pan1202:
The pclose () function will return the termination state of the shell, which is the end state of the subroutine.
If an error occurs, pclose() function returns -1, and the cause of the error is stored in errno.

  Signed-off-by: Irene Pan <irene_pan@edge-core.com>

### DIFF
--- a/common/exec.cpp
+++ b/common/exec.cpp
@@ -34,7 +34,7 @@ int exec(const string &cmd, string &stdout)
     }
 
     int ret = pclose(pipe);
-    if (ret != 0)
+    if (ret == -1)
     {
         SWSS_LOG_ERROR("%s: %s", cmd.c_str(), strerror(errno));
     }


common/exec.cpp
```diff
@@ -34,7 +34,7 @@ int exec(const string &cmd, string &stdout)
     }
 
     int ret = pclose(pipe);
-    if (ret != 0)
+    if (ret == -1)
```
qiluo-msft, line 37: >if (ret == -1) [](start = 4, length = 14)

Thanks for the fix!
Could you please add some unit test cases to cover:
1. return positive value
2. return -1 and see the strerror
3. return other negative value

BTW, what is the behavior if pclose's wait successfully but `cmd` returns -1? Can we detect this case and not treat it as ERROR?

PR #344, opened 2020-05-26
Add modifyRedis flag to consumer table object

kcudnik:


common/consumertable.h
```diff
@@ -18,8 +18,19 @@ class ConsumerTable : public ConsumerTableBase, public TableName_KeyValueOpQueue
...
     void pops(std::deque<KeyOpFieldsValuesTuple> &vkco, const std::string &prefix = EMPTY_PREFIX);
 
+    void setModifyRedis(bool modify);
 private:
     std::string m_shaPop;
+
+    /**
+     * @brief Modify Redis database.
+     *
+     * If set to false, will not make changes to database durin POPs operation.
```
qiluo-msft, line 28:
```suggestion
     * If set to false, will not make changes to database during POPs operation.
```

common/consumertable.cpp
```diff
@@ -16,7 +16,8 @@ namespace swss {
 
 ConsumerTable::ConsumerTable(DBConnector *db, const string &tableName, int popBatchSize, int pri)
     : ConsumerTableBase(db, tableName, popBatchSize, pri)
-    , TableName_KeyValueOpQueues(tableName)
+    , TableName_KeyValueOpQueues(tableName),
```
qiluo-msft, line 19: >, [](start = 43, length = 1)

Could you help move comma to next line so they follow above lines aligning on punctuations?

Issue #322, opened 2019-12-04
make database config optional


ishidawataru:
[Thie PR](https://github.com/Azure/sonic-swss-common/pull/319) made the existence of the database config mandatory.

To preserve previous behavior, I think it's better to make the config optional.
Is it possible to change the implementation to use DEFAULT_UNIXSOCKET if the database config doesn't exist?

dzhangalibaba:
we don't want to use DEFAULT_UNIXSOCKET anymore, later we will always have database config file which is built in images by default. Everything related to DB will be read from the config file in future. database config doesn't exist is a invalid case in furture.

ishidawataru:
Thanks for your prompt response.
In my use case, I'm not using the official SONiC container image. ( I admit it is a rare use case )
It would be nice if this library can be used easily with fewer dependencies.
I think it will be beneficial for testing and debugging as well.

qiluo-msft:
@ishidawataru Could you give a use case? The new feature should be backward-compatible by design.

ishidawataru:
@qiluo-msft Actually, I was suggesting to maintain the backward compatibility which was broken by https://github.com/Azure/sonic-swss-common/pull/319.

Before the PR, I could use this library without `database_config.json`. But the PR made the existence of the file mandatory.

qiluo-msft:
@ishidawataru Yes, we are considering this. I am asking for a true use case so we could improve the implementation for backward compatibility. Do you have some sample code?

ishidawataru:
I'm not using the official SONiC container image. I'm developing a minimalistic SONiC package that could be installed on any Linux distribution.
I'm not sure I can say this is a true use case though.

PoC code here. https://github.com/ishidawataru/usonic

qiluo-msft:
Offline discussed with @ishidawataru, this issue is a feature requirement. Use case:
1. If database_config.json is provided, use it
2. If no database_config.json at all, fallback to single instance.

Currently swss-common library's downstream modules are using dbName string to access a Redis database, we need to implement a mapping from dbName to dbID (integer) without a configuration file.

dzhangalibaba:
> Offline discussed with @ishidawataru, this issue is a feature requirement. Use case:
> 
> 1. If database_config.json is provided, use it
> 2. If no database_config.json at all, fallback to single Redis instance.
> 
[...]

PR #345, opened 2020-05-28
[DBConnector]: Add database config env support

ishidawataru:
fixes https://github.com/Azure/sonic-swss-common/issues/322


vasant17:
Hi Ishida, when can I expect this to bemerged?

PR #347, opened 2020-06-02
swss-common: Changes to support SONiC Gearbox Manager

sydlogan:
* add class GearboxUtils, contains APIs useful for gearbox support across various submodules.
* add additional REDIS tables to support second (PHY) syncd.

  Signed-off-by: syd.logan@broadcom.com


common/gearboxutils.cpp
```diff
@@ -0,0 +1,459 @@
...
+    SWSS_LOG_DEBUG("Parsed key:%s, Return:(%s)(%s)(%s)", key_str.c_str(), str1.c_str(), str2.c_str(), str3.c_str());
+
+    return make_tuple(str1, str2, str3);
+}
+
+bool GearboxUtils::platformHasGearbox()
+{
+    bool ret = false;
+
+    if (access("/usr/share/sonic/hwsku/gearbox_config.json", F_OK) != -1)
```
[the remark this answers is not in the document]
dan-arsenault: Hi Kamil;
I understand what you're saying w.r.t. hard-coding paths. The gearbox_config.json file (along with it's associated phy configuration file) are however based on the specific hardware sku (much like port_config.ini). This is why we chose the well known path; "/usr/share/sonic/hwsku" which is available to all the dockers. Please let me know your thoughts on this.

common/table.cpp
```diff
@@ -19,14 +19,18 @@ const std::string TableBase::TABLE_NAME_SEPARATOR_COLON = ":";
...
-   { STATE_DB,        TABLE_NAME_SEPARATOR_VBAR  }
+   { APPL_DB,           TABLE_NAME_SEPARATOR_COLON },
+   { ASIC_DB,           TABLE_NAME_SEPARATOR_COLON },
+   { COUNTERS_DB,       TABLE_NAME_SEPARATOR_COLON },
+   { LOGLEVEL_DB,       TABLE_NAME_SEPARATOR_COLON },
+   { CONFIG_DB,         TABLE_NAME_SEPARATOR_VBAR  },
+   { PFC_WD_DB,         TABLE_NAME_SEPARATOR_COLON },
+   { FLEX_COUNTER_DB,   TABLE_NAME_SEPARATOR_COLON },
+   { STATE_DB,          TABLE_NAME_SEPARATOR_VBAR  },
+   { ASIC_DB2,          TABLE_NAME_SEPARATOR_VBAR  },
```
lguohan, line 30: what is ASIC_DB2? it is unclear what it is.

common/Makefile.am
```diff
@@ -53,6 +53,7 @@ libswsscommon_la_SOURCES = \
     notificationconsumer.cpp  \
     notificationproducer.cpp  \
     linkcache.cpp             \
+    gearboxutils.cpp          \
```
lguohan, line 56: why not put this code in swss? why put them under swss-common?

common/table.cpp
dan-arsenault: Hi Guohan;

The new ASIC_DB2 (as well as counters, flex, and state) database designations are being used by Gearbox and the new PHY SYNCD DOCKER.
These definitions are also specified in the specific HWSKU context_config.json files, such as in the following excerpt;

{
    "guid" : 1,
    "name" : "phy1",
    "dbAsic" : "ASIC_DB2",
    "dbCounters" : "COUNTERS_DB2",
    "dbFlex": "FLEX_COUNTER_DB2",
    "dbState" : "STATE_DB2",
    "switches": [
        {
            "index" : 1,
            "hwinfo" : ""
        }
    ]
}

common/Makefile.am
sydlogan: The intent for placing the GearboxUtils class here was to make its API (e.g., for determining if gearbox is supported or not) available to other code outside of swss, if needed.

sydlogan: @lguohan , there is no library as of yet in sonic-swss. Options if we move this from swss-common to swss:

1. Create a new lib dir in sonic-swss and either make each using subdirectory makefile add ../lib/gearboxutils.cpp to its sources with each recompiling that file (also change -I so they can find the header). 

2. Or, create a Makefile in the new sonic-swss/lib and generate a lib and have each Makefike link against this.  Just be a fex functions in the .a, and won't be shared outside of swss.

I don't want to duplicate this code wherever needed within swss, so some kind of new lib dir seems to be needed.

Not only is swss-common functioning like this lib option for multiple places in swss, it also encourages coders not to re-invent the wheel and instead use swss-common code to check for gearbox support. So, still would think swss-common is a good place for gearboxutils.

sydlogan: Okay, will remove it.

sydlogan:
@lguohan  removed gearboxutils, explained why the naming convention was chosen. Re-requesting review. Thanks!

common/table.cpp
lguohan: I think all the phy db can be put into the same db since there are not much query/config activities. I think PHY_DB is a better name and give user more information on the nature of this db.

common/schema.h
```diff
@@ -14,10 +14,15 @@ namespace swss {
 #define FLEX_COUNTER_DB 5
 #define STATE_DB        6
 #define SNMP_OVERLAY_DB 7
+#define ASIC_DB2        8
```
lguohan, line 17: can you rename to PHY_ASIC_DB?

common/schema.h
```diff
@@ -14,10 +14,15 @@ namespace swss {
...
+#define COUNTERS_DB2    9
+#define FLEX_COUNTER_DB2    10
+#define STATE_DB2      11
```
lguohan, line 20: I would like to see

#define PHY_ASIC_DB             8
#define PHY_COUNTERS_DB    9
#define PHY_FLEX_COUNTER_DB    10

lguohan, line 20: i am not sure we need new STATE_DB for the phy, since STATE_DB maps to APP_DB, and we are not adding a new APP_DB for the gearbox.

common/table.cpp
kcudnik: just to clarify, a added "2" for each of second db names since i didn't have any better name, we could add PHY or GB (as gearbox) prefixx or something more explanatory then just "2"

sydlogan:
@lguohan thank you for the review and approval

PR #348, opened 2020-06-04
[schema] Fine Grained ECMP schema changes

anish-n:
Schema changes per HLD:
https://github.com/Azure/SONiC/blob/master/doc/ecmp/fine_grained_next_hop_hld.md


lguohan:
can you resolve the conflict

PR #350, opened 2020-06-10
[warm reboot] Warm Reboot Support for EVPN VXLAN

srj102:
1. Added new enums to WarmStartState
2. Added a new function to get the warm start state.

Refer to HLD: https://github.com/Azure/SONiC/pull/437

### DIFF
--- a/common/warm_restart.cpp
+++ b/common/warm_restart.cpp
@@ -192,8 +192,10 @@ void WarmStart::getWarmStartState(const std::string &app_name, WarmStartState &s
 
     state = RECONCILED;
 
-    if(!isWarmStart())
-      return;
+    if (!isWarmStart())
+    {
+        return;
+    }
 
     warmStart.m_stateWarmRestartTable->hget(app_name, "state", statestr);
 
@@ -202,9 +204,9 @@ void WarmStart::getWarmStartState(const std::string &app_name, WarmStartState &s
      */
     state = WSUNKNOWN;
 
-    for(auto it = warmStartStateNameMap.begin(); it != warmStartStateNameMap.end(); it++)
+    for (auto it = warmStartStateNameMap.begin(); it != warmStartStateNameMap.end(); it++)
     {
-        if(it->second == statestr)
+        if (it->second == statestr)
         {
             state = it->first;
             break;
--- a/common/warm_restart.cpp
+++ b/common/warm_restart.cpp
@@ -10,7 +10,10 @@ const WarmStart::WarmStartStateNameMap WarmStart::warmStartStateNameMap =
 {
     {INITIALIZED,   "initialized"},
     {RESTORED,      "restored"},
-    {RECONCILED,    "reconciled"}
+    {REPLAYED,      "replayed"},
+    {RECONCILED,    "reconciled"},
+    {WSDISABLED,    "disabled"},
+    {WSUNKNOWN,     "unknown"}
 };
 
 const WarmStart::DataCheckStateNameMap WarmStart::dataCheckStateNameMap =
@@ -183,7 +186,6 @@ bool WarmStart::isSystemWarmRebootEnabled(void)
 
 void WarmStart::getWarmStartState(const std::string &app_name, WarmStartState &state)
 {
-    //WarmStartStateNameMap::iterator it;
     std::string statestr;
 
     auto& warmStart = getInstance();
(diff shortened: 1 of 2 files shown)


prsunny:
Request review from @dzhangalibaba , @zhenggen-xu

common/warm_restart.cpp
```diff
@@ -181,6 +181,41 @@ bool WarmStart::isSystemWarmRebootEnabled(void)
     return warmStart.m_systemWarmRebootEnabled;
 }
 
+void WarmStart::getWarmStartState(const std::string &app_name, WarmStartState &state)
+{
+    //WarmStartStateNameMap::iterator it;
```
qiluo-msft, line 186: >//WarmStartStateNameMap::iterator it; [](start = 4, length = 37)

Remove commented code

common/warm_restart.cpp
```diff
@@ -181,6 +181,41 @@ bool WarmStart::isSystemWarmRebootEnabled(void)
     return warmStart.m_systemWarmRebootEnabled;
 }
 
+void WarmStart::getWarmStartState(const std::string &app_name, WarmStartState &state)
```
qiluo-msft, line 184: >WarmStartState [](start = 63, length = 14)

Why not just return this state?

common/warm_restart.cpp
```diff
@@ -181,6 +181,41 @@ bool WarmStart::isSystemWarmRebootEnabled(void)
...
+      return;
+
+    warmStart.m_stateWarmRestartTable->hget(app_name, "state", statestr);
+
+    /* If warm-start is enabled, state cannot be assumed as Reconciled
+     * It should be set to unknown
+     */
+    state = WSUNKNOWN;
+
+    for(std::map<WarmStart::WarmStartState, std::string>::const_iterator it = warmStartStateNameMap.begin(); it != warmStartStateNameMap.end(); it++)
```
qiluo-msft, line 203: >std::map<WarmStart::WarmStartState, std::string>::const_iterator [](start = 8, length = 64)

auto?

common/warm_restart.cpp
```diff
@@ -181,6 +181,41 @@ bool WarmStart::isSystemWarmRebootEnabled(void)
...
+    {
+      if(it->second == statestr)
+      {
+         state = it->first;
+         break;
+      }
```
qiluo-msft, line 209: >   [](start = 4, length = 2)

Use 4x spaces as indentation

common/warm_restart.cpp
```diff
@@ -181,6 +181,41 @@ bool WarmStart::isSystemWarmRebootEnabled(void)
...
+
+    if(!isWarmStart())
+      return;
+
+    warmStart.m_stateWarmRestartTable->hget(app_name, "state", statestr);
+
+    /* If warm-start is enabled, state cannot be assumed as Reconciled
+     * It should be set to unknown
+     */
+    state = WSUNKNOWN;
```
qiluo-msft, line 201: >state = WSUNKNOWN; [](start = 4, length = 18)

Is it expected state in runtime? If not, just throw an exception.

common/warm_restart.cpp
nkelapur: Don't see any issue with pass by reference. Will consider that while updating other files where this api is used.

common/warm_restart.cpp
nkelapur: Will remove

common/warm_restart.cpp
nkelapur: Yes, this is expected. When Module A check for the status of Module B, module B may not have reached/initialised to any state yet. Hence module A needs to know that Module B is not ready yet an act accordingly.

common/warm_restart.cpp
nkelapur: will update

common/warm_restart.cpp
nkelapur: will change

common/warm_restart.cpp
```diff
@@ -181,6 +184,39 @@ bool WarmStart::isSystemWarmRebootEnabled(void)
...
 
+void WarmStart::getWarmStartState(const std::string &app_name, WarmStartState &state)
+{
+    std::string statestr;
+
+    auto& warmStart = getInstance();
+
+    state = RECONCILED;
+
+    if(!isWarmStart())
```
prsunny, line 195: Please have a space after `if`, also with open/close braces. Please do it for the rest of the `if `and `for `below

nkelapur: Done.

prsunny (approved):

qiluo-msft (approved):

PR #353, opened 2020-06-18
Support to register for RAW Netlink Msg handling to support EVPN Type5 route as described in the PR Azure/SONiC#437

kishorekunal01:
EVPN Type5 msg handling in FPMSYNCD need RAW netlink MSG to extract RMAC information, Adding infra similar to normal MSG handling.

What I did
1) Created new registration for RAW netlink msg handling in NetDispatcher. (PR depend on this Pull request https://github.com/Azure/sonic-swss/pull/1326)

Why I did it
Part of overall EVPN-VXLAN Implementation

How I verified it
Verified on VS

Details if related
HLD Location : https://github.com/Azure/SONiC/pull/437/commits

Signed-off-by: Kishore Kunal kishore.kunal@broadcom.com

### DIFF
--- a/common/netdispatcher.cpp
+++ b/common/netdispatcher.cpp
@@ -23,6 +23,14 @@ void NetDispatcher::registerMessageHandler(int nlmsg_type, NetMsg *callback)
     m_handlers[nlmsg_type] = callback;
 }
 
+void NetDispatcher::registerRawMessageHandler(int nlmsg_type, NetMsg *callback)
+{
+    if (m_rawhandlers.find(nlmsg_type) != m_rawhandlers.end())
+        throw "Trying to registered on already registerd netlink message";
+
+    m_rawhandlers[nlmsg_type] = callback;
+}
+
 void NetDispatcher::unregisterMessageHandler(int nlmsg_type)
 {
     MUTEX;
@@ -51,6 +59,20 @@ NetMsg* NetDispatcher::getCallback(int nlmsg_type)
         return nullptr;
 
     return callback->second;
+}
+
+void NetDispatcher::onNetlinkMessageRaw(struct nl_msg *msg)
+{
+    struct nlmsghdr *nlmsghdr = nlmsg_hdr(msg);
+    NetMsg *reg_callback = NULL;
+    auto callback = m_rawhandlers.find(nlmsghdr->nlmsg_type);
+
+    /* Drop not registered messages */
+    if (callback == m_rawhandlers.end())
+        return;
+
+    reg_callback = (NetMsg *)callback->second;
+    reg_callback->onMsgRaw(nlmsghdr);
 }
 
 void NetDispatcher::onNetlinkMessage(struct nl_msg *msg)
--- a/common/netmsg.h
+++ b/common/netmsg.h
@@ -9,5 +9,10 @@ namespace swss
         public:
             /* Called by NetDispatcher when netmsg matches filters */
             virtual void onMsg(int nlmsg_type, struct nl_object *obj) = 0;
+
+            /* Called by NetDispatcher when raw msg is send for matches filters */
+            virtual void onMsgRaw(struct nlmsghdr *)
+            {
+            }
     };
 }
--- a/common/producerstatetable.h
+++ b/common/producerstatetable.h
@@ -24,11 +24,6 @@ public:
                      const std::string &op = DEL_COMMAND,
                      const std::string &prefix = EMPTY_PREFIX);
 
-	virtual void hdel(const std::string &key,
-					 const std::string &field,
-					 const std::string &op = "",
-					 const std::string &prefix = EMPTY_PREFIX);
-
     void flush();
 
     int64_t count();
--- a/common/producerstatetable.h
+++ b/common/producerstatetable.h
@@ -24,6 +24,11 @@ public:
                      const std::string &op = DEL_COMMAND,
                      const std::string &prefix = EMPTY_PREFIX);
 
+	virtual void hdel(const std::string &key,
+					 const std::string &field,
+					 const std::string &op = "",
+					 const std::string &prefix = EMPTY_PREFIX);
+
     void flush();
 
     int64_t count();
--- a/common/netdispatcher.h
+++ b/common/netdispatcher.h
@@ -23,6 +23,13 @@ namespace swss
              */
             void registerMessageHandler(int nlmsg_type, NetMsg *callback);
 
+            /*
+             * Register callback class according to message-type.
+             *
+             * Throw exception if
+             */
+            void registerRawMessageHandler(int nlmsg_type, NetMsg *callback);
+
             /** Called by NetLink or FpmLink classes as indication of new packet arrival. */
             void onNetlinkMessage(struct nl_msg *msg);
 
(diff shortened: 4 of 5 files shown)


common/producerstatetable.h
```diff
@@ -24,6 +24,11 @@ class ProducerStateTable : public TableBase, public TableName_KeySet
                      const std::string &op = DEL_COMMAND,
                      const std::string &prefix = EMPTY_PREFIX);
 
+	virtual void hdel(const std::string &key,
```
qiluo-msft, line 27: >hdel [](start = 14, length = 4)

Why you need this function? The implementation is significant different from set/del.

common/netmsg.h
```diff
@@ -9,5 +9,10 @@ namespace swss
         public:
             /* Called by NetDispatcher when netmsg matches filters */
             virtual void onMsg(int nlmsg_type, struct nl_object *obj) = 0;
+
+            /* Called by NetDispatcher when raw msg is send for matches filters */
+            virtual void onMsgRaw(struct nlmsghdr *)
+            {
+            }
```
qiluo-msft, line 16: Do you want a pure virtual function?

common/netdispatcher.cpp
```diff
@@ -53,6 +61,20 @@ NetMsg* NetDispatcher::getCallback(int nlmsg_type)
     return callback->second;
 }
 
+void NetDispatcher::onNetlinkMessageRaw(struct nl_msg *msg)
```
qiluo-msft, line 64: >onNetlinkMessageRaw [](start = 20, length = 19)

Is it really necessary?

How about using onNetlinkMessageJ(), and inside callback, call `struct nlmsghdr *nlmsghdr = nlmsg_hdr(msg);`

common/producerstatetable.h
preetham-singh: Here requirement is to modify certain selected field and consumer should get notification once all fields are updated(publish should happen after multiple fields updated).
In specific scenario, where route nhop is updated from overlay to underlay, we need to ensure only vni & rmac fields are deleted instead of deleting the route completely and re-adding same with new underlay nhop(since doing delete followed by add will cause data path disruption). Setting default values in set in above scenario would require changes in consumer as it deviates from current underlay route APP_DB ROUTE_TABLE schema(which has only ifname & nexthop).
Note: this is borrowed from Table::hdel().

qiluo-msft: You could not achieve "consumer should get notification" if:
1. you `hdel` in ProducerStateTable 
2. you are using ConsumerStateTable as consumer

---
In reply to: [444392253](https://github.com/Azure/sonic-swss-common/pull/353#discussion_r444392253) [](ancestors = 444392253)

common/netmsg.h
kishorekunal01: Since all the derived class might not be interested in this onRawMsg handling function, Hence I have not define this as pure virtual function. If I make this as pure virtual function I need to modify all the derived class to define this function.

common/netdispatcher.cpp
kishorekunal01: This is to keep the callback for RawMsg Handling separate then Normal Msg handling. With this approach a derived class can register for RawMsg as well as Normal msg for same netlink msg.

As you can see in the pull request Azure/sonic-swss#1326.
FPMSYNCD is interest in RTM_NEWROUTE/RTM_DELROUTE for both
i) Normal Msg(All the exiting implementation for ip route add and VNET route add).
ii) RawMsg to handle EVPN route from FRR. As RMAC, VNI, VLAN is encoded in the netlink msg.

qiluo-msft: I see the general requirement to support multiple callback on one nlmsg_type. Instead of creating new `m_rawhandlers` just for 2nd handler, you may change `m_handlers` to `multimap` so it will support more than 2 handlers.

After that, my previous comment still applicable.

---
In reply to: [443930827](https://github.com/Azure/sonic-swss-common/pull/353#discussion_r443930827) [](ancestors = 443930827)

common/producerstatetable.h
qiluo-msft: Please be aware this comment will be a blocking issue.

---
In reply to: [444414066](https://github.com/Azure/sonic-swss-common/pull/353#discussion_r444414066) [](ancestors = 444414066,444392253)

kishorekunal01: I will let Preetham reply for this.

common/netdispatcher.cpp
kishorekunal01: Agreed, it can be done to register with multimap, I feel this is more readable. 
But I am still not sure how can we call different callback from onNetlinkMessage when both the message type is registered.

qiluo-msft: I cannot see how onNetlinkMessageRaw is used in your PR. Could you explain the use case of this function?

---
In reply to: [446460758](https://github.com/Azure/sonic-swss-common/pull/353#discussion_r446460758) [](ancestors = 446460758,443930827)

kishorekunal01: > I cannot see how onNetlinkMessageRaw is used in your PR. Could you explain the use case of this function?
> 
> In reply to: [446460758](https://github.com/Azure/sonic-swss-common/pull/353#discussion_r446460758) [](ancestors = 446460758,443930827)

As you can see in the pull request Azure/sonic-swss#1326.
FPMSYNCD is interest in RTM_NEWROUTE/RTM_DELROUTE for both message type (raw message as well as rtnl message)
i) RTNL Msg(All the exiting implementation for ip route add and VNET route add).
ii) RawMsg to handle EVPN route from FRR. As RMAC, VNI, VLAN is encoded in the netlink msg.

qiluo-msft: Thanks for the link!

> But I am still not sure how can we call different callback from onNetlinkMessage when both the message type is registered.

This should be easy achievable in fpmsyncd. You can register both handlers, and you could check `isRawProcessing` function and only effectively run one of the handler.

---
In reply to: [446480443](https://github.com/Azure/sonic-swss-common/pull/353#discussion_r446480443) [](ancestors = 446480443,446460758,443930827)

kishorekunal01: > Thanks for the link!
> 
> > But I am still not sure how can we call different callback from onNetlinkMessage when both the message type is registered.
> 
> This should be easy achievable in fpmsyncd. You can register both handlers, and you could check `isRawProcessing` function and only effectively run one of the handler.
> 
> In reply to: [446480443](https://github.com/Azure/sonic-swss-common/pull/353#discussion_r446480443) [](ancestors = 446480443,446460758,443930827)

As you are suggesting we are also going the same in fpmsyncd, if the msg need to be handled as raw, we call the corresponding raw msg handle.

Snippet from fpmlink.cpp file
            if (isRaw)
            {
                NetDispatcher::getInstance().onNetlinkMessageRaw(msg);
            }
            else
            {
                NetDispatcher::getInstance().onNetlinkMessage(msg);
            }

common/producerstatetable.h
kishorekunal01: We will be deleting hdel API, as you suggested

common/netdispatcher.cpp
kishorekunal01: Sure, updating the description section

kperumalbfn (approved):

qiluo-msft (changes requested):
Block until we get final resolution on this discussion https://github.com/Azure/sonic-swss/pull/1326#discussion_r504966317

kishorekunal01:
Closing this pull request, as per review comment from @qiluo-msft, Just handled the msg raw processing in the fpmsyncd itself

PR #355, opened 2020-06-21
[python3]: build python3-swsscommon for python3 support

lguohan:


.gitignore
```diff
@@ -8,6 +8,10 @@
 ####################
 pyext/swsscommon.py
 pyext/swsscommon_wrap.cpp
```
qiluo-msft, line 10: Are they needed?

m4/python3.m4
```diff
@@ -0,0 +1,192 @@
+## ------------------------                                 -*- Autoconf -*-
```
qiluo-msft, line 1: Is this autogenerated? Should we ignore it in source control?

lguohan: no, this is manually added.

lguohan:
retest this please

.gitignore
lguohan: just ignore the previous artifacts.

PR #360, opened 2020-06-29
Update .gitignore

pavel-shirshov:
Updated .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,8 @@ pyext/py3/swsscommon_wrap.cpp
 
 # Packaging Files #
 ###################
+debian/.debhelper/
+debian/debhelper-build-stamp
 debian/files
 debian/libswsscommon-dev.debhelper.log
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -43,10 +49,18 @@ stamp-h1
...
+m4/lt~obsolete.m4
 
 # Executables #
 ###############
 common/swssloglevel
 tests/tests
 
+# IDE #
+#######
+.idea
```
qiluo-msft, line 65: Is this only in your dev env? If yes, could you ignore it in your home directory `.gitignore_global`?

pavel-shirshov: Sure I can Qi. But this line here doesn't interfere with anything else in the repo. Can we keep it here?

qiluo-msft: `.gitignore_global` in your home dir will be much better and generalized for all your repos.

PR #361, opened 2020-07-01
[Dynamic buffer calc] Support dynamic buffer calculation

stephenxs:
1. Database schema updates.
2. Move the json file parsing from swss to swss-common.


neethajohn (changes requested):

common/json.cpp
```diff
@@ -36,4 +39,63 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
...
+                        if ((*cur_obj_it).is_number())
+                            value_str = to_string((*cur_obj_it).get<int>());
+                        else if ((*cur_obj_it).is_string())
+                            value_str = (*cur_obj_it).get<string>();
+                        kfvFieldsValues(cur_db_item).push_back(FieldValueTuple(field_str, value_str));
+                    }
+                }
+                else
+                {
+                    kfvOp(cur_db_item) = cur_obj.get<string>();
```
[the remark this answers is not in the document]
stephenxs: This is a redundant sentence and has been removed.

common/json.cpp
```diff
@@ -36,4 +39,59 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
     }
 }
 
+bool JSon::loadJsonFromFile(ifstream &fs, vector<KeyOpFieldsValuesTuple> &db_items)
```
liat-grozovik, line 42: can you please add some documentation explaining the motivation of this and why it is needed?
explanation of the return code and the output can also be nice and what is the file type expected.

common/json.h
```diff
@@ -10,9 +11,12 @@ namespace swss {
 
 class JSon
 {
+private:
+    static const int el_count = 2;
```
liat-grozovik, line 15: what is 2 stands for as the default?

common/schema.h
```diff
@@ -87,6 +87,13 @@ namespace swss {
...
 #define APP_MACSEC_INGRESS_SA_TABLE_NAME    "MACSEC_INGRESS_SA_TABLE"
 
+#define APP_BUFFER_POOL_TABLE_NAME      "BUFFER_POOL_TABLE"
+#define APP_BUFFER_PROFILE_TABLE_NAME   "BUFFER_PROFILE_TABLE"
+#define APP_BUFFER_PG_TABLE_NAME        "BUFFER_PG_TABLE"
+#define APP_BUFFER_QUEUE_TABLE_NAME     "BUFFER_QUEUE_TABLE"
+#define APP_BUFFER_PORT_INGRESS_PROFILE_LIST_NAME   "BUFFER_PORT_INGRESS_PROFILE_LIST_TABLE"
+#define APP_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME    "BUFFER_PORT_EGRESS_PROFILE_LIST_TABLE"
+
 /***** TO BE REMOVED *****/
```
liat-grozovik, line 97: although this is not news, what should be removed?

liat-grozovik (changes requested):
Is there any other doc file which schema should be updated as well?

common/json.cpp
stephenxs: Done. add some description in the header.

common/json.h
stephenxs: the number of elements in each object: 2 stands for the data and the op.
comment added for it as well

common/schema.h
stephenxs: I can't say for sure, but I think probably it should be the following QoS tables.
The qosorch consumes CONFIG_DB table and the tables in APPL_DB aren't used.

liat-grozovik (dismissed):

liat-grozovik:
@neetha please review and provide your feedback. this PR is needed to be the first to be merged so other related PRs can be tested properly in Jenkins PR CI

common/json.cpp
```diff
@@ -36,4 +39,59 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
...
+                    for (nlohmann::json::iterator cur_obj_it = cur_obj.begin(); cur_obj_it != cur_obj.end(); cur_obj_it++)
+                    {
+                        string field_str = cur_obj_it.key();
+                        string value_str;
+                        if ((*cur_obj_it).is_number())
+                            value_str = to_string((*cur_obj_it).get<int>());
+                        else if ((*cur_obj_it).is_string())
+                            value_str = (*cur_obj_it).get<string>();
+                        kfvFieldsValues(cur_db_item).push_back(FieldValueTuple(field_str, value_str));
+                    }
```
neethajohn, line 84: Where is the 'OP' field getting parsed?

stephenxs: Added that logic. If `op` isn't `SET`, ignore the item and log an error.

neethajohn (approved):

liat-grozovik (approved):

liat-grozovik (approved):

liat-grozovik:
@neethajohn your approval was lost with the last push. can you please approve and merge?

neethajohn (approved):

liat-grozovik:
@qiluo-msft can you please sign off?

common/json.cpp
```diff
@@ -36,4 +39,70 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
     }
 }
 
+bool JSon::loadJsonFromFile(ifstream &fs, vector<KeyOpFieldsValuesTuple> &db_items)
```
qiluo-msft, line 42: >fs [](start = 38, length = 2)

What happens if `fs` is already `close()` or in bad status before enter this function?

common/json.cpp
```diff
@@ -36,4 +39,70 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
...
+{
+    nlohmann::json json_array;
+    fs >> json_array;
```
qiluo-msft, line 45: >fs >> json_array; [](start = 4, length = 17)

What happens if there is IO error, or invalid json format? Will any case throw exception?

common/json.cpp
```diff
@@ -36,4 +39,70 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
...
+        if (arr_item.is_object())
+        {
+            if (el_count != arr_item.size())
+            {
+                SWSS_LOG_ERROR("Child elements must have both key and op entry. %s",
+                               arr_item.dump().c_str());
+                return false;
+            }
+
+            db_items.push_back(KeyOpFieldsValuesTuple());
```
qiluo-msft, line 66: >push_back [](start = 21, length = 9)

`emplace_back` will be better, one less temp object.

common/json.cpp
```diff
@@ -36,4 +39,70 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
...
+            auto &cur_db_item = db_items[db_items.size() - 1];
```
qiluo-msft, line 67: >db_items[db_items.size() - 1] [](start = 32, length = 29)

`db_items.back()`

common/json.cpp
```diff
@@ -36,4 +39,70 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
...
+
+            for (nlohmann::json::iterator child_it = arr_item.begin(); child_it != arr_item.end(); child_it++) {
```
qiluo-msft, line 69: >nlohmann::json::iterator [](start = 17, length = 24)

auto

qiluo-msft, line 69: >{ [](start = 111, length = 1)

Move `}` to next line

common/json.cpp
```diff
@@ -36,4 +39,70 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
...
+                            value_str = (*cur_obj_it).get<string>();
+                        kfvFieldsValues(cur_db_item).push_back(FieldValueTuple(field_str, value_str));
+                    }
+                }
+                else
+                {
+                    auto op = cur_obj.get<string>();
+                    if (op != "SET")
+                    {
+                        SWSS_LOG_ERROR("Child elements'op field must be SET, but got %s, ignored", op.c_str());
```
qiluo-msft, line 91: >op [](start = 55, length = 2)

Add one blank before?

qiluo-msft, line 91: >must be [](start = 64, length = 7)

Why must be? Is there any document?

qiluo-msft (changes requested):
As comments

common/json.cpp
stephenxs: There can be an exception which will be catched by the caller.

common/json.cpp
stephenxs: fixed

common/json.cpp
stephenxs: fixed

common/json.cpp
stephenxs: fixed

stephenxs: fixed

common/json.cpp
stephenxs: fixed

stephenxs: The are two operations defined: `SET` and `DEL`. The `DEL` doesn't make sense here because it's about to insert the items into the database.

common/json.cpp
```diff
@@ -36,4 +39,73 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
...
+        if (arr_item.is_object())
+        {
+            if (el_count != arr_item.size())
+            {
+                SWSS_LOG_ERROR("Child elements must have both key and op entry. %s",
+                               arr_item.dump().c_str());
+                return false;
+            }
+
+            db_items.emplace_back(KeyOpFieldsValuesTuple());
```
qiluo-msft, line 66: >KeyOpFieldsValuesTuple [](start = 34, length = 22)

You can actually `emplace_back()`. It's like combining ctor and push into one statement.

common/json.cpp
```diff
@@ -36,4 +39,73 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
...
+                    kfvKey(cur_db_item) = cur_obj_key;
+                    for (nlohmann::json::iterator cur_obj_it = cur_obj.begin(); cur_obj_it != cur_obj.end(); cur_obj_it++)
+                    {
+                        string field_str = cur_obj_it.key();
+                        string value_str;
+                        if ((*cur_obj_it).is_number())
+                            value_str = to_string((*cur_obj_it).get<int>());
+                        else if ((*cur_obj_it).is_string())
+                            value_str = (*cur_obj_it).get<string>();
+                        kfvFieldsValues(cur_db_item).emplace_back(FieldValueTuple(field_str, value_str));
```
qiluo-msft, line 85: >emplace_back(FieldValueTuple(field_str, value_str)) [](start = 53, length = 51)

`emplace_back(field_str, value_str)`

common/json.cpp
qiluo-msft: Sorry, I mean `elements' op`, ie. before `op`

---
In reply to: [532396054](https://github.com/Azure/sonic-swss-common/pull/361#discussion_r532396054) [](ancestors = 532396054)

qiluo-msft: It depends on the use case. In general, it still makes sense to support all the `op` options.

---
In reply to: [532396572](https://github.com/Azure/sonic-swss-common/pull/361#discussion_r532396572) [](ancestors = 532396572)

common/json.cpp
qiluo-msft: Then better to return false instead of throwing?

---
In reply to: [532393445](https://github.com/Azure/sonic-swss-common/pull/361#discussion_r532393445) [](ancestors = 532393445)

common/json.cpp
stephenxs: sounds reasonable. 
so we just need to check whether it's `SET` or `DEL` and log an error only if neither matched. and move the test to the caller.

common/json.cpp
stephenxs: OK.
In that case, can we declare the method as "noexcept"?

common/json.cpp
```diff
@@ -36,4 +39,82 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
...
 
+bool JSon::loadJsonFromFile(ifstream &fs, vector<KeyOpFieldsValuesTuple> &db_items)
+{
+    nlohmann::json json_array;
+
+    try
+    {
+        fs >> json_array;
+    }
+    catch (...)
```
qiluo-msft, line 50: >catch (...) [](start = 4, length = 11)

Do not catch general exception except in most outmost level. Catch by as specific type as possible. I guess there are two types: fstream exception and json deserialization exception.
Also need to log exception messages.

stephenxs: It can throw the following exceptions:
- bad_alloc in constructors in case of insufficient memory
- domain_error, out_of_range and invalid_argument in case of the syntax error
To cover them all, I tend to catch the `std::exception`. even though it indeed is a big set but it's the most concrete exception which covers all the possibilities.
@qiluo-msft how do you think?

qiluo-msft: `exception` is still too general. So my comment still applies. Could you break it down to more specific exception type or multiple exception types?

---
In reply to: [533090214](https://github.com/Azure/sonic-swss-common/pull/361#discussion_r533090214) [](ancestors = 533090214)

stephenxs: ok.
for `domain_error`, `out_of_range` and `invalid_argument`, they are all derived from `logic_error` so we can catch it.
for `bad_alloc`, it needs to be handled separately.

stephenxs: @qiluo-msft done.

common/json.cpp
```diff
@@ -36,4 +39,87 @@ void JSon::readJson(const string &jsonstr, vector<FieldValueTuple> &fv)
...
+        fs >> json_array;
+    }
+    catch (const std::logic_error &e)
+    {
+        SWSS_LOG_ERROR("Unable to parse json from the input stream: %s", e.what());
+        return false;
+    }
+    catch (const std::bad_alloc &e)
+    {
+        SWSS_LOG_ERROR("Unable to parse json from the imput stream: %s", e.what());
```
qiluo-msft, line 57: >imput [](start = 54, length = 5)

`input`

stephenxs: fixed

qiluo-msft (approved):

liat-grozovik:
@qiluo-msft can you please merge?

liat-grozovik (approved):

PR #364, opened 2020-07-09
DBConnector classes to understand the namespace.

judyjoseph:
Why:
The dbconnector classes here need to understand the namespace, to connect to redis databases running in database containers in different linux network namespaces in case of multi-asic platforms.

What:
The class data structures are modified to handle the new dimension of namespace. New api's are added to parse the database_global.json file and to retrieve the list of namespaces + parse the respective database_config.json file. The namespaces will be ""(default namespace for the linux host) and "asic0", "aisic1" .... as present in the /var/run/redis/sonic-db/database_global.json file.

How verified:
Verification was done on a multi-asic platform with pmon scripts which uses swsscommon dbconnector classes.
**Note** : testcases to be updated, in this or in a follow up PR.

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -140,6 +140,8 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
 
     // Set it as the global config file is already parsed and init done.
     m_global_init = true;
+    // Make regular init also done
+    m_init = true;
 }
 
 void SonicDBConfig::initialize(const string &file, const string &nameSpace)
@@ -175,28 +177,28 @@ void SonicDBConfig::initialize(const string &file, const string &nameSpace)
 string SonicDBConfig::getDbInst(const string &dbName, const string &nameSpace)
 {
     if (!m_init)
-        initialize(nameSpace);
+        initialize(DEFAULT_SONIC_DB_CONFIG_FILE, nameSpace);
     return m_db_info[nameSpace].at(dbName).instName;
 }
 
 int SonicDBConfig::getDbId(const string &dbName, const string &nameSpace)
 {
     if (!m_init)
-        initialize(nameSpace);
+        initialize(DEFAULT_SONIC_DB_CONFIG_FILE, nameSpace);
     return m_db_info[nameSpace].at(dbName).dbId;
 }
 
 string SonicDBConfig::getSeparator(const string &dbName, const string &nameSpace)
 {
     if (!m_init)
-        initialize(nameSpace);
+        initialize(DEFAULT_SONIC_DB_CONFIG_FILE, nameSpace);
     return m_db_info[nameSpace].at(dbName).separator;
 }
 
 string SonicDBConfig::getSeparator(int dbId, const string &nameSpace)
 {
     if (!m_init)
-        initialize(nameSpace);
+        initialize(DEFAULT_SONIC_DB_CONFIG_FILE, nameSpace);
     return m_db_separator[nameSpace].at(dbId);
 }
 
@@ -222,21 +224,21 @@ string SonicDBConfig::getSeparator(const DBConnector* db)
 string SonicDBConfig::getDbSock(const string &dbName, const string &nameSpace)
 {
     if (!m_init)
-        initialize(nameSpace);
+        initialize(DEFAULT_SONIC_DB_CONFIG_FILE, nameSpace);
     return m_inst_info[nameSpace].at(getDbInst(dbName)).unix_socket_path;
 }
 
 string SonicDBConfig::getDbHostname(const string &dbName, const string &nameSpace)
 {
     if (!m_init)
-        initialize(nameSpace);
+        initialize(DEFAULT_SONIC_DB_CONFIG_FILE, nameSpace);
     return m_inst_info[nameSpace].at(getDbInst(dbName)).hostname;
 }
 
 int SonicDBConfig::getDbPort(const string &dbName, const string &nameSpace)
 {
     if (!m_init)
-        initialize(nameSpace);
+        initialize(DEFAULT_SONIC_DB_CONFIG_FILE, nameSpace);
     return m_inst_info[nameSpace].at(getDbInst(dbName)).port;
 }
 
@@ -369,16 +371,7 @@ string DBConnector::getNamespace() const
 DBConnector *DBConnector::newConnector(unsigned int timeout) const
 {
     DBConnector *ret;
-    if (getContext()->connection_type == REDIS_CONN_TCP)
-        ret = new DBConnector(getDbId(),
-                               getContext()->tcp.host,
-                               getContext()->tcp.port,
-                               timeout);
-    else
-        ret = new DBConnector(getDbId(),
-                               getContext()->unix_sock.path,
-                               timeout);
-    ret->m_dbName = m_dbName;
+    ret = new DBConnector(getDbName(), timeout, (getContext()->connection_type == REDIS_CONN_TCP), getNamespace());
     return ret;
 }
 
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -69,7 +69,7 @@ void SonicDBConfig::parseDatabaseConfig(const string &file,
 
 void SonicDBConfig::initializeGlobalConfig(const string &file)
 {
-    std::string local_file, ns_name;
+    std::string local_file, dir_name, ns_name;
     std::unordered_map<std::string, SonicDBInfo> db_entry;
     std::unordered_map<std::string, RedisInstInfo> inst_entry;
     std::unordered_map<int, std::string> separator_entry;
@@ -82,9 +82,14 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
         return;
     }
 
+
     ifstream i(file);
     if (i.good())
     {
+        // Get the directory name from the file path given as input.
+        std::string::size_type pos = file.rfind("/");
+        dir_name = file.substr(0,pos+1);
+
         try
         {
             json j;
@@ -92,7 +97,7 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
 
             for (auto& element : j["INCLUDES"])
             {
-                local_file.append(DEFAULT_SONIC_DB_CONFIG_DIR);
+                local_file.append(dir_name);
                 local_file.append(element["include"]);
 
                 if(element["namespace"].empty())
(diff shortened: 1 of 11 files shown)


common/dbconnector.h
```diff
@@ -59,13 +78,14 @@ class DBConnector
      */
     DBConnector(int dbId, const std::string &hostname, int port, unsigned int timeout);
     DBConnector(int dbId, const std::string &unixPath, unsigned int timeout);
-    DBConnector(const std::string &dbName, unsigned int timeout, bool isTcpConn = false);
+    DBConnector(const std::string &dbName, unsigned int timeout, bool isTcpConn = false, const std::string &nameSpace = EMPTY_NAMESPACE);
```
qiluo-msft, line 81: >EMPTY_NAMESPACE [](start = 120, length = 15)

To make default value more efficient, check [https://stackoverflow.com/a/32851509/2514803](https://stackoverflow.com/a/32851509/2514803)

common/dbconnector.h
```diff
@@ -84,6 +104,7 @@ class DBConnector
     redisContext *m_conn;
     int m_dbId;
     std::string m_dbName;
+    std::string m_nameSpace;
```
qiluo-msft, line 107: >m_nameSpace [](start = 16, length = 11)

m_namespace or m_nsName

common/dbconnector.h
```diff
@@ -7,11 +7,20 @@
 #include <utility>
 
 #include <hiredis/hiredis.h>
+#define EMPTY_NAMESPACE ""
 
 namespace swss {
 
 class DBConnector;
 
+class SonicInstInfo
```
qiluo-msft, line 16: >SonicInstInfo [](start = 6, length = 13)

RedisInstInfo

common/dbconnector.h
```diff
@@ -7,11 +7,20 @@
...
+{
+public:
+    std::string socket;
```
qiluo-msft, line 19: >socket [](start = 16, length = 6)

Please follow the names in config file "unixSocketPath"

common/dbconnector.h
```diff
@@ -23,26 +32,36 @@ class SonicDBInfo
...
-    static std::string getDbInst(const std::string &dbName);
-    static int getDbId(const std::string &dbName);
-    static std::string getSeparator(const std::string &dbName);
-    static std::string getSeparator(int dbId);
+    static void parse_config(const std::string &file, 
+                             std::unordered_map<std::string, SonicInstInfo> &inst_entry,
+                             std::unordered_map<std::string, SonicDBInfo> &db_entry, 
+                             std::unordered_map<int, std::string> &separator_entry);
+    static void initialize(const std::string &file = DEFAULT_SONIC_DB_CONFIG_FILE, const std::string &nameSpace = EMPTY_NAMESPACE);
+    static void initialize_global_config(const std::string &file = DEFAULT_SONIC_DB_GLOBAL_CONFIG_FILE);
```
qiluo-msft, line 40: >initialize_global_config [](start = 16, length = 24)

initializeGlobalConfig

common/dbconnector.h
```diff
@@ -23,26 +32,36 @@ class SonicDBInfo
 class SonicDBConfig
 {
 public:
-    static void initialize(const std::string &file = DEFAULT_SONIC_DB_CONFIG_FILE);
-    static std::string getDbInst(const std::string &dbName);
-    static int getDbId(const std::string &dbName);
-    static std::string getSeparator(const std::string &dbName);
-    static std::string getSeparator(int dbId);
+    static void parse_config(const std::string &file,
```
qiluo-msft, line 35: >parse_config [](start = 16, length = 12)

Why public function?

qiluo-msft, line 35: >parse_config [](start = 16, length = 12)

parseDatabaseConfig

qiluo-msft:
@dzhangalibaba Please help review

common/dbconnector.h
```diff
@@ -23,26 +32,36 @@ class SonicDBInfo
...
+    static void parse_config(const std::string &file, 
+                             std::unordered_map<std::string, SonicInstInfo> &inst_entry,
+                             std::unordered_map<std::string, SonicDBInfo> &db_entry, 
+                             std::unordered_map<int, std::string> &separator_entry);
+    static void initialize(const std::string &file = DEFAULT_SONIC_DB_CONFIG_FILE, const std::string &nameSpace = EMPTY_NAMESPACE);
+    static void initialize_global_config(const std::string &file = DEFAULT_SONIC_DB_GLOBAL_CONFIG_FILE);
+    static std::string getDbInst(const std::string &dbName, const std::string &nameSpace = EMPTY_NAMESPACE);
+    static int getDbId(const std::string &dbName, const std::string &nameSpace = EMPTY_NAMESPACE);
+    static std::string getSeparator(const std::string &dbName, const std::string &nameSpace = EMPTY_NAMESPACE);
+    static std::string getSeparator(int dbId, const std::string &nameSpace = EMPTY_NAMESPACE);
```
qiluo-msft, line 44: >getSeparator [](start = 23, length = 12)

Please cover new functionality in unit test.

common/dbconnector.h
qiluo-msft: camelCase will be better

---
In reply to: [453213014](https://github.com/Azure/sonic-swss-common/pull/364#discussion_r453213014) [](ancestors = 453213014)

common/dbconnector.cpp
```diff
@@ -73,32 +67,132 @@ void SonicDBConfig::initialize(const string &file)
...
+            i >> j;
+
+            for (auto& element : j["INCLUDES"])
+            {
+                local_file.append(DEFAULT_SONIC_DB_CONFIG_DIR);
+                local_file.append(element["include"]);
+
+                if(element["namespace"].empty())
+                {
+                    ns_name = EMPTY_NAMESPACE;
```
dzhangalibaba, line 100: for the empty case, do we need set m_init somewhere ? otherwise, when using get func, it parse the database_config.json again ?

common/dbconnector.h
judyjoseph: Thanks Qi, have  updated the #define EMPTY_NAMESPACE std::string().

common/dbconnector.cpp
judyjoseph: Here in this case when element["namespace"].empty(), I could add a check to see if the m_init flag is already set -- IF so don't do anything, as it is already loaded  explicitly from database_config.json file earlier. Please let me know if this was what you were referring to ?

judyjoseph:
> when to use m_global_init in get func ?

With the current approach using the flag m_init itself to cover the cases where initialization is already done once -- either with EMPTY namespace OR with a valid namespace name. m_global_init is  getting used indirectly when a namespace argument is provided to initialize API.

common/dbconnector.cpp
dzhangalibaba: I saw you added the m_init = true at the last. Just to confirm with you,  database_global_config.json always has the namesapce EMPTY case in file, right ?

common/dbconnector.cpp
```diff
@@ -73,32 +67,139 @@ void SonicDBConfig::initialize(const string &file)
...
+        {
+            json j;
+            i >> j;
+
+            for (auto& element : j["INCLUDES"])
+            {
+                local_file.append(dir_name);
+                local_file.append(element["include"]);
+
+                if(element["namespace"].empty())
```
dzhangalibaba, line 103: I saw your changes, that fix one case, when initializeGlobalConfig() invoked first, initialize() won't parse empty case again. But when initialize() invoked first with empty namespace and then initializeGlobalConfig() is invoked, there is no check on m_init before handing empty case, double parsing again.

judyjoseph: Yes, Dong -- I am currently adding that as well , thanks !

judyjoseph:
Rebased due to merge conflict .

abdosi:
@judyjoseph Please submodule pointer of master for sonic-swss-common

judyjoseph:
> @judyjoseph Please update submodule pointer of master for sonic-swss-common

@abdosi , Will need to revert this commit temporarily - as this is causing a build issue in sonic-swss. This is the PR in swss to track that https://github.com/Azure/sonic-swss/pull/1362. There are some test failures in swss which is preventing me from committing the fix. So please don't advance the submodule pointer in 201911 branch as well.

common/dbconnector.cpp
```diff
@@ -73,32 +67,152 @@ void SonicDBConfig::initialize(const string &file)
     }
 }
 
-string SonicDBConfig::getDbInst(const string &dbName)
+void SonicDBConfig::initializeGlobalConfig(const string &file)
```
abdosi, line 70: @judyjoseph Behavior is different when calling this api w.r.t sonic-py-swssdk for single-asic platforms.
In python version this is NOOP on single asic platforms but here it throws run-time exception of file not present.

I feel we should have same behavior.

For now we need to add check in caller to make sure to call this api for multi-asic platforms only.

judyjoseph: Yes it is different here. 

No need to explicitly call SonicDBConfig::initializeGlobalConfig API. If the user db_connects with a namespace  parameter it loads the Global config file automatically (https://github.com/judyjoseph/sonic-swss-common/blob/4be4c3970f8f70523ef85bc7e56b19f9b2d2b1d3/common/dbconnector.cpp#L184)

The next time he calls it will be a NOOP as this variable m_global_init is set already. I felt this will be easier to user ? let me know.

If user needs to call this API  SonicDBConfig::initializeGlobalConfig() explicitly he need to make sure the file exists.

abdosi: @judyjoseph 
a) why it has to be different ? It has be consistent . Any particular reason why it need to be different ? Application should be doing same irrespective of library being used.

b)Also it does not work if we don't call SonicDBConfig::initializeGlobalConfig. 
DBConnector Constructor with particular asic namespace gives seg fault
as SonicDBConfig::getDbId will fail since m_init will be set when call for host/global namespace and when next time called
it will not go for else condition (since m_init is already set) 

https://github.com/judyjoseph/sonic-swss-common/blob/4be4c3970f8f70523ef85bc7e56b19f9b2d2b1d3/common/dbconnector.cpp#L336

judyjoseph: Ok got the use-case, thanks ! here in this case the same application we invoke with the global namespace followed by the namespace specific ones.  Will update to make it similar to how we have it in sonic-py-swssdk of explicitly calling SonicDBConfig::initializeGlobalConfig().

abdosi: @judyjoseph is it possible to add test-case to cover above scenario ?

PR #365, opened 2020-07-17
Fix sai api loglevel

shi-su:
- Why I did it
This PR contains the code changes to fix the SAI API log level issue Azure/sonic-swss-common#366

- Why this issue happens
In the initialization of syncd, the main function first calls linkToDbNative function to initialize the log level for syncd. Then the log levels for SAI APIs are initialized using linkToDb. Since the settingThread of logger is set up by linkToDbNative function and the ConsumerStateTables are set up only in the initialization of the thread, the SAI API log levels in only added to m_settingChangeObservers, whereas the corresponding ConsumerStateTables are not created. Therefore, the SAI API log levels are not successfully initialized.

- How I did it
Instead of creating ConsumerStateTables at the initialization of settingThread, I create ConsumerStateTables in the execution loop of the thread. Such a change fixes the issue by allowing to create the corresponding ConsumerStateTables when there are new items added to m_settingChangeObservers.

- How to verify it
Verify all SAI API log levels are correctly initialized and can be modified in vs tests. A unit test is also added to simulate the scenario and verify the log levels can be properly handled.

### DIFF
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -170,20 +170,19 @@ Logger::Priority Logger::getMinPrio()
 {
     Select select;
     DBConnector db("LOGLEVEL_DB", 0);
-    std::vector<std::shared_ptr<ConsumerStateTable>> selectables;
-    std::set<std::string> selectableKeys;
+    std::map<std::string, std::shared_ptr<ConsumerStateTable>> selectables;
 
-    while(true)
+    while (true)
     {
-        if(selectableKeys.size() < m_settingChangeObservers.size())
+        if(selectables.size() < m_settingChangeObservers.size())
         {
             for (const auto& i : m_settingChangeObservers)
             {
-                if(selectableKeys.find(i.first) == selectableKeys.end())
+                std::string dbName = i.first;
+                if(selectables.find(dbName) == selectables.end())
                 {
-                    std::shared_ptr<ConsumerStateTable> table = std::make_shared<ConsumerStateTable>(&db, i.first);
-                    selectableKeys.insert(i.first);
-                    selectables.push_back(table);
+                    auto table = std::make_shared<ConsumerStateTable>(&db, dbName);
+                    selectables.insert(std::make_pair(dbName, table));
                     select.addSelectable(table.get());
                 }
             }
@@ -191,7 +190,8 @@ Logger::Priority Logger::getMinPrio()
 
         Selectable *selectable = nullptr;
 
-        int ret = select.select(&selectable, 1000);
+        /* TODO Resolve latency caused by timeout at initialization. */
+        int ret = select.select(&selectable, 1000); // Timeout if there is no data in 1000 ms.
 
         if (ret == Select::ERROR)
         {
@@ -201,6 +201,7 @@ Logger::Priority Logger::getMinPrio()
 
         if (ret == Select::TIMEOUT)
         {
+            SWSS_LOG_INFO("%s select timeout", __PRETTY_FUNCTION__);
             continue;
         }
 
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -174,12 +174,12 @@ Logger::Priority Logger::getMinPrio()
 
     while (true)
     {
-        if(selectables.size() < m_settingChangeObservers.size())
+        if (selectables.size() < m_settingChangeObservers.size())
         {
             for (const auto& i : m_settingChangeObservers)
             {
                 std::string dbName = i.first;
-                if(selectables.find(dbName) == selectables.end())
+                if (selectables.find(dbName) == selectables.end())
                 {
                     auto table = std::make_shared<ConsumerStateTable>(&db, dbName);
                     selectables.insert(std::make_pair(dbName, table));
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -178,11 +178,11 @@ Logger::Priority Logger::getMinPrio()
         {
             for (const auto& i : m_settingChangeObservers)
             {
-                std::string dbName = i.first;
+                const std::string &dbName = i.first;
                 if (selectables.find(dbName) == selectables.end())
                 {
                     auto table = std::make_shared<ConsumerStateTable>(&db, dbName);
-                    selectables.insert(std::make_pair(dbName, table));
+                    selectables.emplace(dbName, table);
                     select.addSelectable(table.get());
                 }
             }
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -170,24 +170,37 @@ Logger::Priority Logger::getMinPrio()
 {
     Select select;
     DBConnector db("LOGLEVEL_DB", 0);
-    std::vector<std::shared_ptr<ConsumerStateTable>> selectables(m_settingChangeObservers.size());
-
-    for (const auto& i : m_settingChangeObservers)
-    {
-        std::shared_ptr<ConsumerStateTable> table = std::make_shared<ConsumerStateTable>(&db, i.first);
-        selectables.push_back(table);
-        select.addSelectable(table.get());
-    }
+    std::vector<std::shared_ptr<ConsumerStateTable>> selectables;
+    std::set<std::string> selectableKeys;
 
     while(true)
     {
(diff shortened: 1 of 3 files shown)


qiluo-msft:
@gechiang Could you help test this PR?

common/logger.cpp
```diff
@@ -170,27 +170,40 @@ Logger::Priority Logger::getMinPrio()
...
+    std::vector<std::shared_ptr<ConsumerStateTable>> selectables;
+    std::set<std::string> selectableKeys;
 
     while(true)
     {
+        if(selectableKeys.size() < m_settingChangeObservers.size())
+        {
+            for (const auto& i : m_settingChangeObservers)
+            {
+                if(selectableKeys.find(i.first) == selectableKeys.end())
```
qiluo-msft, line 182: >i.first [](start = 39, length = 7)

`i.first` is used multiple times. Could you give it a name to improve readability?

common/logger.cpp
```diff
@@ -170,27 +170,40 @@ Logger::Priority Logger::getMinPrio()
...
+                {
+                    std::shared_ptr<ConsumerStateTable> table = std::make_shared<ConsumerStateTable>(&db, i.first);
```
qiluo-msft, line 184: >std::shared_ptr<ConsumerStateTable> [](start = 20, length = 35)

auto

lguohan:
@shi-su, can you provide more detail description of the pr? what is the bug, how the bug is fixed.

common/logger.cpp
qiluo-msft, line 182: >selectableKeys [](start = 19, length = 14)

You can `insert` and check return value. No need to find beforehand.

common/logger.cpp
```diff
@@ -170,27 +170,40 @@ Logger::Priority Logger::getMinPrio()
...
+                    selectableKeys.insert(i.first);
+                    selectables.push_back(table);
```
qiluo-msft, line 186: >push_back [](start = 32, length = 9)

emplace_back?

common/logger.cpp
```diff
@@ -170,27 +170,40 @@ Logger::Priority Logger::getMinPrio()
...
-    for (const auto& i : m_settingChangeObservers)
-    {
-        std::shared_ptr<ConsumerStateTable> table = std::make_shared<ConsumerStateTable>(&db, i.first);
-        selectables.push_back(table);
-        select.addSelectable(table.get());
-    }
+    std::vector<std::shared_ptr<ConsumerStateTable>> selectables;
+    std::set<std::string> selectableKeys;
 
     while(true)
```
qiluo-msft, line 176: >while(true) [](start = 4, length = 11)

Add one blank after `while`

common/logger.cpp
```diff
@@ -170,27 +170,40 @@ Logger::Priority Logger::getMinPrio()
 {
     Select select;
     DBConnector db("LOGLEVEL_DB", 0);
-    std::vector<std::shared_ptr<ConsumerStateTable>> selectables(m_settingChangeObservers.size());
```
qiluo-msft, line 173: >(m_settingChangeObservers.size()); [](start = 64, length = 34)

Thanks to fix this bug.
Recommend to `reserve` if you know it will probably occupy.

gechiang:
> @gechiang Could you help test this PR?

sure I will help test this out

common/logger.cpp
```diff
@@ -170,27 +170,40 @@ Logger::Priority Logger::getMinPrio()
...
-
-    for (const auto& i : m_settingChangeObservers)
-    {
-        std::shared_ptr<ConsumerStateTable> table = std::make_shared<ConsumerStateTable>(&db, i.first);
-        selectables.push_back(table);
-        select.addSelectable(table.get());
-    }
+    std::vector<std::shared_ptr<ConsumerStateTable>> selectables;
+    std::set<std::string> selectableKeys;
```
qiluo-msft, line 174: Seems `selectables` is only used to manage memory. You may use a map instead of vector + set.

common/logger.cpp
```diff
@@ -170,27 +170,40 @@ Logger::Priority Logger::getMinPrio()
...
+                    selectables.push_back(table);
+                    select.addSelectable(table.get());
+                }
+            }
+        }
+
         Selectable *selectable = nullptr;
 
-        int ret = select.select(&selectable);
+        int ret = select.select(&selectable, 1000);
```
qiluo-msft, line 194: >1000 [](start = 45, length = 4)

Add comment about constant unit.

qiluo-msft, line 194: >select [](start = 25, length = 6)

Add a unit test case to simulate the fixed bug.

common/logger.cpp
```diff
@@ -170,27 +170,40 @@ Logger::Priority Logger::getMinPrio()
...
 
         if (ret == Select::ERROR)
         {
             SWSS_LOG_NOTICE("%s select error %s", __PRETTY_FUNCTION__, strerror(errno));
             continue;
         }
 
+        if (ret == Select::TIMEOUT)
+        {
+            continue;
```
kcudnik, line 204: if this is not intended to happen, please add log line

gechiang:
> > @gechiang Could you help test this PR?
> 
> sure I will help test this out
> Is there a buildimage that I can find to load this to my DUT?

I have tested this change and find it resolved the issue I raised.  I was able to set the levels of all the SAI component via the "-a" option as well as individual SAI component via the "-c" option.  Fix looks good. 
Thanks!

common/logger.cpp
shi-su: Done.

common/logger.cpp
shi-su: Done.

common/logger.cpp
shi-su: Since the vector selectables and the set selectableKeys in now combined as a map, inserting into the map requires a pointer to a consumerStateTable. In such a case, I think it might be better to find before creating the consumerStateTable.

common/logger.cpp
shi-su: Blank added.

common/logger.cpp
shi-su: I updated selectables as a map.

common/logger.cpp
shi-su: Fixed as prescribed.

common/logger.cpp
shi-su: Done.

shi-su: A unit test is added.

common/logger.cpp
shi-su: It now becomes inserting into a map.

common/logger.cpp
shi-su: Since a timeout is added to select selectables, the timeout is actually intended. I added an info level log to the timeout.

common/logger.cpp
```diff
@@ -170,27 +170,41 @@ Logger::Priority Logger::getMinPrio()
...
     {
-        std::shared_ptr<ConsumerStateTable> table = std::make_shared<ConsumerStateTable>(&db, i.first);
-        selectables.push_back(table);
-        select.addSelectable(table.get());
-    }
+        if (selectables.size() < m_settingChangeObservers.size())
+        {
+            for (const auto& i : m_settingChangeObservers)
+            {
+                std::string dbName = i.first;
```
qiluo-msft, line 181: >std::string [](start = 16, length = 11)

`const std::string&` or `const auto&`

common/logger.cpp
qiluo-msft: [https://en.cppreference.com/w/cpp/container/map/emplace](https://en.cppreference.com/w/cpp/container/map/emplace)

---
In reply to: [456727155](https://github.com/Azure/sonic-swss-common/pull/365#discussion_r456727155) [](ancestors = 456727155)

shi-su: Changed to emplace.

common/logger.cpp
shi-su: Done.

lguohan:
@shi-su, can you updat the submodule pointer on sonic-buildimage?

Issue #327, opened 2020-01-16
swss::exec return value and message has nothing to do with actual command return value


stepanblyschak:
Consider the following log:
```
Jan 16 09:57:09.359886 CYS05-0101-1105-14T1 ERR swss#intfmgrd: :- exec: /sbin/ip address add 10.106.116.191/31 dev PortChannel30: Success
Jan 16 09:57:09.359886 CYS05-0101-1105-14T1 DEBUG swss#intfmgrd: :- exec: /sbin/ip address add 10.106.116.191/31 dev PortChannel30 : 
Jan 16 09:57:09.359886 CYS05-0101-1105-14T1 ERR swss#intfmgrd: :- setIntfIp: Command '/sbin/ip address add 10.106.116.191/31 dev PortChannel30' failed with rc 512
```

exec prints ': Success' because that's what returns ```strerror(errno)```, however ```errno``` is set per process and this is not the same ```errno``` from the failed one.
Also, return code (512 in this case) is ambiguous as this is not in range of return codes from shell commands

PR #368, opened 2020-07-23
Fix swss::exec return value

qiluo-msft:
Fix https://github.com/Azure/sonic-swss-common/issues/327

### DIFF
--- a/common/exec.cpp
+++ b/common/exec.cpp
@@ -9,10 +9,9 @@ using namespace std;
 
 namespace swss {
 
-const int buffsz = 128;
-
 int exec(const string &cmd, string &stdout)
 {
+    const int buffsz = 128;
     array<char, buffsz> buffer;
     FILE* pipe = popen(cmd.c_str(), "r");
 
@@ -39,7 +38,7 @@ int exec(const string &cmd, string &stdout)
     int status = pclose(pipe);
     if (status == -1)
     {
-        ret = -1;
+        ret = EXEC_ERROR_PCLOSE;
         SWSS_LOG_ERROR("%s: %s", cmd.c_str(), strerror(errno));
     }
     else if (WIFEXITED(status))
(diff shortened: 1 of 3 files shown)


qiluo-msft:
Retest this please

qiluo-msft:
@stepanblyschak Could you also help review?

common/exec.h
```diff
@@ -4,6 +4,15 @@
 
 namespace swss {
 
+enum EXEC_RC
```
pavel-shirshov, line 7: is it possible to use enum class here?

qiluo-msft: Technically enum class is working but not convenient. I just need named constant integer, so enum is good enough.

---
In reply to: [460169126](https://github.com/Azure/sonic-swss-common/pull/368#discussion_r460169126) [](ancestors = 460169126)

pavel-shirshov: enum class is c++ feature which helps gives us more type checking in compile time. Why is it non-convenient?

qiluo-msft: There are no implicit conversions from the values of a scoped enumerator to integral types.

---
In reply to: [460172825](https://github.com/Azure/sonic-swss-common/pull/368#discussion_r460172825) [](ancestors = 460172825,460169126)

common/exec.cpp
```diff
@@ -33,12 +33,43 @@ int exec(const string &cmd, string &stdout)
         }
     }
 
-    int ret = pclose(pipe);
-    if (ret != 0)
+    int ret;
+    string msg;
+    int status = pclose(pipe);
+    if (status == -1)
```
tahmed-dev, line 39: nit: the chain of if else if.. lends itself well to a switch-case statement.

tahmed-dev: nevermind. this will assume knowledge about the ranges each class of error code occupy.

common/exec.h
pavel-shirshov: Sorry. I didn't get. The feature of enum class I meant is you can compare C enum value with any integer value, and C++ enum class must be compared with this enum class instance only, any integer will not work. sonic-swss-common is C++, so should we use C++ features, not C?

qiluo-msft: I agree `enum class` has improvement on `enum`, however, the type safety does not help/hurt my use case. I don't even use comparison.

I use them for readable `return NNN;` statements. Using `enum class` requires explicit type conversion, which hurts the readability and gains not much.

---
In reply to: [460183560](https://github.com/Azure/sonic-swss-common/pull/368#discussion_r460183560) [](ancestors = 460183560,460172825,460169126)

PR #370, opened 2020-07-30
[warmboot] add disable warmstart timer value

dzhangalibaba:
* adding definition for disable warmstart teamsyncd timer value
 

dzhangalibaba:
@pavel-shirshov

common/warm_restart.h
```diff
@@ -6,6 +6,7 @@
 #include "table.h"
 
 #define MAXIMUM_WARMRESTART_TIMER_VALUE 9999
```
qiluo-msft, line 8: According to your schema https://github.com/Azure/sonic-swss/pull/1364/files
The max should be 3600

dzhangalibaba: if you check the CLI codes in sonic-utilities/config/maim.py, you will find the max value for neighsyncd_timer is 9999 and for teamsyncd it is 1-3600, hence I correct the doc which should be 3600 based on the CLI codes. For the history , why some are 9999 and others are 3600 , I am not sure, maybe the owner knows. There is no chance for teamsyncd to have values more than 3600.

```
@warm_restart.command('neighsyncd_timer')
@click.argument('seconds', metavar='<seconds>', required=True, type=int)
@click.pass_context
def warm_restart_neighsyncd_timer(ctx, seconds):
    db = ctx.obj['db']
    if seconds not in range(1,9999):
        ctx.fail("neighsyncd warm restart timer must be in range 1-9999")
    db.mod_entry('WARM_RESTART', 'swss', {'neighsyncd_timer': seconds})

@warm_restart.command('bgp_timer')
@click.argument('seconds', metavar='<seconds>', required=True, type=int)
@click.pass_context
def warm_restart_bgp_timer(ctx, seconds):
    db = ctx.obj['db']
    if seconds not in range(1,3600):
        ctx.fail("bgp warm restart timer must be in range 1-3600")
    db.mod_entry('WARM_RESTART', 'bgp', {'bgp_timer': seconds})

@warm_restart.command('teamsyncd_timer')
@click.argument('seconds', metavar='<seconds>', required=True, type=int)
@click.pass_context
def warm_restart_teamsyncd_timer(ctx, seconds):
    db = ctx.obj['db']
    if seconds not in range(1,3600):
        ctx.fail("teamsyncd warm restart timer must be in range 1-3600")
    db.mod_entry('WARM_RESTART', 'teamd', {'teamsyncd_timer': seconds})
```

dzhangalibaba:
> As comment

@qiluo-msft any more concern ? The other two PR depends on this one.

PR #371, opened 2020-08-03
Reintroducing PR #364 - DBconnector to support namespaces.

judyjoseph:
Adding PR #364 back. 

It was reverted earlier as there was a build failure with **sonic-swss/tests/mock_tests/mock_dbconnector.cpp** failing to build due to DBConnnector constructor being redefined there (Issue mentioned in https://github.com/Azure/sonic-swss/issues/1363). 
I have addressed it now by keeping the old constructor and delegating that to the new constructor which has "namespace" as additional argument.

This PR, I have included the changes to further align the dbConnector class here in sonic-swss-common with one present in sonic-py-swssdk
       1. Ignore ( don't throw exception ) if the database_global.json file is not present.
       2. validate the namespace before using it.

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -16,17 +16,11 @@ using namespace std;
 
 namespace swss {
 
-void SonicDBConfig::initialize(const string &file)
+void SonicDBConfig::parseDatabaseConfig(const string &file,
+                    std::unordered_map<std::string, RedisInstInfo> &inst_entry,
+                    std::unordered_map<std::string, SonicDBInfo> &db_entry,
+                    std::unordered_map<int, std::string> &separator_entry)
 {
-
-    SWSS_LOG_ENTER();
-
-    if (m_init)
-    {
-        SWSS_LOG_ERROR("SonicDBConfig already initialized");
-        throw runtime_error("SonicDBConfig already initialized");
-    }
-
     ifstream i(file);
     if (i.good())
     {
@@ -40,7 +34,7 @@ void SonicDBConfig::initialize(const string &file)
                string socket = it.value().at("unix_socket_path");
                string hostname = it.value().at("hostname");
                int port = it.value().at("port");
-               m_inst_info[instName] = {socket, {hostname, port}};
+               inst_entry[instName] = {socket, hostname, port};
             }
 
             for (auto it = j["DATABASES"].begin(); it!= j["DATABASES"].end(); it++)
@@ -49,12 +43,12 @@ void SonicDBConfig::initialize(const string &file)
                string instName = it.value().at("instance");
                int dbId = it.value().at("id");
                string separator = it.value().at("separator");
-               m_db_info[dbName] = {instName, dbId, separator};
+               db_entry[dbName] = {instName, dbId, separator};
 
-               m_db_separator.emplace(dbId, separator);
+               separator_entry.emplace(dbId, separator);
             }
-            m_init = true;
         }
+
         catch (domain_error& e)
         {
             SWSS_LOG_ERROR("key doesn't exist in json object, NULL value has no iterator >> %s\n", e.what());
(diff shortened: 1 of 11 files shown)


judyjoseph:
retest this please

common/dbconnector.cpp
```diff
@@ -73,32 +67,185 @@ void SonicDBConfig::initialize(const string &file)
...
+        {
+            SWSS_LOG_ERROR("Sonic database config file syntax error >> %s\n", e.what());
+            throw runtime_error("Sonic database config file syntax error >> " + string(e.what()));
+        }
+
+        // Set it as the global config file is already parsed and init done.
+        m_global_init = true;
+
+        // Make regular init also done
+        m_init = true;
```
qiluo-msft, line 151: Please make the logic same as python version

common/dbconnector.cpp
```diff
@@ -73,32 +67,185 @@ void SonicDBConfig::initialize(const string &file)
...
+    }
+    else
+    {
+        SWSS_LOG_ERROR("Sonic database config global file doesn't exist at %s\n", file.c_str());
+    }
+}
+
+void SonicDBConfig::initialize(const string &file, const string &nameSpace)
```
qiluo-msft, line 159: >const string &nameSpace [](start = 51, length = 23)

Why this function has extra parameter than the python version `load_sonic_db_config` ?

common/dbconnector.cpp
judyjoseph: Sure, @qiluo-msft I was thinking about this yesterday, I finally moved it here as I thought this is logical to set the init flags to TRUE only if the database_global.json file exists and we have parsed it correct ? 
Let me know what you think , I could change it back in python version if you too feel the same.

common/dbconnector.cpp
judyjoseph: Will update - thanks !

common/dbconnector.cpp
qiluo-msft: If there is a bug in python version, we need to fix it first.
If not, make the C++ version behaves the same as python version.

---
In reply to: [464586440](https://github.com/Azure/sonic-swss-common/pull/371#discussion_r464586440) [](ancestors = 464586440)

judyjoseph: I have updated so that it matches the python version. I agree that setting the flag m_global_init is ok here -- because this is an API which is invoked by user and he should provide the valid database_global.json file as input.

abdosi:
Is this comment on previous PR is addressed:

a) why it has to be different ? It has be consistent . Any particular reason why it need to be different ? Application should be doing same irrespective of library being used.

b)Also it does not work if we don't call SonicDBConfig::initializeGlobalConfig.
DBConnector Constructor with particular asic namespace gives seg fault
as SonicDBConfig::getDbId will fail since m_init will be set when call for host/global namespace and when next time called
it will not go for else condition (since m_init is already set)

judyjoseph:
> Is this comment on previous PR is addressed:
> 
> a) why it has to be different ? It has be consistent . Any particular reason why it need to be different ? Application should be doing same irrespective of library being used.
> 
> b)Also it does not work if we don't call SonicDBConfig::initializeGlobalConfig.
> DBConnector Constructor with particular asic namespace gives seg fault
> as SonicDBConfig::getDbId will fail since m_init will be set when call for host/global namespace and when next time called
> it will not go for else condition (since m_init is already set)

Yes @abdosi we don't fail now if we the database_global.json is not present

common/dbconnector.cpp
```diff
@@ -240,19 +407,15 @@ string DBConnector::getDbName() const
...
-        ret = new DBConnector(getDbId(),
-                               getContext()->tcp.host,
-                               getContext()->tcp.port,
-                               timeout);
-    else
-        ret = new DBConnector(getDbId(),
-                               getContext()->unix_sock.path,
-                               timeout);
-    ret->m_dbName = m_dbName;
+    ret = new DBConnector(getDbName(), timeout, (getContext()->connection_type == REDIS_CONN_TCP), getNamespace());
```
judyjoseph, line 418: There was a failure  I was debugging in sonic-swss/tests/mock_tests, even with the change in the PR where we do constructor delegation.

The reason being in this line 418, when we create a newDBConnector we call the DBConnector API with namespace argument. Since we don't define DBConnector() constructor with the namespace argument in the sonic-swss/tests/mock_tests/mock_dbconnector.cpp .. it come to this actual non-mocked handler and fails .

```
make[3]: Entering directory '/sonic/src/sonic-swss/tests/mock_tests'
make[4]: Entering directory '/sonic/src/sonic-swss/tests/mock_tests'
../../config/test-driver: line 107: 12950 Aborted                 (core dumped) "$@" > $log_file 2>&1
FAIL: tests
============================================================================
Testsuite summary for sonic-swss 1.0
============================================================================
# TOTAL: 1
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
terminate called after throwing an instance of 'std::system_error'
  what():  Unable to connect to redis: Cannot assign requested address
FAIL tests (exit status: 134)
```
**Two options to fix this**

1. Add the fix https://github.com/Azure/sonic-swss/pull/1362. Though there are tests failing in swss, they are not related.

2. Other option is to revert this function DBConnector::newConnector(unsigned int timeout) , to as it was before (https://github.com/Azure/sonic-swss-common/blob/00bcd1d0acfdb1fd33362beafaf41ea3f5c301aa/common/dbconnector.cpp#L243). So this API won't support the usage of namespace for now , can add it back later.

abdosi:
can you update submodule ? is it still blocked on sonic-swss ?

judyjoseph:
> can you update submodule ? is it still blocked on sonic-swss ?

Yes there is still a VS PR build failure .. though locally build went fine with the changes.  To look at why the PR build fails

common/dbconnector.h
```diff
@@ -60,12 +79,14 @@ class DBConnector
     DBConnector(int dbId, const std::string &hostname, int port, unsigned int timeout);
     DBConnector(int dbId, const std::string &unixPath, unsigned int timeout);
     DBConnector(const std::string &dbName, unsigned int timeout, bool isTcpConn = false);
+    DBConnector(const std::string &dbName, unsigned int timeout, bool isTcpConn, const std::string &nameSpace);
```
qiluo-msft, line 82: >nameSpace [](start = 100, length = 9)

nameSpace -> namespace

judyjoseph: @Qi cannot use namespace as this is a C++ keyword. may be namespaceName or name_space ?

qiluo-msft: Thanks! how about `netns` ?

PR #376, opened 2020-08-14
Dbconnector namespace support

judyjoseph:
Why:
  The commit for support for namespaces in DBConnector (#371) class was reverted earlier as it failed couple of tests in swss. It was found that the SWSS docker virtual switch API's create the DBConnector object using the constructor API's which was present before the multi-DB changes went in. During those times, the DBConnector classes used dbId alone and dbName was not filled. (https://github.com/Azure/sonic-swss/blob/eac1d898e12d03040cd7b4e3c4059960dbf0ae8d/tests/dvslib/dvs_database.py#L30)

The reason for the tests failure was because the dbName was coming as NULL when we try to create newConnector().

What:
Added the check to see if the dbName, namespace is empty, if so create the DBConnector using the old constructors which takes dbId as input. If not call the new constructor where we pass the dbName and namespaceName.

**With the above changes -- no further change needed in sonic-swss/tests/mock_tests/mock_dbconnector.cpp**

How:
Recreated the swss tests setup locally and saw the tests cases were passing.
```
johnar@c1ac715ae2d8:/sonic/swss/tests$ sudo make -j1 check
Making check in mock_tests
make[1]: Entering directory '/sonic/swss/tests/mock_tests'
make  check-TESTS
make[2]: Entering directory '/sonic/swss/tests/mock_tests'
make[3]: Entering directory '/sonic/swss/tests/mock_tests'
PASS: tests
============================================================================
Testsuite summary for sonic-swss 1.0
============================================================================
# TOTAL: 1
# PASS:  1
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
make[3]: Leaving directory '/sonic/swss/tests/mock_tests'
make[2]: Leaving directory '/sonic/swss/tests/mock_tests'
make[1]: Leaving directory '/sonic/swss/tests/mock_tests'
make[1]: Entering directory '/sonic/swss/tests'
make  check-TESTS
make[2]: Entering directory '/sonic/swss/tests'
make[3]: Entering directory '/sonic/swss/tests'
PASS: tests
============================================================================
Testsuite summary for sonic-swss 1.0
============================================================================
# TOTAL: 1
# PASS:  1
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
make[3]: Leaving directory '/sonic/swss/tests'
make[2]: Leaving directory '/sonic/swss/tests'
[...]

PR #378, opened 2020-09-01
Removed DB specific get api's from Selectable class

abdosi:
Why/What I did:-

Change 1:
As per review comments on https://github.com/Azure/sonic-swss-common/pull/376
removed DB specific API from selectable class
to keep it generic.

Change 2: 
However to access to the Dervied Class Redis Select
DB API's via python we need to downcast the the Selectable Object
to RedisSelect object. 
To do this added Helper function  in swig .i file Ref: http://www.swig.org/Doc3.0/Python.html

How I verify:
After this change updated client caclmgrd to call new downcast API and verified we are getting correct information.
Sample Client Side code after this change -

        (state, selectableObj) = sel.select(SELECT_TIMEOUT_MS)
            # Continue if select is timeout or selectable object is not return
            if state != swsscommon.Select.OBJECT:
                continue
            # Get the corresponding namespace from selectable object

            redisSelectObj = swsscommon.CastSelectableToRedisSelectObj(selectableObj)

            namespace = redisSelectObj.getDbConnector().getNamespace()

### DIFF
--- a/common/redisselect.h
+++ b/common/redisselect.h
@@ -20,8 +20,7 @@ public:
     bool hasCachedData() override;
     bool initializedWithData() override;
     void updateAfterRead() override;
-    int getDbConnectorId();
-    std::string getDbNamespace();
+    const DBConnector* getDbConnector() const;
 
     /* Create a new redisContext, SELECT DB and SUBSCRIBE */
     void subscribe(DBConnector* db, const std::string &channelName);
--- a/common/redisselect.h
+++ b/common/redisselect.h
@@ -20,8 +20,8 @@ public:
     bool hasCachedData() override;
     bool initializedWithData() override;
     void updateAfterRead() override;
-    int getDbConnectorId() override;
-    std::string getDbNamespace() override;
+    int getDbConnectorId();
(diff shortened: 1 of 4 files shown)


common/redisselect.h
```diff
@@ -20,8 +20,8 @@ class RedisSelect : public Selectable
     bool hasCachedData() override;
     bool initializedWithData() override;
     void updateAfterRead() override;
-    int getDbConnectorId() override;
-    std::string getDbNamespace() override;
+    int getDbConnectorId();
+    std::string getDbNamespace();
```
qiluo-msft, line 24: Suggest add only one function `const DBConnector* getDbConnector() const`, and caller could call `ret->getNamespace()`.

abdosi: @qiluo-msft Updated.

abdosi:
@judyjoseph Please update sonic-py-common client accordingly.

judyjoseph:
> @judyjoseph Please update sonic-py-common client accordingly.

sure will take care of updating the pmon application ..Thanks for this change !

PR #380, opened 2020-09-10
Implementation of System ports initialization, Interface & Neighbor Synchronization...

minionatwork:
This PR includes implementation of System ports initialization & orchestration, Interface & Neighbor Synchronization using GLOBAL_APP_DB, Inband Interface Configuration.

HLD: https://github.com/Azure/SONiC/pull/639


common/table.cpp
```diff
@@ -29,7 +29,8 @@ const TableNameSeparatorMap TableBase::tableNameSeparatorMap = {
    { STATE_DB,            TABLE_NAME_SEPARATOR_VBAR  },
    { GB_ASIC_DB,          TABLE_NAME_SEPARATOR_VBAR  },
    { GB_COUNTERS_DB,      TABLE_NAME_SEPARATOR_VBAR  },
-   { GB_FLEX_COUNTER_DB,  TABLE_NAME_SEPARATOR_VBAR  }
+   { GB_FLEX_COUNTER_DB,  TABLE_NAME_SEPARATOR_VBAR  },
+   { GLOBAL_APP_DB,       TABLE_NAME_SEPARATOR_COLON }
```
lguohan, line 33: better to use VBAR

common/schema.h
```diff
@@ -19,6 +19,7 @@ namespace swss {
 #define GB_ASIC_DB      8
 #define GB_COUNTERS_DB  9
 #define GB_FLEX_COUNTER_DB  10
+#define GLOBAL_APP_DB   11
```
vganesan-nokia, line 22: GLOBAL_APP_DB will be changed to CHASSIS_APP_DB

common/schema.h
```diff
@@ -267,6 +270,12 @@ namespace swss {
 #define CFG_FG_NHG_PREFIX                           "FG_NHG_PREFIX"
 #define CFG_FG_NHG_MEMBER                           "FG_NHG_MEMBER"
 
+#define CFG_SYSTEM_PORT_TABLE_NAME                  "SYSTEM_PORT"
+#define CFG_VOQ_INBAND_INTERFACE_TABLE_NAME         "VOQ_INBAND_INTERFACE"
+
+#define VOQ_SYSTEM_INTERFACE_TABLE_NAME             "SYSTEM_INTERFACE"
+#define VOQ_SYSTEM_NEIGH_TABLE_NAME                 "SYSTEM_NEIGH"
```
vganesan-nokia, line 277: Change VOQ to CHASSIS_APP to refer to the DB

common/table.cpp
vganesan-nokia: o.k. We'll change to use VBAR.

vganesan-nokia: Fixed

common/schema.h
vganesan-nokia: Fixed

common/schema.h
vganesan-nokia: Fixed

common/schema.h
```diff
@@ -21,6 +21,10 @@ namespace swss {
 #define GB_COUNTERS_DB  10
 #define GB_FLEX_COUNTER_DB  11
 
+//For chassis app db in supervisor/control card redis in chassis systems.
+//This DB ID is not to be used in local redis.
+#define CHASSIS_APP_DB  8
```
BrynXu, line 26: due to issue [5351](https://github.com/Azure/sonic-buildimage/issues/5351), the db id has to be unique even they are used in different redis server. In Arista PR 5283, it is set to 12, we need to be in sync the value.

vganesan-nokia: o.k. I'll set to CHASSIS_APP_DB id to 12.

vganesan-nokia: Fixed

eswaranb (approved):

eswaranb (approved):

lguohan (approved):

common/database_config.json
```diff
@@ -4,6 +4,11 @@
             "hostname" : "127.0.0.1",
             "port" : 6379,
             "unix_socket_path" : "/var/run/redis/redis.sock"
+        },
+        "redis_chassis":{
+            "hostname" : "redis_chassis.server",
+            "port" : 6380,
+            "unix_socket_path" : "/var/run/redis/redis_chassis.sock"
```
qiluo-msft, line 11: Please move chassis specific configuration out of the default database_config.json. The file here is to maintain backward-compatiblity to before multi-DB feature and minimal configuration for any sonic system.

For example, this file does not work with sonic-utilities.

If your use case need customization, just customize it and overwrite it in your use case only. Do not break others.

PR #381, opened 2020-09-11
Throw out_of_range exceptions with meaningful message for multi-DB and multi-namespace

qiluo-msft:
The original exception has too general info for debugging.
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  _Map_base::at
Aborted (core dumped)
```

After this fix:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  Failed to find 15 database in  namespace
Aborted (core dumped)
```

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -256,9 +256,9 @@ RedisInstInfo& SonicDBConfig::getRedisInfo(const std::string &dbName, const std:
         SWSS_LOG_ERROR("%s", msg.c_str());
         throw out_of_range(msg);
     }
-    auto& infos = foundNetns->second;
-    auto foundRedis = infos.find(getDbInst(dbName));
-    if (foundRedis == infos.end())
+    auto& redisInfos = foundNetns->second;
+    auto foundRedis = redisInfos.find(getDbInst(dbName, netns));
+    if (foundRedis == redisInfos.end())
     {
         string msg = "Failed to find the Redis instance for " + dbName + " database in " + netns + " namespace";
         SWSS_LOG_ERROR("%s", msg.c_str());
@@ -286,6 +286,14 @@ string SonicDBConfig::getSeparator(int dbId, const string &netns)
 {
     if (!m_init)
         initialize(DEFAULT_SONIC_DB_CONFIG_FILE);
+
+    if (!netns.empty())
+    {
+        if (!m_global_init)
+        {
+            SWSS_LOG_THROW("Initialize global DB config using API SonicDBConfig::initializeGlobalConfig");
+        }
+    }
     auto foundNetns = m_db_separator.find(netns);
     if (foundNetns == m_db_separator.end())
     {
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -191,50 +191,117 @@ void SonicDBConfig::validateNamespace(const string &netns)
         // If global initialization is not done, ask user to initialize global DB Config first.
         if (!m_global_init)
         {
-            SWSS_LOG_ERROR("Initialize global DB config first using API SonicDBConfig::initializeGlobalConfig \n");
-            throw runtime_error("Initialize global DB config using API SonicDBConfig::initializeGlobalConfig");
+            SWSS_LOG_THROW("Initialize global DB config using API SonicDBConfig::initializeGlobalConfig");
         }
 
         // Check if the namespace is valid, check if this is a key in either of this map
         unordered_map<string, unordered_map<string, RedisInstInfo>>::const_iterator entry = m_inst_info.find(netns);
         if (entry == m_inst_info.end())
         {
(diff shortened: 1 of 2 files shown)


qiluo-msft:
@dzhangalibaba Please also help review.

common/dbconnector.cpp
```diff
@@ -191,50 +191,117 @@ void SonicDBConfig::validateNamespace(const string &netns)
...
+        {
+            SWSS_LOG_THROW("Initialize global DB config using API SonicDBConfig::initializeGlobalConfig");
+        }
+    }
+    auto foundNetns = m_db_info.find(netns);
+    if (foundNetns == m_db_info.end())
+    {
+        string msg = "Namespace " + netns + " is not a valid namespace name in config file";
+        SWSS_LOG_ERROR("%s", msg.c_str());
+        throw out_of_range(msg);
```
daall, line 225: What's the difference between this and `SWSS_LOG_THROW`?

daall (commented):
LGTM, just one small q. Thanks!

common/dbconnector.cpp
qiluo-msft: `SWSS_LOG_THROW` will throw runtime_error. Since the unit test already cover the exception type, I keep original behavior.

---
In reply to: [487145183](https://github.com/Azure/sonic-swss-common/pull/381#discussion_r487145183) [](ancestors = 487145183)

daall: Makes sense, thanks!

common/dbconnector.cpp
```diff
@@ -191,50 +191,117 @@ void SonicDBConfig::validateNamespace(const string &netns)
...
+    {
+        if (!m_global_init)
+        {
+            SWSS_LOG_THROW("Initialize global DB config using API SonicDBConfig::initializeGlobalConfig");
+        }
+    }
+    auto foundNetns = m_db_info.find(netns);
+    if (foundNetns == m_db_info.end())
+    {
+        string msg = "Namespace " + netns + " is not a valid namespace name in config file";
```
abdosi, line 223: @qiluo-msft general question for empty namepsace the message will will not have default/global/empty string in message ?

qiluo-msft: It just print zero length string. I have an example in PR description. Since it is for debug purpose, I prefer not convert it.

---
In reply to: [487166852](https://github.com/Azure/sonic-swss-common/pull/381#discussion_r487166852) [](ancestors = 487166852)

common/dbconnector.cpp
```diff
@@ -191,50 +191,117 @@ void SonicDBConfig::validateNamespace(const string &netns)
...
+        SWSS_LOG_ERROR("%s", msg.c_str());
+        throw out_of_range(msg);
+    }
+    auto& infos = foundNetns->second;
```
judyjoseph, line 227: Could change the variable name to like dbInfos ?

common/dbconnector.cpp
```diff
@@ -191,50 +191,117 @@ void SonicDBConfig::validateNamespace(const string &netns)
...
+        }
+    }
+    auto foundNetns = m_inst_info.find(netns);
+    if (foundNetns == m_inst_info.end())
+    {
+        string msg = "Namespace " + netns + " is not a valid namespace name in Redis instances in config file";
+        SWSS_LOG_ERROR("%s", msg.c_str());
+        throw out_of_range(msg);
+    }
+    auto& infos = foundNetns->second;
```
judyjoseph, line 259: Could change the variable name to like instInfos ?

common/dbconnector.cpp
```diff
@@ -191,50 +191,117 @@ void SonicDBConfig::validateNamespace(const string &netns)
...
+    auto foundRedis = infos.find(getDbInst(dbName));
```
judyjoseph, line 260: Need to pass the netns as second argument to getDbInst()  to get the RedisInfo from the right namespace.

common/dbconnector.cpp
```diff
@@ -191,50 +191,117 @@ void SonicDBConfig::validateNamespace(const string &netns)
...
+    return getDbInfo(dbName, netns).separator;
 }
 
 string SonicDBConfig::getSeparator(int dbId, const string &netns)
 {
     if (!m_init)
         initialize(DEFAULT_SONIC_DB_CONFIG_FILE);
-    validateNamespace(netns);
-    return m_db_separator[netns].at(dbId);
+    auto foundNetns = m_db_separator.find(netns);
```
judyjoseph, line 289: To add the check for netns.empty() here.

Or may be I was thinking, since we are using this similar logic in the other API's as well -- can we call validateNamespace() itself, and if  the namespace is valid access with m_db_info[netns]/m_inst_info[netns]/m_db_separator[netns].

judyjoseph (approved):

PR #384, opened 2020-09-15
Throw proper exceptions when talking with Redis

qiluo-msft:
Original we don't handle error code when talking with Redis by hiredis APIs. So we will realize it happening far away from the first place and get confusing error message.

Old
```
C++ exception with description "Memory exception, reply is null: Cannot allocate memory" thrown in the test body.
```

New
```
C++ exception with description "RedisResponseError: Failed to redisGetReply with CLIENT GETNAME: err=3: errstr=Server closed the connection" thrown in the test body.
```

### DIFF
--- a/common/redisreply.h
+++ b/common/redisreply.h
@@ -2,12 +2,36 @@
 #define __REDISREPLY__
 
 #include <hiredis/hiredis.h>
+#include <string>
 #include <stdexcept>
 #include "rediscommand.h"
 
 namespace swss {
 
 class DBConnector;
+
+class RedisResponseError : public std::runtime_error
+{
+    int m_err;
+    std::string m_errstr;
+    mutable std::string m_message;
+public:
+    RedisResponseError(const std::string& arg, redisContext *ctx)
+        : std::runtime_error(arg)
+        , m_err(ctx->err)
+        , m_errstr(ctx->errstr)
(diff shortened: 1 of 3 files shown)


prsunny (dismissed):
lgtm

common/redisreply.h
```diff
@@ -2,13 +2,37 @@
...
+        , m_err(ctx->err)
+        , m_errstr(ctx->errstr)
+    {
+    }
+
+    const char *what() const noexcept override
+    {
+        if (m_message.empty())
+        {
+            m_message = std::string("RedisResponseError: ") + std::runtime_error::what() + ": err=" + std::to_string(m_err) + ": errstr=" + m_errstr;
```
shi-su, line 30: I am not sure if the colon is the best choice in `": errstr="`. I feel a comma might be better.

shi-su (commented):
Is it necessary to do something for `redisGetReply(m_db->getContext(), (void**)&reply);` at line 74 in `common/redispipeline.h`?

common/redisreply.h
qiluo-msft: Thanks! Fix according to suggestion

prsunny (approved):

shi-su (approved):

PR #385, opened 2020-09-18
add RESTAPI_DB and renumber the GB_*_DB ids to match what is in sonic…

slogan621:
swss-common: sync db id numbers with changes in sonic-buildimage

* add in restapi_db constant that was missing
* renumber gearbox tables accordingly.

  Signed-off-by: syd.logan@broadcom.com

### DIFF
--- a/common/database_config.json
+++ b/common/database_config.json
@@ -52,18 +52,23 @@
             "separator": "|",
             "instance" : "redis"
         },
-        "GB_ASIC_DB" : {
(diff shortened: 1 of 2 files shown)


slogan621:
@lguohan - as discussed, change the DB numbers here to match sonic-buildimage.

prsunny:
I dint see this PR but raised another one https://github.com/Azure/sonic-swss-common/pull/386

PR #387, opened 2020-09-28
Implement DBInterface/SonicV2Connector in C++

qiluo-msft:
This is the first step for merging sonic-py-swsssdk into sonic-swss-common. The whole plan is roughly listed as below
- [ ] This PR
- [ ] Replace production code DBInterface/SonicV2Connector with C++ implementation. Test code could still using old python implementation
     - [ ] https://github.com/Azure/sonic-snmpagent/pull/162
- [ ] Replace DBInterface with C++ implementation in sonic-py-swsssdk
- [ ] Replace production code SonicV2Connector with C++ implementation
     - [ ] sonic-utilites
     - [ ] ...
- [ ] Depreciate SonicV2Connector in sonic-py-swsssdk
- [ ] Move remaining classes in sonic-py-swsssdk into sonic-swss-common
- [ ] Remove sonic-py-swsssdk from built image
- [ ] TBD: remove sonic-py-swsssdk from test code

What I did in this PR:
1. Extract RedisContext class from DBConnector class
2. Extract DBConnector methods: subscribe/psubscribe/publish/config_set
3. Implement DBInterface/SonicV2Connector in C++ which deriving from python implementation in https://github.com/Azure/sonic-py-swsssdk/
   - maintain the interfaces through the SWIG wrapper functions
   - Add set_redis_kwargs for easy handling python kwargs.


common/dbinterface.h
```diff
@@ -104,4 +106,115 @@ class DBInterface
     int m_port = 6379;
 };
 
+class SonicV2Connector
```
kcudnik, line 109: move this to separate file ? as 1 class per file

kcudnik: also why not split this to .h/.cpp and move implementaion to cpp file ? problem here is that if you want to update 1 function, then all files in all repos where you used that file .h will need to recompile since you touched that file, but if you only change cpp file, then only that file needs to be recompiled and only .o would need to be linked

common/dbinterface.h
```diff
@@ -104,4 +106,115 @@ class DBInterface
...
+    {
+        return SonicDBConfig::getDbHostname(db_name, m_netns);
+    }
+
+    int get_db_port(const std::string& db_name)
+    {
+        return SonicDBConfig::getDbPort(db_name, m_netns);
+    }
+
+    DBInterface dbintf;
```
kcudnik, line 215: prefix m_ as other private members ?

common/dbinterface.cpp
```diff
@@ -0,0 +1,347 @@
...
+                , std::forward_as_tuple(dbId, m_host, m_port, 0));
+    }
+    else
+    {
+        rc = m_redisClient.emplace(std::piecewise_construct
+                , std::forward_as_tuple(dbName)
+                , std::forward_as_tuple(dbId, m_unix_socket_path, 0));
+    }
+    bool inserted = rc.second;
+    if (inserted)
```
shi-su, line 322: Is `inserted` being false also a failure for this `_onetime_connect`? If so, is it necessary to throw an exception so that it could retry in `_persistent_connect`?

common/sonicv2connector.cpp
```diff
@@ -0,0 +1,107 @@
...
+    : m_use_unix_socket_path(use_unix_socket_path)
+    , m_netns(netns)
+{
+}
+
+void SonicV2Connector::connect(const std::string& db_name, bool retry_on)
+{
+    if (m_use_unix_socket_path)
+    {
+        SWSS_LOG_INFO("connec1: %s", get_db_socket(db_name).c_str());
```
shi-su, line 17: Would it be better to log this as something like unix socket rather than "connec1"? I think "connec1" is not a very clear concept.

common/sonicv2connector.cpp
```diff
@@ -0,0 +1,107 @@
...
+        m_dbintf.set_redis_kwargs(get_db_socket(db_name), "", 0);
+    }
+    else
+    {
+        SWSS_LOG_INFO("connec2: %s %d", get_db_hostname(db_name).c_str(), get_db_port(db_name));
```
shi-su, line 22: I guess this should be TCP then. How about explicitly log this as TCP?

tests/redis_ut.cpp
```diff
@@ -311,6 +313,21 @@ TEST(DBConnector, RedisClientName)
...
+    DBInterface dbintf;
+    dbintf.set_redis_kwargs("", "127.0.0.1", 6379);
+    dbintf.connect(15, "TEST_DB");
+
+    SonicV2Connector db;
+    db.connect("TEST_DB");
+    db.set("TEST_DB", "key0", "field1", "value2");
+    auto fvs = db.get_all("TEST_DB", "key0");
+    auto rc = fvs.find("field1");
+    EXPECT_EQ(rc != fvs.end(), true);
```
shi-su, line 327: I remember there are functions `EXPECT_NE` and `EXPECT_TRUE`. I think they may be better here than expecting != to be true.

common/dbinterface.h
```diff
@@ -0,0 +1,109 @@
...
+    // z - Sorted set commands
+    // x - Expired events (events generated every time a key expires)
+    // e - Evicted events (events generated when a key is evicted for maxmemory)
+    // A - Alias for g$lshzxe, so that the "AKE" string means all the events.
+    // ACS Redis db mainly uses hash, therefore h is selected.
+    static constexpr const char *KEYSPACE_EVENTS = "KEA";
+
+    std::unordered_map<std::string, std::shared_ptr<DBConnector>> keyspace_notification_channels;
+
+    std::unordered_map<std::string, DBConnector> m_redisClient;
```
shi-su, line 102: I thought you were deprecating the redisClient class in another PR. I understand this `m_redisClient` is different from that class, just wondering if this concept should continue to be used.

kcudnik (dismissed):

common/dbinterface.cpp
qiluo-msft: If this happens, retry will not help, since the dbName is already in `m_redisClient`. It is safe to ignore because this database is already connected by a previous connector.

---
In reply to: [500470075](https://github.com/Azure/sonic-swss-common/pull/387#discussion_r500470075) [](ancestors = 500470075)

common/dbinterface.h
qiluo-msft: It is rewritten from python [https://github.com/Azure/sonic-py-swsssdk/blob/master/src/swsssdk/interface.py](https://github.com/Azure/sonic-py-swsssdk/blob/master/src/swsssdk/interface.py)
To make code similar, I partially kept the name.

---
In reply to: [500486310](https://github.com/Azure/sonic-swss-common/pull/387#discussion_r500486310) [](ancestors = 500486310)

common/sonicv2connector.cpp
qiluo-msft: Sorry, debug code removed

---
In reply to: [500481974](https://github.com/Azure/sonic-swss-common/pull/387#discussion_r500481974) [](ancestors = 500481974)

common/sonicv2connector.cpp
qiluo-msft: Sorry, debug code removed

---
In reply to: [500482827](https://github.com/Azure/sonic-swss-common/pull/387#discussion_r500482827) [](ancestors = 500482827)

tests/redis_ut.cpp
qiluo-msft: Resolved.

---
In reply to: [500484365](https://github.com/Azure/sonic-swss-common/pull/387#discussion_r500484365) [](ancestors = 500484365)

tests/test_redis_ut.py
```diff
@@ -1,7 +1,17 @@
 import time
+import pytest
 from threading import Thread
 from pympler.tracker import SummaryTracker
 from swsscommon import swsscommon
+from swsscommon.swsscommon import DBInterface, SonicV2Connector, SonicDBConfig
+
+existing_file = "./tests/redis_multi_db_ut_config/database_config.json"
+nonexisting_file = "./tests/redis_multi_db_ut_config/database_config_nonexisting.json"
+global_existing_file = "./tests/redis_multi_db_ut_config/database_global.json"
```
shi-su, line 10: Just wondering how are `nonexisting_file` and `global_existing_file` useful here?

shi-su (dismissed):

shi-su (approved):

PR #389, opened 2020-10-01
Schema update for BGP internal neighbor table

judyjoseph:
Add a new BGP_INTERNAL_NEIGHBOR table in the DB schema. 

This table will contain the internal BGP sessions created between BGP instances running per ASIC namespace in a multi-asic platform.


pavel-shirshov (approved):

judyjoseph:
Add a new BGP_INTERNAL_NEIGHBOR table in the DB schema. 
This table will contain the internal BGP sessions created between BGP instances running per ASIC namespace in a multi-asic platform.

PR #390, opened 2020-10-06
Add REDIS async support

martin-belanger:
This pull request is mainly to add support for REDIS asynchronous APIs (from the [hiredis](https://github.com/redis/hiredis) library).

I also fixed a potential memory leak in common/rediscommand.cpp.

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -688,4 +688,67 @@ shared_ptr<string> DBConnector::blpop(const string &list, int timeout)
     throw runtime_error("GET failed, memory exception");
 }
 
+
+/*******************************************************************************
+    / \   ___ _   _ _ __   ___
+   / _ \ / __| | | | '_ \ / __|
+  / ___ \\__ \ |_| | | | | (__
+ /_/   \_\___/\__, |_| |_|\___|
+              |___/
+*******************************************************************************/
+DBConnector_async::DBConnector_async(const std::string  & db_name,
+                                     void               * user_ctx_p) :
+    db_name_m(db_name),
+    db_id_m(SonicDBConfig::getDbId(db_name_m)),
+    sock_addr_m(SonicDBConfig::getDbSock(db_name_m)),
+    user_ctx_pm(user_ctx_p)
+{
+    ac_pm = redisAsyncConnectUnix(sock_addr_m.c_str());
+
+    if (ac_pm->err != 0)
+    {
+        std::string errmsg("Unable to connect to redis: (" + std::to_string(ac_pm->err) + ')');
+
+        if ((ac_pm->errstr != nullptr) && (ac_pm->errstr[0] != '\0'))
+             errmsg += " " + std::string(ac_pm->errstr);
+
+        redisAsyncFree(ac_pm);
+        ac_pm = nullptr;
+
+        throw std::system_error(std::make_error_code(errc::address_not_available), errmsg);
+    }
+
+    ac_pm->data = this;
+
+    command(nullptr, nullptr, "SELECT %d", db_id_m);
+}
+
+DBConnector_async::~DBConnector_async()
+{
+    if (ac_pm != nullptr)
+    {
+        // We can't use redisAsyncFree() here because there may
+        // be pending messages to be sent or received. redisAsyncDisconnect()
+        // will ensure that all pending messages are processed before the
+        // context gets deleted.
+        redisAsyncDisconnect(ac_pm);
+        ac_pm = nullptr;
+    }
+}
+
+int DBConnector_async::command(redisCallbackFn * cb_func_p, void * cb_data_p, const char * format_p, ...)
+{
+    va_list ap;
+    int status;
+    va_start(ap,format_p);
+    status = redisvAsyncCommand(ac_pm, cb_func_p, cb_data_p, format_p, ap);
+    va_end(ap);
+    return status;
+}
+
+int DBConnector_async::formatted_command(redisCallbackFn * cb_func_p, void * cb_data_p, const char * cmd_p, size_t len)
+{
+    return redisAsyncFormattedCommand(ac_pm, cb_func_p, cb_data_p, cmd_p, len);
+}
+
 }
--- a/common/dbconnector.h
+++ b/common/dbconnector.h
@@ -8,6 +8,7 @@
 #include <memory>
 
 #include <hiredis/hiredis.h>
+#include <hiredis/async.h>      // redisAsyncContext
 #include "rediscommand.h"
 #include "redisreply.h"
 #define EMPTY_NAMESPACE std::string()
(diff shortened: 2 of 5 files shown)


lguohan:
can you separate the bug fix from this pr? we can review and merge it separately.

common/rediscommand.cpp
```diff
@@ -18,6 +18,12 @@ RedisCommand::~RedisCommand()
 
 void RedisCommand::format(const char *fmt, ...)
 {
+    if (temp != nullptr)
+    {
+        redisFreeCommand(temp);
+        temp = nullptr;
+    }
```
qiluo-msft, line 25: This is a valid bug fix. Could you please add a unit test, so it will fail on old code but pass with your fix.

martin-belanger: Hi @qiluo-msft . I'm not sure how to add a unit test for this. The leak will only happen if someone calls `RedisCommand::format()` (or `RedisCommand::formatArgv()`) more than once. The pointer to the leaked memory, `temp`, is a private member and cannot be tested outside the class `RedisCommand` itself. I'm running out of ideas on how to write a unit test for this. Any suggestions would be greatly appreciated.

martin-belanger: BTW, while looking at the existing unit tests, I found a place where a memory leak would occur. The leak is in file `tests/redis_state_ut.cpp`, line 738, as follows:
```
738)  RedisCommand keys;
739)  keys.format("KEYS %s*", (c.getStateHashPrefix() + tableName).c_str());
740)  RedisReply r(&db, keys, REDIS_REPLY_ARRAY);
741)  EXPECT_EQ(r.getContext()->elements, (size_t) 0);
742)  // Verify number of objects
743)  keys.format("KEYS %s:*", tableName.c_str());
744)  RedisReply r2(&db, keys, REDIS_REPLY_ARRAY);
745)  EXPECT_EQ(r2.getContext()->elements, (size_t) numOfKeys);
``` 
At line 739 we invoke `keys.format()` a first time. At line 743 we invoke `keys.format()` a second time. The second time the function is invoked, the memory that was allocated the first time the function was invoked is not freed and will leak.

qiluo-msft: Understand the difficulty. Could you separate the bug fix from this pr? The bug fix could be merged sooner.
We need more effort to review the new feature.

---
In reply to: [501238042](https://github.com/Azure/sonic-swss-common/pull/390#discussion_r501238042) [](ancestors = 501238042)

martin-belanger: Ok I will submit a separate pull request

martin-belanger: Submitted this pull request that only contains the memory leak fix.
https://github.com/Azure/sonic-swss-common/pull/392

lguohan:
@qiluo-msft, any update on this pr?

common/dbconnector.h
```diff
@@ -171,5 +172,148 @@ void DBConnector::hmset(const std::string &key, InputIterator start, InputIterat
...
+     * @param cmd A formatted command to be sent to the REDIS server
+     * @param len The length of %cmd
+     *
+     * @return 0 on success, otherwise the status returned by
+     *         redisAsyncFormattedCommand()
+     */
+    int formatted_command(redisCallbackFn * cb_func_p, void * cb_data_p, const char * cmd_p, size_t len);
+
+private:
+    const std::string    db_name_m;
```
qiluo-msft, line 311: >db_name_m [](start = 25, length = 9)

We are using m_cameCase as member variable name convention. For this case, suggest `m_dbName`

common/dbconnector.h
```diff
@@ -171,5 +172,148 @@ void DBConnector::hmset(const std::string &key, InputIterator start, InputIterat
...
+     *
+     */
+    redisAsyncContext * context() const { return ac_pm; }
+
+    /**
+     * @brief Return the user context that was provided in the constructor.
+     *
+     * @return User context pointer.
+     */
+    void * get_user_ctx() const { return user_ctx_pm; }
```
qiluo-msft, line 259: >get_user_ctx [](start = 11, length = 12)

Naming convention: cameCaseForFunctionName. So suggest `getUserContext()`

common/dbconnector.h
```diff
@@ -171,5 +172,148 @@ void DBConnector::hmset(const std::string &key, InputIterator start, InputIterat
...
+     *  g_source_attach(redis_source_new(db.context()), main_ctx_p);
+     *  .
+     *  .
+     *
+     *  g_main_loop_run(loop_p);
+     *
+     * @endcode
+     *
+     */
+    redisAsyncContext * context() const { return ac_pm; }
```
qiluo-msft, line 252: >  [](start = 23, length = 1)

Suggest remove extra blank after `*`

common/dbconnector.h
```diff
@@ -171,5 +172,148 @@ void DBConnector::hmset(const std::string &key, InputIterator start, InputIterat
...
+     */
+    void * get_user_ctx() const { return user_ctx_pm; }
+
+    /**
+     * @brief Get the DB name (e.g. "CONFIG_DB", "APPL_DB", etc.). This is the
+     *        same name that was provided to the constructor.
+     *
+     * @return The DB name associated with this connector.
+     */
+    const std::string & db_name() const { return db_name_m; }
```
qiluo-msft, line 267: >  [](start = 21, length = 1)

Suggest remove extra blank before `&`

qiluo-msft, line 267: >string [](start = 15, length = 6)

Better return non reference string, so caller does not worry about scope.

common/dbconnector.h
```diff
@@ -171,5 +172,148 @@ void DBConnector::hmset(const std::string &key, InputIterator start, InputIterat
...
+ *          The hiredis library provides a number of adapters for 3rd party
+ *          event loop libraries such as GLib, libevent, qt, etc.
+ *          Ref: <a href="https://github.com/redis/hiredis/tree/master/adapters">hiredis/adapters/[*.h]</a>
+ *
+ *          No additional thread is created when using the async hiredis APIs.
+ *          The hiredis context is simply "hooked up" to the event loop and
+ *          that allows the event loop to dispatch the "events" (i.e. replies
+ *          from the REDIS server) to their corresponding callback functions.
+ */
+class DBConnector_async
```
qiluo-msft, line 200: >DBConnector_async [](start = 6, length = 17)

Could you provide use cases? We also require unit test cases for new classes.

common/dbconnector.cpp
```diff
@@ -688,4 +688,67 @@ shared_ptr<string> DBConnector::blpop(const string &list, int timeout)
...
+{
+    va_list ap;
+    int status;
+    va_start(ap,format_p);
+    status = redisvAsyncCommand(ac_pm, cb_func_p, cb_data_p, format_p, ap);
+    va_end(ap);
+    return status;
+}
+
+int DBConnector_async::formatted_command(redisCallbackFn * cb_func_p, void * cb_data_p, const char * cmd_p, size_t len)
```
qiluo-msft, line 749: >formatted_command [](start = 23, length = 17)

`command()` and `formatted_command` fit better in the class `RedisReply`, or its subclass.
Is it a good idea to have a `redisCallbackFn` member in `RedisReply`, or its subclass?

common/dbconnector.h
```diff
@@ -171,5 +172,148 @@ void DBConnector::hmset(const std::string &key, InputIterator start, InputIterat
...
+     * @return 0 on success, otherwise the status returned by
+     *         redisAsyncFormattedCommand()
+     */
+    int formatted_command(redisCallbackFn * cb_func_p, void * cb_data_p, const char * cmd_p, size_t len);
+
+private:
+    const std::string    db_name_m;
+    const int            db_id_m = -1;
+    std::string          sock_addr_m;
+    redisAsyncContext  * ac_pm = nullptr;
```
qiluo-msft, line 314: >redisAsyncContext [](start = 4, length = 17)

Suggest create a wrapper class: `RedisAsyncContext` derived from `RedisContext`.
`m_conn` should point to `ac_pm->c`

common/table.cpp
```diff
@@ -241,3 +241,46 @@ string Table::stripSpecialSym(const string &key)
...
+}
+
+int Table_async::hset(redisCallbackFn * cb_func_p, void * cb_data_p, const std::string & key_r, const std::string & field_r, const std::string & value_r)
+{
+    std::string  key = getKeyName(key_r);
+    swss::RedisCommand cmd;
+    cmd.formatHSET(key, field_r, value_r);
+    return dbconn_rm.formatted_command(cb_func_p, cb_data_p, cmd.c_str(), cmd.length());
+}
+
```
qiluo-msft, line 286: Remove extra blank line

common/table.cpp
```diff
@@ -241,3 +241,46 @@ string Table::stripSpecialSym(const string &key)
...
+int Table_async::hdel(redisCallbackFn * cb_func_p, void * cb_data_p, const std::string & key_r, const std::string & field_r)
+{
+    std::string  key = getKeyName(key_r);
+    swss::RedisCommand cmd;
+    cmd.formatHDEL(key, field_r);
+    return dbconn_rm.formatted_command(cb_func_p, cb_data_p, cmd.c_str(), cmd.length());
+}
+
+int Table_async::hget(redisCallbackFn * cb_func_p, void * cb_data_p, const std::string & key_r, const std::string & field_r)
```
qiluo-msft, line 271: >hget [](start = 17, length = 4)

Functions in this file fit better in the class `DBConnector` or its subclass.
It has less relationship with the `Table` concept.

qiluo-msft (changes requested):
As comments. Please also resolve conflicts and add testcase.

common/dbconnector.h
martin-belanger: The use cases are described in the Doxygen text above. Basically, this allows designing processes around an event loop such as GLib, libevent, qt, etc. as described in hiredis documentation: https://github.com/redis/hiredis/tree/master/adapters

common/dbconnector.cpp
martin-belanger: The RedisReply class is designed specifically for synchronous calls (i.e. blocking calls). When using the Async library (i.e. non-blocking) we do not wait for a reply. The reply comes back at a later time and is dispatched by the event loop (implemented by GLib, libevent, qt, or others) to a callback function.  

Async is a completely different approach and the RedisReply does not apply.

common/dbconnector.h
martin-belanger: I cannot derive RedisAsyncContext from RedisContext. RedisAsyncContext is not my definition. It is defined in the hiredis library (a 3rd party library).

common/table.cpp
martin-belanger: I'm just following the same model that was used for `class Table`. Just like `class Table` defines `Table::hget()`, `Table::hset()`, etc., `class Table_async` defines `Table_async::hget()`, `Table_async::hset()`, etc.

common/dbconnector.h
qiluo-msft: Thanks I read the Doxygen text.
I am asking about how do you plan to use the new classes in downstream repoes? Do you already have a PR following?
Require unit test cases in this PR. That also serves as use case.

---
In reply to: [532125398](https://github.com/Azure/sonic-swss-common/pull/390#discussion_r532125398) [](ancestors = 532125398)

common/dbconnector.h
qiluo-msft: Sorry for misleading. I did not mean `redisAsyncContext` in hiredis repo.
I mean in this repo, create a wrapper class: `RedisAsyncContext` derived from `RedisContext` (not `redisContext` in hiredis repo).

---
In reply to: [532126556](https://github.com/Azure/sonic-swss-common/pull/390#discussion_r532126556) [](ancestors = 532126556)

common/table.cpp
```diff
@@ -241,3 +241,46 @@ string Table::stripSpecialSym(const string &key)
...
+{
+    std::string  key = getKeyName(key_r);
```
qiluo-msft, line 273: >   [](start = 15, length = 2)

Use one blank character

common/dbconnector.h
martin-belanger: This will be used by hamd (pull request: https://github.com/Azure/sonic-buildimage/pull/5553).

hamd is the Host Account Management Daemon. It is designed as an asynchronous, non-blocking, event-driven process. It uses the GLib event-driven main-loop framework (ref. https://www.freedesktop.org/software/gstreamer-sdk/data/docs/latest/glib/glib-The-Main-Event-Loop.html). 

The hiredis library provides GLib bindings. hamd uses the hiredis asynchronous APIs to make non-blocking, asynchronous, requests to the REDIS server. When responses or events are received from the REDIS server, they get dispatched to callback functions by the GLib main-loop. 

The following files show how hamd uses the new async APIs:
https://github.com/martin-belanger/sonic-buildimage/blob/master/src/ham/hamd/hamd_redis.h
https://github.com/martin-belanger/sonic-buildimage/blob/master/src/ham/hamd/hamd_redis.cpp

This shows how hamd's GLib Main-loop is created:
https://github.com/martin-belanger/sonic-buildimage/blob/master/src/ham/hamd/hamd_main.cpp

common/dbconnector.h
martin-belanger: I see that a new class named RedisContext was introduced after I initially submitted my pull request. I understand what you are asking. 

By the way, since I submitted this pull request (2 months ago) I have been reassigned to a new project. I have very little time for SONiC now, but let me see see what I can do. Maybe one of my colleagues can take this over.

martin-belanger:
I just added a new commit to address the syntax issues (camel-casing, blanks, etc.) 

I will look into adding a class `RedisAsyncContext` derived from `RedisContext` when I find the time or I will find someone else to do that.

common/dbconnector.h
qiluo-msft: Sorry for misleading, I propose to  return basic `std::string`.

---
In reply to: [532125270](https://github.com/Azure/sonic-swss-common/pull/390#discussion_r532125270) [](ancestors = 532125270)

PR #394, opened 2020-10-14
Added support in Swig file to cast Selectable object to Subscriber Table object

abdosi:
What I did:
Added support in Swig file to cast Selectable object to SubscriberTable Object. 
Updated to use template function.

Why I did:
This way in application code we can  find out which subscribe
table select has return.

How I verify:
Verify after cast able to call getTableName() API and it return correct value.

### DIFF
--- a/pyext/swsscommon.i
+++ b/pyext/swsscommon.i
@@ -52,17 +52,15 @@
 }
 
 %inline %{
-swss::RedisSelect *CastSelectableToRedisSelectObj(swss::Selectable *temp) {
-  return dynamic_cast<swss::RedisSelect *>(temp);
+template <typename T>
+T castSelectableObj(swss::Selectable *temp)
+{
+    return dynamic_cast<T>(temp);
 }
 %}
 
-%inline %{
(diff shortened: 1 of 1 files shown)


pyext/swsscommon.i
```diff
@@ -57,6 +57,13 @@ swss::RedisSelect *CastSelectableToRedisSelectObj(swss::Selectable *temp) {
 }
 %}
 
+%inline %{
+swss::SubscriberStateTable *CastSelectableToSubscriberTableObj(swss::Selectable *temp) {
+  return dynamic_cast<swss::SubscriberStateTable *>(temp);
+}
+%}
```
qiluo-msft, line 64: I am ok with the code.
I am wondering if you can merge two `inline` blocks. Extract a template a function, and specialized on the 2 types.

abdosi: @qiluo-msft Thanks for the suggestion. I have updated accordingly.

qiluo-msft (approved):

PR #395, opened 2020-10-14
CHASSIS_STATE_DB on control-card for chassis state

mprabhu-nokia:
On a modular chassis, add a GLOBAL_STATE_DB where all state
information from line-cards can be persisted.

One example is thermal sensors information from each of the line
cards are stored in GLOBAL_STATE_DB. Control card can access this
information to feed to the thermal-policy, fan algorithm etc.


common/database_config.json
```diff
@@ -71,6 +71,11 @@
             "id" : 11,
             "separator": "|",
             "instance" : "redis"
+        },
+        "GLOBAL_STATE_DB" : {
```
Staphylo, line 75: Is `GLOBAL` the right terminology here?
I remember `GLOBAL` was used for multi-asic and `CHASSIS` for modular systems.

Staphylo: https://github.com/Azure/sonic-buildimage/pull/5624/files

mprabhu-nokia: Fixed.

Staphylo (approved):

PR #396, opened 2020-10-19
Fix SonicV2Connector interfaces

qiluo-msft:
1. pyext will pass empty string instead of None to SonicV2Connector ctor
2. add database name attributes
3. FieldValueMap class emulate the python dict interface


shi-su (commented):
LGTM

lguohan (approved):

PR #400, opened 2020-10-23
Implement FieldValueMap update method

qiluo-msft:
So in python, it will emulate a dict.

### DIFF
--- a/pyext/swsscommon.i
+++ b/pyext/swsscommon.i
@@ -43,14 +43,9 @@
             return defval
 
     def _FieldValueMap__update(self, *args, **kwargs):
-        if args:
-            if len(args) > 1:
-                raise TypeError("update expected at most 1 arguments, got %d" % len(args))
-            other = dict(args[0])
-            for key in other:
-                self[key] = other[key]
-        for key in kwargs:
-            self[key] = kwargs[key]
+        other = dict(*args, **kwargs)
+        for key in other:
+            self[key] = other[key]
 
     FieldValueMap.get = _FieldValueMap__get
     FieldValueMap.update = _FieldValueMap__update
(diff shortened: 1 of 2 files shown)


tests/test_redis_ut.py
```diff
@@ -145,9 +145,23 @@ def test_DBInterface():
...
     assert fvs.get("field1", "default") == "value2"
     assert fvs.get("nonfield", "default") == "default"
 
+    # Test dict.update()
+    other = { "field1": "value3", "field4": "value4" }
+    fvs.update(other)
+    assert len(fvs) == 2
+    assert fvs["field1"] == "value3"
+    assert fvs["field4"] == "value4"
+    fvs.update()
```
shi-su, line 159: Does this `fvs.update()` mean we keep existing fields and values unchanged and do not add new ones? It appears that nothing is checked for this operation.

pyext/swsscommon.i
```diff
@@ -42,7 +42,13 @@
         else:
             return defval
 
+    def _FieldValueMap__update(self, *args, **kwargs):
+        other = dict(*args, **kwargs)
```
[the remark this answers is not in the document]
qiluo-msft: It is actually preferred by some. sample: [https://www.programiz.com/python-programming/methods/dictionary/update](https://www.programiz.com/python-programming/methods/dictionary/update)

---
In reply to: [510627420](https://github.com/Azure/sonic-swss-common/pull/400#discussion_r510627420) [](ancestors = 510627420)

tests/test_redis_ut.py
qiluo-msft: Add comment and more asserts.

---
In reply to: [510641351](https://github.com/Azure/sonic-swss-common/pull/400#discussion_r510641351) [](ancestors = 510641351)

shi-su (approved):

PR #404, opened 2020-10-29
Change logging from ERROR to INFO for debugs related to database_global

judyjoseph:
Fix for the issue : https://github.com/Azure/sonic-buildimage/issues/5680

There are platforms which don't have the database_global.json file and hence, change the error level from ERROR to INFO here. This is in sync with sonic-swsssdk-py/dbconnector implementation as well.

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -77,7 +77,7 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
 
     if (m_global_init)
     {
-        SWSS_LOG_ERROR("SonicDBConfig Global config is already initialized");
+        SWSS_LOG_INFO("SonicDBConfig Global config is already initialized");
         return;
     }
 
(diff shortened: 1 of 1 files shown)


common/dbconnector.cpp
```diff
@@ -77,7 +77,7 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
 
     if (m_global_init)
     {
-        SWSS_LOG_ERROR("SonicDBConfig Global config is already initialized");
+        SWSS_LOG_INFO("SonicDBConfig Global config is already initialized");
```
qiluo-msft, line 80: >SWSS_LOG_INFO [](start = 8, length = 13)

ERROR seems reasonable. If "there are platforms which don't have the database_global.json file and hence", why call this fuction?

judyjoseph: Here, it is that initializeGlobalConfig() happened to be invoked again. I felt it is not a real an error per say....it is just that  the global config is already initialized -- and need to inform the user on that.
In addition I felt adding platform conditional check at places where we invoke GlobalConfigInit might not be a good idea. I thought in single asic devices -- or devices where the global-database.config is not present it is a no-op. 

your comments please @qiluo-msft ...

judyjoseph:
Discussed with Qi, concluded on leaving it as Error so that user is aware of the result. If we keep it as info, the user might overlook the actual result. So closing this PR.

PR #406, opened 2020-11-02
Add lua script for redis multi keys api hmset and del

kcudnik:
When using SAI bulk in syncd, like create/remove route_entry, we want to have ability to update database as fast as possible, so best way would be to execute every mutable db command inside lua script to make things faster, rather than execute 1 by 1 operation which is done currently inside syncd

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -10,6 +10,7 @@
 
 #include "common/dbconnector.h"
 #include "common/redisreply.h"
+#include "common/redisapi.h"
 
 using json = nlohmann::json;
 using namespace std;
@@ -487,6 +488,9 @@ void DBConnector::select(DBConnector *db)
 
     RedisReply r(db, select, REDIS_REPLY_STATUS);
     r.checkStatusOK();
+
+    std::string luaScript = loadLuaScript("redis_multi.lua");
+    db->m_shaRedisMulti = loadRedisScript(db, luaScript);
 }
 
 DBConnector::DBConnector(const DBConnector &other)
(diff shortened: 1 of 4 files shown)


kcudnik:
retest vs please

kcudnik:
retest swss-common please

kcudnik:
retest this please

lguohan (approved):

common/dbconnector.cpp
```diff
@@ -776,3 +780,57 @@ int64_t DBConnector::publish(const string &channel, const string &message)
...
+    command.format(
+        "EVALSHA %s 1 %s %s",
+        m_shaRedisMulti.c_str(),
+        strJson.c_str(),
+        "mhset");
+
+    RedisReply r(this, command, REDIS_REPLY_NIL);
+}
+
+void DBConnector::hdel(const std::vector<std::string>& keys)
```
qiluo-msft, line 815: >hdel [](start = 18, length = 4)

We already has `hdel()` function which is similar to [https://redis.io/commands/hdel](https://redis.io/commands/hdel)
I believe you don't mean this. The function name is confusing.

common/dbconnector.cpp
```diff
@@ -487,6 +488,9 @@ void DBConnector::select(DBConnector *db)
 
     RedisReply r(db, select, REDIS_REPLY_STATUS);
     r.checkStatusOK();
+
+    std::string luaScript = loadLuaScript("redis_multi.lua");
```
qiluo-msft, line 492: >loadLuaScript [](start = 28, length = 13)

Not all DBConnector are using the new functions. Could you only load on demand?

common/dbconnector.cpp
kcudnik: ok, so my naming may be wrong here since this function is doing "del" for multiple hashes, do you want me to rename ?

common/dbconnector.cpp
kcudnik: i can load this on demand inside functions we use them

dzhangalibaba: @qiluo-msft  @kcudnik  I already handle it together with another similar changes, please help review, https://github.com/Azure/sonic-swss-common/pull/409

kcudnik: you beat me by  6 minutes: https://github.com/kcudnik/sonic-swss-common/commit/8afd6f6a55e10d11b87aed671e0badee13754ea9

qiluo-msft: Let's move the discussion to [https://github.com/Azure/sonic-swss-common/pull/409](https://github.com/Azure/sonic-swss-common/pull/409).
This one is resolved here.

---
In reply to: [517085706](https://github.com/Azure/sonic-swss-common/pull/406#discussion_r517085706) [](ancestors = 517085706)

PR #407, opened 2020-11-03
Revert SonicV2Connector to unblock test

qiluo-msft:
Due to https://github.com/Azure/sonic-buildimage/issues/5697
Temporarily revert SonicV2Connector feature and will revisit later.


lguohan:
what about other repo？like sairedis？ swss？

qiluo-msft:
I checked and there should be no impact to other repos.

PR #408, opened 2020-11-03
Fix: treat DBConnector timeout=0 as infinite timeout

qiluo-msft:
Fixed a bug introduced by https://github.com/Azure/sonic-swss-common/pull/387
Fixed https://github.com/Azure/sonic-buildimage/issues/5697


lguohan (approved):

qiluo-msft:
Tested with buildimage PR image on Mellanox DUT.

PR #301, opened 2019-09-05
multi-DB Part 2: C++ interface API changes / swsscommon unit test / LOGLEVEL_DB apply new API

dzhangalibaba:
* add a new RedisConfig class which has static member and func, to parse/store/update/get database info from database_config.json file.

* add new DBConnector() constructor without port and unixsocket parameters. using RedisConfig class get func to get those info dynamically, all other DBConnector logic are same.

* the previous static sha member in tables are not good for multi-db cases, since on each db instance, there should be a SHA value, we cannot reuse only one SHA value, hence remove  the static keyword and keep SHA value as a member for each DBConnector object

* add updateDBConfigFile() to support vs test/ swsscommon test cases where this is no database_config.json at init time, we can point to load the config file.

* adding swsscommon unit test to check the  updateDBConfigFile() func, also check RedisConfig get APIs could return correct db info.

* relpace LOGLEVEL_DB with new DBConnector construct, LOGLEVEL_DB only has three places to changes, they are also in swsscommon repo, hence change here.  Since the database_config.json has only one redis inst for now. LOGLEVEL_DB behaves the same as before , though it is using new DBConnector constructor. All the other DBConnector replacement will be done in other PR.

* swsscommon test result ：
dzhang@343ca72c4bf5:/sonic/src/sonic-swss-common/tests$ ./tests --gtest_filter=DBConnector.multi_db_test
Running main() from gtest_main.cc
Note: Google Test filter = DBConnector.multi_db_test
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from DBConnector
[ RUN      ] DBConnector.multi_db_test
Default : load init db config file, isInit = 0
Update : load local db config file, isInit = 1
JSON file # dbid: 0 ,socket: /var/run/redis/redis.sock .port: 6379
API get   # dbid: 0 ,socket: /var/run/redis/redis.sock .port: 6379
JSON file # dbid: 1 ,socket: /var/run/redis/redis1.sock .port: 6380
API get   # dbid: 1 ,socket: /var/run/redis/redis1.sock .port: 6380
JSON file # dbid: 4 ,socket: /var/run/redis/redis2.sock .port: 6381
API get   # dbid: 4 ,socket: /var/run/redis/redis2.sock .port: 6381
JSON file # dbid: 2 ,socket: /var/run/redis/redis1.sock .port: 6380
API get   # dbid: 2 ,socket: /var/run/redis/redis1.sock .port: 6380
[...]

PR #409, opened 2020-11-04
[lua] load lua script on demand

dzhangalibaba:
* object load lua script when using it and only load once
* handle late comment on https://github.com/Azure/sonic-swss-common/pull/301#pullrequestreview-523030087 and https://github.com/Azure/sonic-swss-common/pull/406#pullrequestreview-523029861

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -489,15 +489,13 @@ void DBConnector::select(DBConnector *db)
 
     RedisReply r(db, select, REDIS_REPLY_STATUS);
     r.checkStatusOK();
-
-    std::string luaScript = loadLuaScript("redis_multi.lua");
-    db->m_shaRedisMulti = loadRedisScript(db, luaScript);
 }
 
 DBConnector::DBConnector(const DBConnector &other)
     : RedisContext(other)
     , m_dbId(other.m_dbId)
     , m_namespace(other.m_namespace)
+    , m_shaRedisMulti(string())
 {
     select(this);
 }
@@ -506,6 +504,7 @@ DBConnector::DBConnector(int dbId, const RedisContext& ctx)
     : RedisContext(ctx)
     , m_dbId(dbId)
     , m_namespace(EMPTY_NAMESPACE)
+    , m_shaRedisMulti(string())
 {
     select(this);
 }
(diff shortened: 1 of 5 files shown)


dzhangalibaba:
@qiluo-msft

kcudnik:
take a look here: https://github.com/kcudnik/sonic-swss-common/commit/8afd6f6a55e10d11b87aed671e0badee13754ea9

common/dbconnector.cpp
```diff
@@ -489,15 +489,13 @@ void DBConnector::select(DBConnector *db)
...
-
-    std::string luaScript = loadLuaScript("redis_multi.lua");
-    db->m_shaRedisMulti = loadRedisScript(db, luaScript);
 }
 
 DBConnector::DBConnector(const DBConnector &other)
     : RedisContext(other)
     , m_dbId(other.m_dbId)
     , m_namespace(other.m_namespace)
+    , m_shaRedisMulti(string())
```
kcudnik, line 498: this is not needed, string default constructor is empty

common/dbconnector.cpp
```diff
@@ -841,3 +845,12 @@ void DBConnector::hdel(const std::vector<std::string>& keys)
 
     RedisReply r(this, command, REDIS_REPLY_NIL);
 }
+
+void DBConnector::loadLua(string luaPath)
```
kcudnik, line 849: name seems this function is generic, but internally it will always populate m_shaRedisMulti so rather name should be loadRedisMultiLuaScript etc ?

qiluo-msft: In order to be generic, suggest add m_shaRedisMulti as the other argument. Suggest the Function name as `lazyLoadRedisScriptFile`.
Event better, it could be extracted out of this class, `common/redisapi.h` is a better place.

---
In reply to: [517163546](https://github.com/Azure/sonic-swss-common/pull/409#discussion_r517163546) [](ancestors = 517163546)

common/dbconnector.cpp
dzhangalibaba: updated

common/dbconnector.cpp
dzhangalibaba: updated

qiluo-msft (dismissed):

common/redisapi.h
```diff
@@ -140,4 +140,13 @@ static inline int peekRedisContext(redisContext *c)
     return rc;
 }
 
+static inline void lazyLoadRedisScriptFile(RedisContext* ctx, std::string luaPath, std::string &m_sha)
```
kcudnik, line 143: parameter naming: m_sha is not class memeber field, should be just "sha"

dzhangalibaba: Updated

kcudnik (approved):

qiluo-msft (approved):

PR #412, opened 2020-11-07
Reduce notice logging in loadRedisScript()

qiluo-msft:
Fixed https://github.com/Azure/sonic-buildimage/issues/5854

### DIFF
--- a/common/redisapi.h
+++ b/common/redisapi.h
@@ -27,8 +27,7 @@ static inline std::string loadRedisScript(RedisContext* ctx, const std::string& 
 
     std::string sha = r.getReply<std::string>();
 
-    SWSS_LOG_NOTICE("lua script loaded, sha: %s", sha.c_str());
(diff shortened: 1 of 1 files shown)


kcudnik (approved):

common/redisapi.h
```diff
@@ -27,8 +27,7 @@ static inline std::string loadRedisScript(RedisContext* ctx, const std::string&
 
     std::string sha = r.getReply<std::string>();
 
-    SWSS_LOG_NOTICE("lua script loaded, sha: %s", sha.c_str());
-    SWSS_LOG_INFO("lua script: %s", script.c_str());
+    SWSS_LOG_INFO("lua script %s loaded, sha: %s", script.c_str(), sha.c_str());
```
lguohan, line 30: should this be debug level?

qiluo-msft: It contains useful information, so I prefer keeping it as INFO level.

lguohan (approved):

PR #413, opened 2020-11-12
Fix: SWIG dict.get() should have optional default value parameter

qiluo-msft:


lguohan:
feature or bug fix?

qiluo-msft:
It's a bug fix.

shi-su (approved):

PR #416, opened 2020-11-18
[swig] translate C++ `del` to python `delete`

qiluo-msft:
Otherwise SWIG will "smartly" translate it to `_del`
```
Warning 314: 'del' is a python keyword, renaming to '_del'
```

Since some application is using old function name, I keep some functions backward compatible.


stephenxs (dismissed):
LGTM.

jleveque (approved):

PR #417, opened 2020-11-19
Table to store fdb restore table for fdbsyncd

kishorekunal01:
[FDBSYNCD] Added support for FDBSYNCD for EVPN as described in the PR Azure/SONiC#437

Table to restore the fdb entry during warm boot, This change is required to compile fdbsyncd changes
https://github.com/Azure/sonic-swss/pull/1276

What I did
Initial infra to support FDBSYNCD warm reboot

Why I did it
Part of overall EVPN-VXLAN Implementation

How I verified it
Not yet compiled or verified

Details if related
HLD Location : https://github.com/Azure/SONiC/pull/437/commits

Signed-off-by: Kishore Kunal kishore.kunal@broadcom.com


kishorekunal01:
It is not required, hence removing it

PR #418, opened 2020-11-21
[swig] wrap logger.h

qiluo-msft:
Related to https://github.com/Azure/sonic-buildimage/issues/5992
It turns out that we need to set log level in snmpagent repo.

### DIFF
--- a/common/logger.h
+++ b/common/logger.h
@@ -37,6 +37,11 @@ void err_exit(const char *fn, int ln, int e, const char *fmt, ...)
         err_exit(__FUNCTION__, __LINE__, e, (fmt), ##args); \
     }
 
+#ifdef __GNUC__
+#    define ATTRIBUTE_NORTEURN __attribute__ ((noreturn))
+#else
+#    define ATTRIBUTE_NORTEURN
+#endif
 
 class Logger
 {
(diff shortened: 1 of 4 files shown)


common/logger.h
```diff
@@ -130,7 +130,11 @@ class Logger
 
     static void swssPrioNotify(const std::string &component, const std::string &prioStr);
     static void swssOutputNotify(const std::string &component, const std::string &outputStr);
-    [[ noreturn ]] void settingThread();
+
+#ifndef SWIG
+    [[ noreturn ]]
+#endif
+    void settingThread();
```
jleveque, line 137: This is a bit hard to read. Is there a way we could create a macro, e.g.,

```c++
#ifdef SWIG
#    define ATTRIBUTE_NORTEURN
#else
#    define ATTRIBUTE_NORTEURN [[ noreturn ]]
#endif
```

or

```c++
#ifdef SWIG
#    define ATTRIBUTE_NORTEURN
#else
#    define ATTRIBUTE_NORTEURN __declspec(noreturn)
#endif
```

Then use:

```c++
ATTRIBUTE_NORTEURN void settingThread();
```

qiluo-msft: Thanks for the feedback! Actually I found similar code in this file and followed. And positively testing GNU make more sense. Do you read better?

---
In reply to: [528070062](https://github.com/Azure/sonic-swss-common/pull/418#discussion_r528070062) [](ancestors = 528070062)

jleveque: Sorry, but it's not much better. I was hoping to separate the `#ifdef` to a macro that can be reused so that this can become a one-line solution, e.g., 

```c++
ATTRIBUTE_NORTEURN void settingThread();
```

kcudnik (approved):

jleveque (approved):

PR #421, opened 2020-11-24
[swss-common] Add LINKMGR CFG and MUX LINKMGR state table names

tahmed-dev:
Adding LINKMGR cfg table name to config db macros and
MUX_LINKMGR state table to state db macros.
    
signed-off-by: Tamer Ahmed <tamer.ahmed@microsoft.com>

### DIFF
--- a/common/schema.h
+++ b/common/schema.h
@@ -356,7 +356,6 @@ namespace swss {
 
 #define STATE_MUX_CABLE_TABLE_NAME                  "MUX_CABLE_TABLE"
 #define STATE_HW_MUX_CABLE_TABLE_NAME               "HW_MUX_CABLE_TABLE"
-
 #define STATE_MUX_LINKMGR_TABLE_NAME                "MUX_LINKMGR_TABLE"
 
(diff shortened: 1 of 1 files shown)


lguohan:
not sure why we need this table? isn't in metadata|localhost?

common/schema.h
```diff
@@ -356,6 +357,8 @@ namespace swss {
 #define STATE_MUX_CABLE_TABLE_NAME                  "MUX_CABLE_TABLE"
 #define STATE_HW_MUX_CABLE_TABLE_NAME               "HW_MUX_CABLE_TABLE"
 
+#define STATE_MUX_LINKMGR_TABLE_NAME                "MUX_LINKMGR_TABLE"
```
jleveque, line 360: Why not group with the other mux-related table names above above as you did on line 306?

jleveque (approved):

common/schema.h
tahmed-dev: OK, will do

prsunny (approved):

PR #423, opened 2020-12-02
[dbconnector] protect db information with mutex

yxieca:
STL containers are only thread safe for reads. But not for writes. protect the db information variables with rwlock.

Signed-off-by: Ying Xie <ying.xie@microsoft.com>

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -82,6 +82,7 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
         return;
     }
 
+    pthread_rwlock_wrlock(&db_info_lock);
     ifstream i(file);
     if (i.good())
     {
@@ -141,11 +142,13 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
         catch (domain_error& e)
         {
             SWSS_LOG_ERROR("key doesn't exist in json object, NULL value has no iterator >> %s\n", e.what());
+            pthread_rwlock_unlock(&db_info_lock);
             throw runtime_error("key doesn't exist in json object, NULL value has no iterator >> " + string(e.what()));
         }
         catch (exception &e)
         {
             SWSS_LOG_ERROR("Sonic database config file syntax error >> %s\n", e.what());
+            pthread_rwlock_unlock(&db_info_lock);
             throw runtime_error("Sonic database config file syntax error >> " + string(e.what()));
         }
     }
@@ -153,6 +156,8 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
     {
         SWSS_LOG_ERROR("Sonic database config global file doesn't exist at %s\n", file.c_str());
     }
+
+    pthread_rwlock_unlock(&db_info_lock);
 
     // Set it as the global config file is already parsed and init done.
     m_global_init = true;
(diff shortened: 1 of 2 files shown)


yxieca:
This change is intended to address issue https://github.com/Azure/sonic-buildimage/issues/6103.

However, after added this tentative protection, I realized that there is no code removing anything from m_db_info. So I guess this protection won't really protect us against the issue above?

common/dbconnector.h
```diff
@@ -54,6 +55,7 @@ class SonicDBConfig
     static bool isGlobalInit() { return m_global_init; };
 
 private:
+    static pthread_rwlock_t db_info_lock;
```
kcudnik, line 58: make this static seems like bad idea to me

kcudnik:
@yxieca  What is main motivation for this change? You are manually locking/unlocking lock in many places and it could be easy to forget to do that when code will grow. At least wrapping that to class and release in destructor. Also is **std::shared_lock** what you are looking for?
I would be against using rwlock lock here, unless you really know what you are doing and you know what your flow/data is. I'm assuming that you are worrying that some of the threads will be starved to death? Do you have any data on which you are basing this proposal?

also mistake in title "dbconnectgor"

common/dbconnector.h
yxieca: All the information under protection are static variables. That is why this lock is also static.

kcudnik:
Is issue related to m_inst_info/m_db_info ?  those are static map, and in general it's advised to not uses statics, since they "wired" things can happen process is shutdown, depends on complexity, and threads running, it could happen that some destroyed data is still in use which can lead to some really hard to track errors

kcudnik:
This SonicDBConfig class don't see to me as critical performance class, so std::recursive_mutex would make more sense and more clean solution.

common/dbconnector.h
kcudnik: yea, entire class is static, which is bad XD

qiluo-msft:
> void SonicDBConfig::initializeGlobalConfig(const string &file)

All the `initialize...` functions should be run in single thread environment. Then all other functions are read-only.
Seems an application usage bug if the assumption is broken. Why application use multiple threads for any write operation?

---
Refers to: common/dbconnector.cpp:70 in 74cf24e. [](commit_id = 74cf24ec38964dcef03bf18c7aefed2cb6f59af9, deletion_comment = False)

common/dbconnector.cpp
```diff
@@ -141,11 +142,13 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
         catch (domain_error& e)
         {
             SWSS_LOG_ERROR("key doesn't exist in json object, NULL value has no iterator >> %s\n", e.what());
+            pthread_rwlock_unlock(&db_info_lock);
```
qiluo-msft, line 145: >pthread_rwlock_unlock [](start = 12, length = 21)

Better to use `std::lock_guard`

yxieca: Great idea. If we decide to move forward. I'll make change accordingly

kcudnik: yea, and use recursive lock as discussed

common/dbconnector.h
```diff
@@ -6,6 +6,8 @@
 #include <unordered_map>
 #include <utility>
 #include <memory>
+#include <pthread.h>
```
kcudnik, line 9: you dont need probably this include

kcudnik (approved):

common/dbconnector.h
yxieca: yes. I added that earlier. it is no longer needed now.

kcudnik (approved):

kcudnik (approved):

qiluo-msft (approved):

PR #425, opened 2020-12-07
Add fdb_flush.v2.lua script to makefile

kcudnik:


common/Makefile.am
```diff
@@ -8,7 +8,8 @@ EXTRA_DIST = \
     producer_state_table_apply_view.lua \
     table_dump.lua \
     redis_multi.lua \
-    fdb_flush.lua
+    fdb_flush.lua \
+    fdb_flush.v2.lua
```
qiluo-msft, line 12: Is it better to move application specific lua script into application repo (sairedis). Since it is not shared, you may simply remove v1 version.

kcudnik: v1 can be remove after we will shift everything and submodules on buldimage repo

kcudnik:
ping

lguohan (approved):

PR #427, opened 2020-12-12
Implement DBConnector::pubsub() to emulate python Redis.pubsub()

qiluo-msft:
This function will be SWIG wrapped and used by sonic-snmpagent repo.
Tested in unit test and manually on DUT.


common/pubsub.cpp
```diff
@@ -0,0 +1,125 @@
...
+PubSub::PubSub(DBConnector *parent)
+    : m_parentConnector(parent)
+{
+}
+
+void PubSub::psubscribe(const std::string &pattern)
+{
+    if (m_subscribe)
+    {
+        m_select.removeSelectable(this);
```
kcudnik, line 18: why on subscribe you remove selectable? should you add? should this variable be called m_subscribed ?

kcudnik (approved):

common/pubsub.cpp
```diff
@@ -0,0 +1,125 @@
...
+            m_keyspace_event_buffer.emplace_back(make_shared<RedisReply>(reply));
+        }
+    }
+    while(reply != nullptr && status == REDIS_OK);
+
+    if (status != REDIS_OK)
+    {
+        throw RedisError("Unable to redisGetReplyFromReader", m_subscribe->getContext());
+    }
+    return 0;
```
shi-su, line 61: Is there any meaning of the return value of this function?

qiluo-msft: My understanding is no. It is expected used only in Select class, and the return value is ignored there.

---
In reply to: [541802204](https://github.com/Azure/sonic-swss-common/pull/427#discussion_r541802204) [](ancestors = 541802204)

shi-su (approved):

PR #430, opened 2020-12-14
Remove extension of tableNameSeparatorMap

qiluo-msft:
Because it is used only in pyext for backward compatible. New code should get separator from `database_config.json`
Fixed the first issue in https://github.com/Azure/sonic-buildimage/issues/5351

### DIFF
--- a/common/table.cpp
+++ b/common/table.cpp
@@ -26,13 +26,7 @@ const TableNameSeparatorMap TableBase::tableNameSeparatorMap = {
    { CONFIG_DB,           TABLE_NAME_SEPARATOR_VBAR  },
    { PFC_WD_DB,           TABLE_NAME_SEPARATOR_COLON },
    { FLEX_COUNTER_DB,     TABLE_NAME_SEPARATOR_COLON },
-   { STATE_DB,            TABLE_NAME_SEPARATOR_VBAR  },
-   { RESTAPI_DB,          TABLE_NAME_SEPARATOR_VBAR  },
-   { GB_ASIC_DB,          TABLE_NAME_SEPARATOR_VBAR  },
-   { GB_COUNTERS_DB,      TABLE_NAME_SEPARATOR_VBAR  },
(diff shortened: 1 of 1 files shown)


common/table.cpp
```diff
@@ -26,13 +26,7 @@ const TableNameSeparatorMap TableBase::tableNameSeparatorMap = {
...
    { PFC_WD_DB,           TABLE_NAME_SEPARATOR_COLON },
    { FLEX_COUNTER_DB,     TABLE_NAME_SEPARATOR_COLON },
-   { STATE_DB,            TABLE_NAME_SEPARATOR_VBAR  },
-   { RESTAPI_DB,          TABLE_NAME_SEPARATOR_VBAR  },
-   { GB_ASIC_DB,          TABLE_NAME_SEPARATOR_VBAR  },
-   { GB_COUNTERS_DB,      TABLE_NAME_SEPARATOR_VBAR  },
-   { GB_FLEX_COUNTER_DB,  TABLE_NAME_SEPARATOR_VBAR  },
-   { CHASSIS_APP_DB,      TABLE_NAME_SEPARATOR_VBAR  },
-   { CHASSIS_STATE_DB,    TABLE_NAME_SEPARATOR_VBAR  }
+   { STATE_DB,            TABLE_NAME_SEPARATOR_VBAR  }
```
jleveque, line 29: Suggest also adding a `NOTE` comment above this map definition stating that it is deprecated and will be removed in the future, and state that new DBs should instead be added to `database_config.json`. This will help prevent future issues/confusion.

jleveque (changes requested):

jleveque (approved):

judyjoseph:
@Qi to confirm that we have moved completely from pyext .. as I see PFC_WD_DB and FLEX_COUNTER_DB still use ":" separator. https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-database/database_config.json.j2#L42

PR #432, opened 2020-12-16
Add support for hexists call

tonytitus:
Add support for hexists call.

Tested using below code snippet:

```
from swsscommon.swsscommon import SonicV2Connector

db = SonicV2Connector(host='127.0.0.1')
db.connect(db.COUNTERS_DB)
db.connect(db.ASIC_DB)

ex = int(db.hexists(db.COUNTERS_DB, "COUNTERS_PORT_NAME_MAP", "Ethernet0"))
if ex == 1:
    print "Exists"
else:
    print "Do not exists"
```

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -619,7 +619,7 @@ bool DBConnector::exists(const string &key)
     }
     rexists.format("EXISTS %s", key.c_str());
     RedisReply r(this, rexists, REDIS_REPLY_INTEGER);
-    return (r.getContext()->integer > 0);
+    return r.getContext()->integer > 0;
 }
 
 int64_t DBConnector::hdel(const string &key, const string &field)
@@ -652,6 +652,7 @@ void DBConnector::set(const string &key, const string &value)
     RedisReply r(this, sset, REDIS_REPLY_STATUS);
 }
 
+2020-05-29 23:41:31.177615 IFCS_LOG_INFO [NODE] [ifcs_init] Initializing IFCS, boot type = 0
 void DBConnector::config_set(const std::string &key, const std::string &value)
 {
     RedisCommand sset;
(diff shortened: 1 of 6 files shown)


common/dbconnector.cpp
```diff
@@ -740,6 +740,14 @@ shared_ptr<string> DBConnector::hget(const string &key, const string &field)
     throw runtime_error("HGET failed, unexpected reply type, memory exception");
 }
 
+bool DBConnector::hexists(const string &key, const string &field)
```
kcudnik, line 743: is this like "DBConnector::hget" ? you could check pointer if its populated, also if field exists, then you probably want to use it's value ?

is this for performance or what is use case?

tonytitus: Yes, you are right. It can be used to check if a field exists in a table before accessing it. Similar to "exists" for keys.

kcudnik (dismissed):

common/dbconnector.cpp
```diff
@@ -740,6 +740,14 @@ shared_ptr<string> DBConnector::hget(const string &key, const string &field)
...
+{
+    RedisCommand rexists;
+    rexists.format("HEXISTS %s %s", key.c_str(), field.c_str());
+    RedisReply r(this, rexists, REDIS_REPLY_INTEGER);
+    return (r.getContext()->integer > 0);
```
qiluo-msft, line 748: >( [](start = 11, length = 1)

Remove the parentheses?

common/dbconnector.cpp
```diff
@@ -652,6 +652,7 @@ void DBConnector::set(const string &key, const string &value)
     RedisReply r(this, sset, REDIS_REPLY_STATUS);
 }
 
+2020-05-29 23:41:31.177615 IFCS_LOG_INFO [NODE] [ifcs_init] Initializing IFCS, boot type = 0
```
qiluo-msft, line 655: Remove this line?

qiluo-msft:
A general recommendation: try not to squash commits when you iterate, we have internal tool to compare between commits so we could quickly identify the iteration improvement.

common/dbconnector.cpp
tonytitus: Done.

qiluo-msft (approved):

PR #434, opened 2020-12-22
Add utility for string and redis

Pterosaur:
This PR provides the following utilities

## AlphaBool

This is a class to provide a unified bool string

```c++
#include <swss/common/boolean.h>

swss::AlphaBool bt(true);
swss::AlphaBool bf(false);

std::cout << bt << std::endl; // "true"
std::cout << bf << std::endl; // "false"
```

## lexical_convert

This function can be used to convert the key in redis to variables with expected type

```c++
#include <swss/common/stringutility.h>

void doTask(const std::string &key, ...)
{
    // Assume the key is an applicant port "ETHERNET0:192.168.0.1:8080"
    std::string interface_name;
    IpAddress ip; // Need to define operator<<(ostream, IpAddress &)
    std::uint16 port;
    auto vs = tokenize(key, ':');
    lexical_convert(vs, interface_name, ip, port);
}
```

## join

This function can be used to generate key for redis by some different material

```c++
#include <swss/common/stringutility.h>

void updateItem(
    const std::string &interface_name,
    const IpAddress &ip, // Need to define operator>>(istream, IpAddress &)
    std::uint16 port,
    const std::vector<FieldValueTuple> &attrs)
{
    const std::string key = join(interface_name, ip, port);
    m_table.set(key, attrs);
}
```

## hex_to_binary

Generate a segment binary according to a hex string

```c++
#include <swss/common/stringutility.h>

struct CipherKey {
    ...
};

bool CreateCipherKey(const std::string &hex_str, CipherKey &key)
{
    // hex_str is "010203FF"
    // key will be 0x010203FF in memory
    return hex_to_binary(hex_str, &key, sizeof(key));
}
```

## binary_to_hex

Serialize a segment binary to a hex string

```c++
#include <swss/common/stringutility.h>

struct CipherKey {
    ...
};

std::string GetCipherKey(CipherKey &key)
{
    // key is 0x010203FF in memory
    // hex_str will be "010203FF"
    return binary_to_hex(&key, sizeof(key));
}
```

## fvtGetValue

This function provides a method to easily get the value from `std::vector<FieldValueTuple>` which is typical exchange format of redis in SONiC

```c++
#include <swss/common/redisutility.h>

void doTask(const std::string &key, std::vector<FieldValueTuple> &attrs)
{
    // Assume attrs includes
    // port_id = 0
    // port_name = Ethernet0
    // is_enable = true

    boost::optional<std::string> t1;
    boost::optional<std::string> t2;
    boost::optional<std::string> t3;
    boost::optional<std::string> t4;

    int port_id;
    std::string port_name;
    swss::AlphaBool is_enable;
    double duration_time;

    if ((t1 = swss::fvtGetValue(attrs, "port_id"))
        && (t2 = swss::fvtGetValue(attrs, "Port_Name", swss::FieldNameStrategy::CASE_INSENSITIVE))
        && (t3 = swss::fvtGetValue(attrs, "is_enable")))
    {
        swss::lexical_convert({t1, t2, t3}, port_id, port_name, is_enable);
        // Do Task
    }
    if (t4 = fvtGetValue(attrs, "duration_time"))
    {
        swss::lexical_convert(t4, duration_time);
    }
    else
    {
        // Error, duration time isn't in attrs
    }
}
```

Depends on: https://github.com/Azure/sonic-swss-common/pull/435

Signed-off-by: Ze Gan <ganze718@gmail.com>


caizhenghui-juniper (approved):

common/stringutility.h
```diff
@@ -0,0 +1,75 @@
+#ifndef __STRINGUTILITY__
```
qiluo-msft, line 1: >__STRINGUTILITY__ [](start = 8, length = 17)

The use of two underscores (`__') in identifiers is reserved for the compiler's internal use according to the ANSI-C standard.
Suggest use `#pragma once`

common/stringutility.h
```diff
@@ -0,0 +1,75 @@
...
+
+template <typename T>
+bool split(const std::string &input, char delimiter, T &output)
+{
+    SWSS_LOG_ENTER();
+    if (input.find(delimiter) != std::string::npos)
+    {
+        return false;
+    }
+    std::istringstream istream(input);
```
qiluo-msft, line 20: >istream [](start = 23, length = 7)

How is delimiter used in the input stream?

common/stringutility.h
```diff
@@ -0,0 +1,75 @@
...
+    {
+        return false;
+    }
+    std::istringstream istream(input.substr(0, pos));
+    istream >> output;
+    return split(input.substr(pos + 1, input.length() - pos - 1), delimiter, args...);
+}
+
+template <typename T>
+std::string join(char delimiter, const T &input)
```
qiluo-msft, line 40: >delimiter [](start = 22, length = 9)

Not used?

common/stringutility.h
```diff
@@ -0,0 +1,75 @@
...
+#include "logger.h"
+
+#include <string>
+#include <sstream>
+#include <cctype>
+
+namespace swss {
+
+template <typename T>
+bool split(const std::string &input, char delimiter, T &output)
```
qiluo-msft, line 13: >split [](start = 5, length = 5)

Did you try boost split and join functions?

common/stringutility.cpp
```diff
@@ -0,0 +1,97 @@
+#include "stringutility.h"
+
+#include <inttypes.h>
+#include <algorithm>
+
+std::istringstream &swss::operator>>(
+    std::istringstream &istream,
+    bool &b)
+{
+    std::string buffer = istream.str();
```
qiluo-msft, line 10: >str [](start = 33, length = 3)

Assume istream is extremely long, calling `str()` is not efficient.

common/stringutility.cpp
```diff
@@ -0,0 +1,97 @@
...
+    std::transform(
+        buffer.begin(),
+        buffer.end(),
+        buffer.begin(),
+        ::tolower);
+    if (buffer == "true" || buffer == "1")
```
qiluo-msft, line 16: >buffer [](start = 8, length = 6)

You did not consume the string in istream

common/stringutility.cpp
```diff
@@ -0,0 +1,97 @@
...
+    }
+    else
+    {
+        throw std::invalid_argument("Invalid bool string : " + buffer);
+    }
+
+    return istream;
+}
+
+bool swss::hex_to_binary(
```
qiluo-msft, line 32: >hex_to_binary [](start = 11, length = 13)

It would be easier if you use boost `unhex` or `stringstream` + `std::hex`

common/stringutility.cpp
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    s.resize(2 * length, '0');
+
+    const unsigned char *input = static_cast<const unsigned char *>(buffer);
+    char *output = &s[0];
+
+    for (size_t i = 0; i < length; i++)
+    {
+        snprintf(&output[i * 2], 3, "%02X", input[i]);
```
qiluo-msft, line 92: >output [](start = 18, length = 6)

You will write one more `\0` into `s`.

common/stringutility.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+        output[buffer_cur] = static_cast<unsigned char>(value);
+        hex_cur += 2;
+        buffer_cur += 1;
+    }
+
+    return true;
+}
+
+std::string swss::binary_to_hex(const void *buffer, size_t length)
```
qiluo-msft, line 74: >binary_to_hex [](start = 18, length = 13)

It would be easier if you use boost `hex` or `stringstream` + `std::hex`

common/stringutility.cpp
```diff
@@ -0,0 +1,97 @@
...
+    char *output = &s[0];
+
+    for (size_t i = 0; i < length; i++)
+    {
+        snprintf(&output[i * 2], 3, "%02X", input[i]);
+    }
+
+    return s;
+}
+
```
qiluo-msft, line 97: Remove extra blank line.

common/redisutility.h
```diff
@@ -0,0 +1,55 @@
...
+#include "stringutility.h"
+
+#include <vector>
+#include <algorithm>
+
+namespace swss
+{
+
+template <class T>
+static bool get_value(
```
qiluo-msft, line 15: >get_value [](start = 12, length = 9)

Better name: fvGetValue

common/redisutility.h
```diff
@@ -0,0 +1,55 @@
...
+    T &value)
+{
+    SWSS_LOG_ENTER();
+
+    std::string target_field = field;
+    std::transform(
+        target_field.begin(),
+        target_field.end(),
+        target_field.begin(),
+        ::tolower);
```
qiluo-msft, line 27: >tolower [](start = 10, length = 7)

What is the use case of case-insensitive comparison?

common/rediscommand.h
```diff
@@ -2,6 +2,7 @@
 #include <string.h>
 #include <utility>
 #include <tuple>
+#include <vector>
```
qiluo-msft, line 5: >vector [](start = 10, length = 6)

This is a valid and obvious fix. You may submit a small standalone PR for it.

qiluo-msft (changes requested):
As comments

common/stringutility.h
Pterosaur: This function is as a terminator to `bool split(const std::string &input, char delimiter, T &output, Args &... args)`, so I assume the last part of input string should not include any delimiter.

common/stringutility.h
Pterosaur: Yes, it isn't used. It's a terminator of `std::string join(char delimiter, const T &input, const Args &... args)`. So this parameters is to adapt the recursive style.
I will add qualifier, `__attribute__((unused))`, before this parameter delimiter.

common/stringutility.h
Pterosaur: I want to directly get some variables with expected type from the input string. I think boost split can just help us to split string into vector<string>.Meanwhile  I know we have had a helper `tokenize` can split string like boost split.
So what's your suggestion about replacing my code to boost split or tokenize and then convert vector<string> to expected type? or just keep it as it is.

common/redisutility.h
Pterosaur: I'm not sure, according to my understanding, if the field name is case-insensitive it will be better to use.
Do you have any concern.

common/rediscommand.h
Pterosaur: https://github.com/Azure/sonic-swss-common/pull/435

common/stringutility.h
qiluo-msft: Suggest you reuse existing function or common libraries (stl/boost/etc) first. If you find anything missing, try improve it based on existing ones by wrapping or preprocessing.
The last resort is to implement yourself.

---
In reply to: [547464854](https://github.com/Azure/sonic-swss-common/pull/434#discussion_r547464854) [](ancestors = 547464854)

common/stringutility.h
Pterosaur: Done

common/stringutility.h
Pterosaur: Done

common/stringutility.cpp
Pterosaur: Done

common/stringutility.cpp
Pterosaur: Done

common/stringutility.cpp
Pterosaur: Done

common/stringutility.cpp
Pterosaur: Done

common/stringutility.cpp
Pterosaur: Done

common/stringutility.cpp
Pterosaur: Done

common/redisutility.h
Pterosaur: Done

common/stringutility.h
qiluo-msft: Sorry for misleading. I mean the caller could just use `tokenize` or any common library function instead of your `split`.
They are roughly the same, I don't see significant benefit added in the `split`.

---
In reply to: [547616749](https://github.com/Azure/sonic-swss-common/pull/434#discussion_r547616749) [](ancestors = 547616749,547464854)

common/stringutility.h
```diff
@@ -0,0 +1,102 @@
...
+    T &output,
+    Args &... args)
+{
+    SWSS_LOG_ENTER();
+    if (begin_itr == end_itr)
+    {
+        return false;
+    }
+    std::istringstream istream(*(begin_itr++));
+    istream >> output;
```
qiluo-msft, line 56: I see you want to support any type `T`. However the construction of istringstream seems inefficient if you have a long vector. Possible to optimize for each type, like `stol` for `long`.
What is the use case? And how critical the new function required?

common/stringutility.cpp
```diff
@@ -0,0 +1,113 @@
...
+{
+    SWSS_LOG_ENTER();
+
+    std::string s;
+    auto _buffer = static_cast<const std::uint8_t *>(buffer);
+
+    boost::algorithm::hex(
+        _buffer,
+        _buffer + length,
+        std::insert_iterator<std::string>(s, s.end()));
```
qiluo-msft, line 110: >std::insert_iterator<std::string>(s, s.end()) [](start = 8, length = 45)

Better to use `std::back_inserter(s)`

common/stringutility.cpp
```diff
@@ -0,0 +1,113 @@
...
+        SWSS_LOG_ERROR("Buffer length isn't sufficient");
+        return false;
+    }
+
+    try
+    {
+        boost::algorithm::unhex(hex_str, buffer);
+        return true;
+    }
+    catch(const boost::algorithm::not_enough_input &e)
```
qiluo-msft, line 88: >boost::algorithm::not_enough_input &e) [](start = 16, length = 38)

It's more efficient to check even number before `unhex`, so no need to catch.

common/stringutility.h
qiluo-msft: Got it. Then better to remove the variable name and just use `char`. No need to add attribute.

---
In reply to: [547451799](https://github.com/Azure/sonic-swss-common/pull/434#discussion_r547451799) [](ancestors = 547451799)

common/stringutility.h
```diff
@@ -0,0 +1,102 @@
...
+
+template <typename T, typename... Args>
+bool split(
+    std::vector<std::string>::iterator begin_itr,
+    std::vector<std::string>::iterator end_itr,
+    T &output,
+    Args &... args)
+{
+    SWSS_LOG_ENTER();
+    if (begin_itr == end_itr)
```
qiluo-msft, line 51: >begin_itr [](start = 8, length = 9)

`begin` `end` are more readable?

common/stringutility.cpp
```diff
@@ -0,0 +1,113 @@
...
+}
+
+bool swss::hex_to_binary(
+    const std::string &hex_str,
+    std::uint8_t *buffer,
+    size_t buffer_length)
+{
+    SWSS_LOG_ENTER();
+
+    if (hex_str.length() > (buffer_length * 2))
```
qiluo-msft, line 77: >> [](start = 25, length = 1)

The use case for `>` is a surprise to me. How is the use case like? Is it better to return false?

common/stringutility.h
Pterosaur: I want to a helper function that I can directly get the variable with expected type. If the caller use tokenize or split in boost, I have to explicitly convert the string to target type at many positions.

common/stringutility.h
Pterosaur: Agree. I replace istringstream by boost lexical_cast

common/stringutility.cpp
Pterosaur: Done

common/stringutility.h
Pterosaur: Done

common/stringutility.cpp
Pterosaur: How about to change `>` to `!=`?

prsunny:
@Pterosaur , could you please provide some examples in description on usage of the utility?

Pterosaur:
> @Pterosaur , could you please provide some examples in description on usage of the utility?

Add some examples in the description of this PR, please check it.

common/redisutility.h
qiluo-msft: So better not to support unused feature and treat comparison case-sensitive.

---
In reply to: [547486695](https://github.com/Azure/sonic-swss-common/pull/434#discussion_r547486695) [](ancestors = 547486695)

common/stringutility.h
```diff
@@ -0,0 +1,112 @@
...
+template <typename T, typename... Args>
+bool split(const std::string &input, char delimiter, T &output, Args &... args)
+{
+    SWSS_LOG_ENTER();
+    auto tokens = tokenize(input, delimiter);
+    return split(tokens.begin(), tokens.end(), output, args...);
+}
+
+template <typename T>
+std::string join(char , const T &input)
```
qiluo-msft, line 79: >  [](start = 21, length = 1)

Remove the blank after char.

common/stringutility.cpp
qiluo-msft: That looks better.

---
In reply to: [547789357](https://github.com/Azure/sonic-swss-common/pull/434#discussion_r547789357) [](ancestors = 547789357)

common/redisutility.h
qiluo-msft: If you want to support both mode:
1. make case-sensitive an option parameter, default to sensitive.
2. don't use tolower in comparison, there may be some std library
3. try make code readable.

---
In reply to: [548387904](https://github.com/Azure/sonic-swss-common/pull/434#discussion_r548387904) [](ancestors = 548387904,547486695)

common/redisutility.h
```diff
@@ -0,0 +1,51 @@
...
+#include "stringutility.h"
+
+#include <vector>
+#include <algorithm>
+
+namespace swss
+{
+
+template <class T>
+static bool fvtGetValue(
```
qiluo-msft, line 14: >fvtGetValue [](start = 12, length = 11)

Better to have 2 functions: getValue extract a string, then another function to cast.

common/stringutility.h
```diff
@@ -0,0 +1,112 @@
...
+#include <vector>
+#include <string>
+#include <sstream>
+#include <cctype>
+#include <type_traits>
+
+namespace swss {
+
+template<typename T>
+T lexical_cast(const std::string &str)
```
qiluo-msft, line 18: >lexical_cast [](start = 2, length = 12)

`cast`

common/stringutility.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+#include <boost/algorithm/hex.hpp>
+
+#include <inttypes.h>
+#include <algorithm>
+#include <iterator>
+
+std::istringstream &swss::operator>>(
+    std::istringstream &istream,
+    bool &b)
```
qiluo-msft, line 11: This function is not very useful because we don't have a flexible schema definition. If it is ambiguous in schema, we need to fix schema itself.

common/stringutility.h
```diff
@@ -0,0 +1,112 @@
...
+{
+    return boost::lexical_cast<T>(str);
+}
+
+template<>
+bool lexical_cast<bool>(const std::string &str);
+
+template <typename T>
+bool split(const std::string &input, char delimiter, T &output)
```
qiluo-msft, line 27: >T &output) [](start = 53, length = 10)

Discussed offline, we need to separate split strings function and cast tuple function.

common/stringutility.h
```diff
@@ -0,0 +1,112 @@
...
+    ostream << input;
+    return ostream.str();
+}
+
+template <typename T, typename... Args>
+std::string join(char delimiter, const T &input, const Args &... args)
+{
+    SWSS_LOG_ENTER();
+    std::ostringstream ostream;
+    ostream << input << delimiter << join(delimiter, args...);
```
qiluo-msft, line 92: >ostream [](start = 4, length = 7)

Optimize ostream in recursive.

common/stringutility.cpp
```diff
@@ -0,0 +1,123 @@
...
+}
+
+bool swss::hex_to_binary(
+    const std::string &hex_str,
+    std::uint8_t *buffer,
+    size_t buffer_length)
+{
+    SWSS_LOG_ENTER();
+
+    if (hex_str.length() % 2 != 0)
```
qiluo-msft, line 86: No need to check even now.

common/stringutility.cpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    try
+    {
+        boost::algorithm::unhex(hex_str, buffer);
+        return true;
+    }
+    catch(const boost::algorithm::non_hex_input &e)
+    {
+        SWSS_LOG_ERROR("Invalid hex string %s", hex_str.c_str());
```
qiluo-msft, line 104: >SWSS_LOG_ERROR [](start = 8, length = 14)

DEBUG

common/stringutility.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    return false;
+}
+
+std::string swss::binary_to_hex(const void *buffer, size_t length)
+{
+    SWSS_LOG_ENTER();
+
+    std::string s;
+    auto _buffer = static_cast<const std::uint8_t *>(buffer);
```
qiluo-msft, line 115: >_buffer [](start = 9, length = 7)

buf

common/redisutility.h
Pterosaur: Done

common/stringutility.h
Pterosaur: Done

common/stringutility.cpp
Pterosaur: Done

common/stringutility.cpp
Pterosaur: Done

common/stringutility.cpp
Pterosaur: Done

common/stringutility.h
Pterosaur: Done

common/stringutility.h
Pterosaur: Done

common/stringutility.cpp
Pterosaur: Done

common/stringutility.h
Pterosaur: Done

common/redisutility.h
Pterosaur: Done

tests/redisutility_ut.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include "gtest/gtest.h"
+
+TEST(REDISUTILITY, fvtGetValue)
+{
+    std::vector<swss::FieldValueTuple> fvt;
+    fvt.push_back(std::make_pair("int", "123"));
+    fvt.push_back(std::make_pair("bool", "true"));
+    fvt.push_back(std::make_pair("string", "name"));
+
+    auto si = swss::fvtGetValue(fvt, "int");
```
qiluo-msft, line 12: >fvt [](start = 20, length = 3)

`fvs` is for a vector of tuples. So it is a better name.
`fvt` is for a tuple.

common/redisutility.h
```diff
@@ -0,0 +1,53 @@
...
+enum FieldNameStrategy
+{
+    CASE_SENSITIVE,
+    CASE_INSENSITIVE,
+};
+
+static inline boost::optional<std::string> fvtGetValue(
+    const std::vector<FieldValueTuple> &fvt,
+    const std::string &field,
+    FieldNameStrategy strategy = FieldNameStrategy::CASE_SENSITIVE)
```
qiluo-msft, line 25: >FieldNameStrategy [](start = 4, length = 17)

Since there are only 2 options, use a bool?

common/redisutility.h
```diff
@@ -0,0 +1,53 @@
...
+{
+    SWSS_LOG_ENTER();
+
+    for (auto itr = fvt.begin(); itr != fvt.end(); itr++)
+    {
+        bool is_equal = false;
+        switch(strategy)
+        {
+            case FieldNameStrategy::CASE_SENSITIVE:
+                is_equal = boost::equals(fvField(*itr), field);
```
qiluo-msft, line 35: >boost::equals [](start = 27, length = 13)

Use `==` for better readability?

common/redisutility.h
```diff
@@ -0,0 +1,53 @@
...
+        }
+        if (is_equal)
+        {
+            return boost::optional<std::string>(fvValue(*itr));
+        }
+    }
+
+    SWSS_LOG_DEBUG("Cannot find field : %s", field.c_str());
+
+    return boost::optional<std::string>();
```
qiluo-msft, line 51: >boost::optional<std::string>() [](start = 11, length = 30)

`boost::none`

common/redisutility.h
```diff
@@ -0,0 +1,53 @@
...
+            default:
+                is_equal = boost::equals(fvField(*itr), field);
+        }
+        if (is_equal)
+        {
+            return boost::optional<std::string>(fvValue(*itr));
+        }
+    }
+
+    SWSS_LOG_DEBUG("Cannot find field : %s", field.c_str());
```
qiluo-msft, line 49: >SWSS_LOG_DEBUG [](start = 4, length = 14)

This is obvious from the return value, no need to logging anything inside this function, including above `SWSS_LOG_ENTER`

common/stringutility.h
```diff
@@ -0,0 +1,164 @@
...
+#include <string>
+#include <sstream>
+#include <cctype>
+
+namespace swss {
+
+template<typename T>
+void cast(const std::string &str, T &t)
+{
+    SWSS_LOG_ENTER();
```
qiluo-msft, line 23: >SWSS_LOG_ENTER [](start = 4, length = 14)

Suggest no log simple functions.

common/stringutility.h
```diff
@@ -0,0 +1,164 @@
...
+#include <iterator>
+#include <vector>
+#include <string>
+#include <sstream>
+#include <cctype>
+
+namespace swss {
+
+template<typename T>
+void cast(const std::string &str, T &t)
```
qiluo-msft, line 21: >cast [](start = 5, length = 4)

`static inline`

tests/redisutility_ut.cpp
```diff
@@ -0,0 +1,30 @@
...
+    EXPECT_TRUE(si);
+    auto sb = swss::fvtGetValue(fvt, "bool");
+    EXPECT_TRUE(sb);
+    auto ss = swss::fvtGetValue(fvt, "string");
+    EXPECT_TRUE(ss);
+
+    int i;
+    bool b;
+    std::string s;
+    ASSERT_NO_THROW(swss::cast({si.get(), sb.get(), ss.get()}, i, b, s));
```
qiluo-msft, line 22: >si.get(), sb.get(), ss.get() [](start = 32, length = 28)

`*si, *sb, *ss` will be shorter.

common/schema.h
```diff
@@ -376,6 +376,9 @@ namespace swss {
 #define CFG_DTEL_QUEUE_REPORT_TABLE_NAME	"DTEL_QUEUE_REPORT"
 #define CFG_DTEL_EVENT_TABLE_NAME			"DTEL_EVENT"
 
+#define TRUE_STRING "true"
```
qiluo-msft, line 379: >"true" [](start = 20, length = 6)

Align with above lines

common/stringutility.h
```diff
@@ -0,0 +1,164 @@
...
+        }
+        else
+        {
+            throw e;
+        }
+    }
+}
+
+template <typename T, typename... Args>
+void cast(
```
qiluo-msft, line 52: >cast [](start = 5, length = 4)

These functions are not expected to be used by applications directly. So suggest move them into another namespace `swss::xxxx_details::`

common/stringutility.h
```diff
@@ -0,0 +1,164 @@
...
+    if (begin == end)
+    {
+        SWSS_LOG_THROW("Insufficient corpus");
+    }
+    cast(*(begin++), t);
+    return cast(begin, end, args...);
+}
+
+template <typename T, typename... Args>
+void cast(const std::vector<std::string> &strs, T &t, Args &... args)
```
qiluo-msft, line 87: >cast [](start = 5, length = 4)

After 2nd thought, I see `cast` is too ambiguous. I propose the name `swss::lexical_convert`.

common/stringutility.h
```diff
@@ -0,0 +1,164 @@
...
+{
+    cast(strs.begin(), strs.end(), t, args...);
+}
+
+template <typename T>
+void join(std::ostringstream &ostream, char, const T &t)
```
qiluo-msft, line 93: >join [](start = 5, length = 4)

These functions are not expected to be used by applications directly. So suggest move them into another namespace swss::xxxx_details::

common/stringutility.h
```diff
@@ -0,0 +1,164 @@
...
+        return false;
+    }
+
+    return true;
+}
+
+template<typename T>
+void hex_to_binary(
+    const std::string &s,
+    T &value)
```
qiluo-msft, line 142: Don't break lines for short parameter list.

common/redisutility.h
```diff
@@ -0,0 +1,53 @@
...
+namespace swss
+{
+
+enum FieldNameStrategy
+{
+    CASE_SENSITIVE,
+    CASE_INSENSITIVE,
+};
+
+static inline boost::optional<std::string> fvtGetValue(
```
qiluo-msft, line 22: >inline [](start = 7, length = 6)

Don't inline non-trivial functions. Separate into cpp will make dependent project easy to build.

qiluo-msft: Then make the include headers as minimal as possible. Only for function declaration purpose, not for function implementation.

---
In reply to: [548762018](https://github.com/Azure/sonic-swss-common/pull/434#discussion_r548762018) [](ancestors = 548762018)

common/stringutility.h
```diff
@@ -0,0 +1,164 @@
...
+namespace swss {
+
+template<typename T>
+void cast(const std::string &str, T &t)
+{
+    SWSS_LOG_ENTER();
+    t = boost::lexical_cast<T>(str);
+}
+
+static inline void cast(const std::string &str, bool &b)
```
qiluo-msft, line 27: >cast [](start = 19, length = 4)

If it is ambiguous in schema, maybe we can improve schema to use only 0/1.
Then we don't need this flexible bool casting.

Pterosaur: I think to support "true"/"false" in schema is useful because it can be consistent with SONiC doc, like https://github.com/Azure/SONiC/blob/b16791f3ca16398153556f40b11831cfbb6aba79/thermal-control-design.md.
image

If we will revise all existed bool to 0/1, otherwise, I think, it will be a little bit confuse to a new developer about how to declare a bool.

common/stringutility.h
Pterosaur: Agree

common/stringutility.h
qiluo-msft: Discussed offline. I anticipate a general solution will be:

1. support any customized pair, even like down/up -> 0/1
2. never mix pair, like up/enable

We could achieve this by new classes, which could be implicitly convert to bool. They are different only when serialize/deserialize.

---
In reply to: [548762444](https://github.com/Azure/sonic-swss-common/pull/434#discussion_r548762444) [](ancestors = 548762444)

Pterosaur: Done

common/stringutility.h
```diff
@@ -0,0 +1,159 @@
...
+//         else if (str == FALSE_STRING)
+//         {
+//             b = false;
+//         }
+//         else
+//         {
+//             throw e;
+//         }
+//     }
+// }
```
qiluo-msft, line 47: Remove unused code

common/boolean.h
```diff
@@ -0,0 +1,30 @@
...
+protected:
+    Boolean() = default;
+    Boolean(bool boolean);
+    bool m_boolean;
+};
+
+class AlphaBoolean : public Boolean
+{
+public:
+    static const std::string ALPHA_TRUE;
```
qiluo-msft, line 20: >ALPHA_TRUE [](start = 29, length = 10)

TRUE_STRING

common/boolean.h
```diff
@@ -0,0 +1,30 @@
...
+    static const std::string ALPHA_FALSE;
```
qiluo-msft, line 21: >ALPHA_FALSE [](start = 29, length = 11)

FALSE_STRING

common/boolean.cpp
```diff
@@ -0,0 +1,68 @@
...
+swss::Boolean::Boolean(bool boolean) : m_boolean(boolean)
+{
+}
+
+swss::AlphaBoolean::AlphaBoolean(bool boolean) : Boolean(boolean)
+{
+}
+
+const std::string swss::AlphaBoolean::ALPHA_TRUE = "true";
+const std::string swss::AlphaBoolean::ALPHA_FALSE = "false";
```
qiluo-msft, line 19: You may inline the super short functions and member constants, to it will be optimized out.

common/boolean.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+std::ostream &swss::operator<<(std::ostream &out, const swss::AlphaBoolean &b)
+{
+    out << (b ? swss::AlphaBoolean::ALPHA_TRUE : swss::AlphaBoolean::ALPHA_FALSE);
```
qiluo-msft, line 23: >swss::AlphaBoolean::ALPHA_TRUE [](start = 16, length = 30)

`b.TRUE_STRING`
If you improve the classes, you may write a generic `<<` for all `Boolean` and its subclasses.

common/stringutility.h
```diff
@@ -0,0 +1,159 @@
...
+
+namespace swss {
+
+template<typename T>
+static inline void lexical_convert(const std::string &str, T &t)
+{
+    t = boost::lexical_cast<T>(str);
+}
+
+// static inline void lexical_convert(const std::string &str, bool &b)
```
qiluo-msft, line 26: >lexical_convert [](start = 22, length = 15)

Do you need to implement lexical_convert for AlphaBoolean?
If there is a default one, is it implemented by stringstream? If yes, I think you can provide a simpler one by overload here.

common/boolean.cpp
```diff
@@ -0,0 +1,68 @@
...
+    return out;
+}
+
+std::istream &swss::operator>>(std::istream &in, swss::AlphaBoolean &b)
```
qiluo-msft, line 27: >>> [](start = 28, length = 2)

For this special class, you can use `in >> boolalpha >> b`

common/stringutility.h
Pterosaur: Yes, it will call the default one. But I cannot understand why we need a special overload lexical_convert for AlphaBoolean?

common/boolean.h
```diff
@@ -0,0 +1,52 @@
...
+namespace swss
+{
+class Boolean
+{
+public:
+    Boolean() = default;
+    Boolean(bool boolean) : m_boolean(boolean)
+    {
+    }
+    operator bool() const
```
qiluo-msft, line 15: >operator bool() [](start = 4, length = 15)

You need another cast operator
```
    operator bool&()
    {
        return m_boolean;
    }
```

qiluo-msft (approved):

PR #437, opened 2020-12-30
Implement ConfigDBConnector and ConfigDBPipeConnector in C++

qiluo-msft:
They are originally implemented by python in https://github.com/Azure/sonic-py-swsssdk/
We plan to deprecated the python version which based on redis-py.
The new implementation is based on hiredis C++ library.

Tested manually with sonic-config-engine unit tests.

### DIFF
--- a/common/dbconnector.h
+++ b/common/dbconnector.h
@@ -223,9 +223,9 @@ private:
 template<typename OutputIterator>
 void DBConnector::hgetall(const std::string &key, OutputIterator result)
 {
-    RedisCommand sincr;
-    sincr.format("HGETALL %s", key.c_str());
-    RedisReply r(this, sincr, REDIS_REPLY_ARRAY);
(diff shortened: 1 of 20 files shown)


lguohan (approved):

common/dbconnector.h
```diff
@@ -168,13 +178,15 @@ class DBConnector : public RedisContext
 
     void del(const std::vector<std::string>& keys);
 
-    std::unordered_map<std::string, std::string> hgetall(const std::string &key);
+    std::map<std::string, std::string> hgetall(const std::string &key);
```
kcudnik, line 181: why there is change from unordered_map to map? map is slower nlogn to sort items, and we dont need sort at hgetall, it will slow things down when we will have 100k entreis in db, that will matter, please reverse that to unordered_map

PR #440, opened 2021-01-04
Change DBConnector::hgetall return type from map to unordered_map

qiluo-msft:
Because of performance issue.

### DIFF
--- a/common/configdb.cpp
+++ b/common/configdb.cpp
@@ -125,7 +125,9 @@ map<string, string> ConfigDBConnector::get_entry(string table, string key)
 {
     auto& client = get_redis_client(m_db_name);
     string _hash = to_upper(table) + TABLE_NAME_SEPARATOR + key;
-    return client.hgetall(_hash);
+    map<string, string> ret;
+    client.hgetall(_hash, inserter(ret, ret.end()));
+    return ret;
 }
 
 // Read all keys of a table from config db.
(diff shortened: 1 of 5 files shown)


kcudnik (dismissed):

common/configdb.cpp
```diff
@@ -84,7 +84,7 @@ void ConfigDBConnector::set_entry(string table, string key, const map<string, st
         for (auto& it: original)
         {
             auto& k = it.first;
-            bool found = data.find(k) == data.end();
+            bool found = data.find(k) != data.end();
```
kcudnik, line 87: nice bug

common/dbconnector.h
```diff
@@ -178,10 +177,13 @@ class DBConnector : public RedisContext
 
     void del(const std::vector<std::string>& keys);
 
-    std::map<std::string, std::string> hgetall(const std::string &key);
+    template <typename ReturnType=std::unordered_map<std::string, std::string>>
+    ReturnType hgetall(const std::string &key);
```
kcudnik, line 181: i think even if you have default template argument specified, then this will also required sairedis fix to function call like this: dbconnector->hgetall<>(key); // template brackets still must be present

kcudnik (approved):

common/dbconnector.h
qiluo-msft: I checked the test case [https://github.com/Azure/sonic-swss-common/blob/869848127cc2cca524b06708f1f3cc46b0d3bd0b/tests/redis_ut.cpp#L370](https://github.com/Azure/sonic-swss-common/blob/869848127cc2cca524b06708f1f3cc46b0d3bd0b/tests/redis_ut.cpp#L370). Seem the old code is still working.

---
In reply to: [552253246](https://github.com/Azure/sonic-swss-common/pull/440#discussion_r552253246) [](ancestors = 552253246)

PR #449, opened 2021-01-20
[packaging]: Add Support For Libboost v1.71.0

tahmed-dev:
This PR defines libboost v1.71.0 as an alternative package
for libswsscommon.

singed-off-by: Tamer Ahmed <tamer.ahmed@microsoft.com>

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: sonic
 Maintainer: Shuotian Cheng <shuche@microsoft.com>
 Section: net
 Priority: optional
-Build-Depends: dh-exec (>=0.3), debhelper (>= 9), autotools-dev, libboost-dev
+Build-Depends: dh-exec (>=0.3), debhelper (>= 9), autotools-dev, libboost-dev | libboost1.71-dev
 Standards-Version: 1.0.0
 
 Package: libswsscommon


qiluo-msft:
@Pterosaur please help review

qiluo-msft (approved):

debian/control
```diff
@@ -2,7 +2,7 @@ Source: sonic
 Maintainer: Shuotian Cheng <shuche@microsoft.com>
 Section: net
 Priority: optional
-Build-Depends: dh-exec (>=0.3), debhelper (>= 9), autotools-dev, libboost-dev
+Build-Depends: dh-exec (>=0.3), debhelper (>= 9), autotools-dev, libboost-dev | libboost1.71-dev
```
jleveque, line 5: Does order matter here? Does this guarantee that it will try to install `libboost1.71-dev` first? I can't seem to find info about the ordering.

qiluo-msft: Even it matters, we should not depends on the order and install explicitly the package in sonic-slave (buster and future).

jleveque: We're trying to eliminate explicit dependency installation as much as possible. If we need to here for the time being, OK, but we should try to avoid it if possible.

tahmed-dev: I manually install libboost1.71-dev on both slave and docker images. @qiluo-msft had similar question. The document does not provide enough details. The solution for now is to call out a specific version in the docker that uses it. The 'OR' is required for now as the package builds for both stretch and buster.

jleveque: I guess we'll have to experiment to see how the order impacts `apt-get install --fix-broken` in the future, since the documentation is poor, and see if we can eliminate the explicit install of `libboost1.71-dev`.

jleveque (approved):

PR #453, opened 2021-02-04
[swig] Add missing functions for ConfigDBConnector

qiluo-msft:
Add missing functions for ConfigDBConnector, including
- ConfigDBConnector.get_redis_client(...)
- ConfigDBConnector.CONFIG_DB
- ConfigDBConnector.get_redis_client(...).flushdb()
- ConfigDBConnector.INIT_INDICATOR
- ConfigDBConnector.db_connect(...): support optional parameters

Tested in unit test:
1. this repo: add unit testcases for all the new functions.
2. sonic-config-engine
3. sonic-utilities

Tested in DUT:
1. `sudo config load_minigraph -y`
2. snmpwalk snmpagent on 1.3.6.1.4.1.9.9.187

### DIFF
--- a/common/sonicv2connector.h
+++ b/common/sonicv2connector.h
@@ -63,7 +63,7 @@ private:
 %pythoncode %{
     class SonicV2Connector(SonicV2Connector_Native):
 
-        ## Note: there is no easy way for SWIG to map ctor parameter netns(C++) to namespace(python),
+        ## Note: there is no easy way for SWIG to map ctor parameter netns(C++) to namespace(python)
         def __init__(self, use_unix_socket_path = False, namespace = '', **kwargs):
             if 'host' in kwargs:
                 # Note: host argument will be ignored, same as in sonic-py-swsssdk
--- a/common/sonicv2connector.h
+++ b/common/sonicv2connector.h
@@ -8,20 +8,12 @@
 namespace swss
 {
 
-class SonicV2Connector
+class SonicV2Connector_Native
 {
 public:
-    SonicV2Connector(bool use_unix_socket_path = false, const char *netns = "");
+    SonicV2Connector_Native(bool use_unix_socket_path = false, const char *netns = "");
 
     std::string getNamespace() const;
-
-#ifdef SWIG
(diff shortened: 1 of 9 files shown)


kcudnik:
why change to native ?

qiluo-msft:
The change is to split the class SonicV2Connector into two classes SonicV2Connector_Native and SonicV2Connector.

SonicV2Connector is used by SWIG and then by python applications, or used by other C++ classes which are also used by SWIG and python application. Not for C++ application usage.

There are some functionalities are not easy/possible to purely implemented in C++, such as namespace property (it is a C++ keyword). So original I implement them in `pythoncode` block. Later I found these features are not available in C++ subclasses, and finally python subclasses. So I made the changes here, and let python subclass inherit like this:
- SonicV2Connector_Native (C++)
- SonicV2Connector (python): SonicV2Connector_Native
- ConfigDBConnector_Native (C++): SonicV2Connector_Native
- ConfigDBConnector (python): SonicV2Connector, ConfigDBConnector_Native

So the python subclass will inherit all the members from python base class.

---
In reply to: [773590454](https://github.com/Azure/sonic-swss-common/pull/453#issuecomment-773590454) [](ancestors = 773590454)

common/sonicv2connector.h
```diff
@@ -68,32 +60,32 @@ class SonicV2Connector
...
-            # set a database name as a constant value attribute.
-            setattr(self, db_name, db_name)
-            getmethod = lambda self: db_name
-            SonicV2Connector.__swig_getmethods__[db_name] = getmethod
-            SonicV2Connector.__swig_setmethods__[db_name] = None
-
-    SonicV2Connector.__init__ = _new_SonicV2Connector__init__
+    class SonicV2Connector(SonicV2Connector_Native):
+
+        ## Note: there is no easy way for SWIG to map ctor parameter netns(C++) to namespace(python),
```
jleveque, line 66: Looks like the second line of this comment is missing.

lguohan (dismissed):

jleveque (approved):

PR #455, opened 2021-02-06
Update to make getNamespaces() API at par with get_ns_list() in swsssdk

judyjoseph:
Update to make getNamespaces() API at par with the get_ns_list() in swsssdk. So this API 
1. Will give back all namespaces including the global namespace ('')
2. Will not preload the Global database_config.json. The application should call SonicDBConfig::initializeGlobalConfig explicity.


judyjoseph:
@Qi, @SuvarnaMeenakshi I don't see this API being used currently in code. So we should be ok to change it to make it at par with get_ns_list() in swsssdk.

qiluo-msft (approved):

PR #458, opened 2021-02-11
[debian/control] libswsscommon-dev depends on libboost-dev

stepanblyschak:
libswsscommon-dev is not usable without libboost-dev package, meaning I cannot compile with headers which include boost libraries. Make it implicitly install by apt-get install -f

Signed-off-by: Stepan Blyshchak <stepanb@nvidia.com>

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -19,7 +19,7 @@ Description: debugging symbols for libswsscommon library.
 
 Package: libswsscommon-dev
 Architecture: any
-Depends: libswsscommon (= ${binary:Version})
+Depends: libswsscommon (= ${binary:Version}), libboost-dev | libboost1.71-dev
 Section: libdevel
 Description: This package contains development files for Switch State Service.
 


liat-grozovik:
@qiluo-msft can you please review and add 202012 label req?

qiluo-msft:
@tahmed-dev please help review

tahmed-dev:
> @tahmed-dev please help review
Thanks @qiluo-msft and @stepanblyschak . Stepan, can you please shed light on the use case, here? libboost-dev is installed in the stretch/buster build dockers

tahmed-dev (approved):
LGTM. Just curious about the use case.

qiluo-msft (approved):

stepanblyschak:
@tahmed-dev @qiluo-msft The use case is to implicitelly install -dev dependencies in sonic-sdk-buildenv docker https://github.com/Azure/sonic-buildimage/pull/6712 instead of installing it explicitelly.

tahmed-dev:
> @tahmed-dev @qiluo-msft The use case is to implicitelly install -dev dependencies in sonic-sdk-buildenv docker [Azure/sonic-buildimage#6712](https://github.com/Azure/sonic-buildimage/pull/6712) instead of installing it explicitelly.

Curious to know which boost v is being installed in the sonic-sdk-buildenv. Which version gets pulled 1.67 or v1.71?

stepanblyschak:
@tahmed-dev 1.67 is getting installed. Which one is required? From what I saw in Build-Depends it can work with either of them

tahmed-dev:
> @tahmed-dev 1.67 is getting installed. Which one is required? From what I saw in Build-Depends it can work with either of them

We are running v1.71 which is backported to buster. However it doesn't resolve dependency on generic `libboost` and so we opted to install it explicitly for buster. swss common dependency is compile time AFAIK, however if it take dependency on specific version, it will pull it and we end up with two `libboost` versions. So, you might wanna install v 1.71 in you target to have `libboost` version aligned across different containers.

stepanblyschak:
> > @tahmed-dev 1.67 is getting installed. Which one is required? From what I saw in Build-Depends it can work with either of them
> 
> We are running v1.71 which is backported to buster. However it doesn't resolve dependency on generic `libboost` and so we opted to install it explicitly for buster. swss common dependency is compile time AFAIK, however if it take dependency on specific version, it will pull it and we end up with two `libboost` versions. So, you might wanna install v 1.71 in you target to have `libboost` version aligned across different containers.

@tahmed-dev If you want to give preference to 1.71 then we should specify "libboost1.71-dev | libboost-dev". With such reorder 1.71 got installed in sonic-sdk.

I would like to avoid explicit dependency installations in my SDK docker. Let me know what do you think

tahmed-dev:
> > > @tahmed-dev 1.67 is getting installed. Which one is required? From what I saw in Build-Depends it can work with either of them
> > 
> > 
> > We are running v1.71 which is backported to buster. However it doesn't resolve dependency on generic `libboost` and so we opted to install it explicitly for buster. swss common dependency is compile time AFAIK, however if it take dependency on specific version, it will pull it and we end up with two `libboost` versions. So, you might wanna install v 1.71 in you target to have `libboost` version aligned across different containers.
> 
> @tahmed-dev If you want to give preference to 1.71 then we should specify "libboost1.71-dev | libboost-dev". With such reorder 1.71 got installed in sonic-sdk.
> 
> I would like to avoid explicit dependency installations in my SDK docker. Let me know what do you think

Thanks @stepanblyschak ! That would work great!

PR #460, opened 2021-02-22
Add a function to SonicDBConfig to get DB instance list

vaibhavhd:
Add a new get function to obtain DB instances. This is to be used by some utility scripts.

Dependent PR in sonic-buildimage which will use this: https://github.com/Azure/sonic-buildimage/pull/6844


qiluo-msft (changes requested):
Please add unit test, and also test with KVM or DUT.

PR #462, opened 2021-02-24
Convert return value of get_all function in SonicV2Connector to dict

shi-su:
The function `get_all` in `SonicV2Connector` returns a FieldValueMap which is not json serializable. This object type will cause failures in fast reboot dump. Converting the object to a dictionary to avoid fast reboot failures.

### DIFF
--- a/common/sonicv2connector.h
+++ b/common/sonicv2connector.h
@@ -88,7 +88,7 @@ private:
             return self.getNamespace()
 
         def get_all(self, db_name, _hash, blocking=False):
-            return dict(super().get_all(db_name, _hash, blocking))
+            return dict(super(SonicV2Connector, self).get_all(db_name, _hash, blocking))
 %}
 #endif
 }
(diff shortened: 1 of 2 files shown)


shi-su:
retest swss-common please

shi-su:
retest this please

common/sonicv2connector.h
```diff
@@ -86,6 +86,9 @@ class SonicV2Connector_Native
         @property
         def namespace(self):
             return self.getNamespace()
+
+        def get_all(self, db_name, _hash, blocking=False):
```
qiluo-msft, line 90: It's a valid fix.

I am wondering if you can fix similar to https://github.com/Azure/sonic-swss-common/blob/b75aab7636021199d3f6498eb2295f8e9dc34d73/pyext/swsscommon.i#L80 so it will be a general solution for this type.

qiluo-msft (approved):

PR #463, opened 2021-03-03
[sonic-swss-common] Add a string utilities function to join items in a container

Junchao-Mellanox:
Why I did this?

Add a string utilities function to join items in a container

How I did this?

Add a template function to join all items in a container and return the joined string

How I verify it?

Manual test.


liat-grozovik:
@qiluo-msft  could you please review ?

common/stringutility.h
```diff
@@ -93,6 +93,24 @@ static inline std::string join(char delimiter, const T &t, const Args &... args)
     return ostream.str();
 }
 
+template <typename Container, typename Delim>
+static inline std::string join(const Delim &delim, const Container &cont)
```
qiluo-msft, line 97: Is it possible to unify the new `join` with existing ones?
1. delim -> delimiter
2. char delimiter
3. If this could be covered by L89 generic version, move this function into `join_detail`.

qiluo-msft: Alternatively, how about just using `boost::algorithm::join` ?

---
In reply to: [589165103](https://github.com/Azure/sonic-swss-common/pull/463#discussion_r589165103) [](ancestors = 589165103)

common/stringutility.h
```diff
@@ -133,4 +151,14 @@ static inline std::string binary_to_hex(const void *buffer, size_t length)
     return s;
 }
 
+static inline void toupper(std::string &str)
```
qiluo-msft, line 154: >toupper [](start = 19, length = 7)

We already have [https://github.com/Azure/sonic-swss-common/blob/f01fede8500a499ce94f3573f0bb6c3c5991efe7/common/converter.h#L66](https://github.com/Azure/sonic-swss-common/blob/f01fede8500a499ce94f3573f0bb6c3c5991efe7/common/converter.h#L66)
Can you use that one?

Junchao-Mellanox:
@qiluo-msft, thanks for your comments. I did not notice that orchagent has boost dependency. I will use boost join and boost to_lower. The PR is not necessary.

common/stringutility.h
Junchao-Mellanox: boost join only accept a container whose element is string, however, in my case, the element is integer. I will put the join implementation to join_detail. For the delimiter, i would like to keep it as a template parameter here because it support situation such as:

```cpp
swss::join("some delimiter", container);
swss::join(some_number, container); 
```

common/stringutility.h
Junchao-Mellanox: thanks for your comments. I did not notice that orchagent has boost dependency. I will use boost to_lower/to_upper.

common/stringutility.h
qiluo-msft: Is it a valid use case of  `swss::join(some_number, container);` ? If `some_number` and `container` items are all non-string, the `join` is weird in semantics.
Even this is a valid use case, just use `swss::join(to_string(some_number), container);`. The advantage is that it plays well together with `join(char delimiter, const T &t, const Args &... args)`

---
In reply to: [589165358](https://github.com/Azure/sonic-swss-common/pull/463#discussion_r589165358) [](ancestors = 589165358,589165103)

common/stringutility.h
```diff
@@ -93,6 +110,14 @@ static inline std::string join(char delimiter, const T &t, const Args &... args)
     return ostream.str();
 }
 
+template <typename Delimiter, typename Container>
```
qiluo-msft, line 113: >template [](start = 0, length = 8)

Please add unit test for new code.

qiluo-msft: Did you missing unit test in your latest iteration?

---
In reply to: [589711891](https://github.com/Azure/sonic-swss-common/pull/463#discussion_r589711891) [](ancestors = 589711891)

common/stringutility.h
```diff
@@ -83,6 +83,23 @@ namespace join_detail
...
+        auto iter = container.begin();
+        auto end = container.end();
+        ostream << *iter;
+        while(++iter != end)
+        {
+            ostream << delimiter << *iter;
+        }
+    }
+
 }
```
qiluo-msft, line 103: remove this extra empty line

common/stringutility.h
```diff
@@ -93,6 +110,14 @@ static inline std::string join(char delimiter, const T &t, const Args &... args)
     return ostream.str();
 }
 
+template <typename Container>
+static inline std::string join_n(char delimiter, const Container &container)
```
qiluo-msft, line 114: >join_n [](start = 26, length = 6)

I see you are not using the `join` function name. Is it due to some technical difficult?
I anticipate that your new function should be unified / intergrated with existing `join` function family. So you function is technically a specialization of
```
    template <typename T>
    void join(std::ostringstream &ostream, char, const T &t)
```

I found some examples with different mechanisms on [https://stackoverflow.com/a/23715236/2514803](https://stackoverflow.com/a/23715236/2514803)

Junchao-Mellanox: Yes, the technical difficult is that there is already a template signature defined in join_detail:

```cpp
template <typename T>
void join(std::ostringstream &ostream, char, const T &t)
```

And my change will also introduce a template to join_detail namespace:

```cpp
template <typename Container>
void join(std::ostringstream &ostream, char delimiter, const Container &container)
```

They have exactly the same template signature and cause compile failure. I read the link you sent, the idea is to do function/template specification for each type of container. It requires to do specification for every container and it does not support customize container. In that case, I would prefer to implement a range based join like below:

```cpp
template <typenmae Iterator>
void join(std::ostringstream &ostream, char delimiter, Iterator begin, Iterator end)
```

How does it sound?

qiluo-msft: It sounds a great and interesting idea.
You may meet new conflictions with 
```
    template <typename T, typename... Args>
    void join(std::ostringstream &ostream, char delimiter, const T &t, const Args &... args)
```
If you can conquer this issue, that would be a perfect integration with existing API design.

---
In reply to: [591846396](https://github.com/Azure/sonic-swss-common/pull/463#discussion_r591846396) [](ancestors = 591846396)

Junchao-Mellanox:
Hi @qiluo-msft , could you please kindly review again?

liat-grozovik:
@Junchao-Mellanox  could you please check the failed test?

qiluo-msft (approved):

liat-grozovik:
@daall looks like ACL IPv6 tests is not working and I don't think it is related to this PR. Can you please take further look?

Junchao-Mellanox:
Hi @qiluo-msft, i don't understand why the test case failed again and again. I just added a new function which should be only called in a unit test case.

liat-grozovik:
@qiluo-msft can we go a head and merge it?

PR #465, opened 2021-03-10
fix getting hash from redis db

dmytroxshevchuk:


common/configdb.cpp
```diff
@@ -248,12 +248,11 @@ map<string, map<string, map<string, string>>> ConfigDBConnector_Native::get_conf
     for (string key: keys)
     {
         size_t pos = key.find(TABLE_NAME_SEPARATOR);
-        string table_name = key.substr(0, pos);
-        string row;
-        if (pos != string::npos)
-        {
-            row = key.substr(pos + 1);
+        if (pos == string::npos) {
+            continue;
```
qiluo-msft, line 252: >continue [](start = 12, length = 8)

Thanks for the fix! Could you add a unit test?

qiluo-msft:
>         if (pos != string::npos)

This one is similar, however it is not likely to happen since  the pattern has separator.
Could you also `continue` if separator not found?

---
Refers to: common/configdb.cpp:183 in a499586. [](commit_id = a49958648bd34660684d3b7660ba56bd4f66f505, deletion_comment = False)

qiluo-msft:
>         if (pos != string::npos)

Could you also `continue` if not found?

---
Refers to: common/configdb.cpp:413 in a499586. [](commit_id = a49958648bd34660684d3b7660ba56bd4f66f505, deletion_comment = False)

common/configdb.cpp
mykolaxgerasymenko: @qiluo-msft,
I added unit test for get_config() and get_table() functions. Please review.

mykolaxgerasymenko:
> > ```
> >     if (pos != string::npos)
> > ```
> 
> This one is similar, however it is not likely to happen since the pattern has separator.
> Could you also `continue` if separator not found?
> 
> Refers to: common/configdb.cpp:183 in [a499586](https://github.com/Azure/sonic-swss-common/commit/a49958648bd34660684d3b7660ba56bd4f66f505). [](commit_id = [a499586](https://github.com/Azure/sonic-swss-common/commit/a49958648bd34660684d3b7660ba56bd4f66f505), deletion_comment = False)

@qiluo-msft , continue was added if separator not found for both cases. Please review.

qiluo-msft (approved):

mykolaxgerasymenko:
please restart

mykolaxgerasymenko:
@qiluo-msft, I pushed with force without changes to restart automatic checks and tests.

qiluo-msft (approved):

allas-nvidia:
@qiluo-msft
Can you please merge it to 202012 as well. Thanks

PR #466, opened 2021-03-11
[swss-common] Remove object type argument from json array() constructor call

rajkumar38:
Default value set to value_t::array in constructor.
Issue: https://github.com/Azure/sonic-sairedis/issues/801

Signed-off-by: Rajkumar Pennadam Ramamoorthy <rpennadamram@marvell.com>

### DIFF
--- a/common/json.hpp
+++ b/common/json.hpp
@@ -1612,7 +1612,7 @@ class basic_json
     static basic_json array(std::initializer_list<basic_json> init =
                                 std::initializer_list<basic_json>())
     {
-        return basic_json(init, false, value_t::array);
+        return basic_json(init, false);
     }
 
     /*!
--- a/tests/json_ut.cpp
+++ b/tests/json_ut.cpp
@@ -75,3 +75,12 @@ TEST(JSON, test)
     }
     file.close();
 }
+
+TEST(JSON, Array)
+{
+    json arr = json::array();
+    json::array_t a;
+    json arr1(a);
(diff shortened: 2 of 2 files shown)


rajkumar38:
@kcudnik . Please review.

common/json.hpp
```diff
@@ -1612,7 +1612,7 @@ class basic_json
     static basic_json array(std::initializer_list<basic_json> init =
                                 std::initializer_list<basic_json>())
     {
-        return basic_json(init, false, value_t::array);
+        return basic_json(init, false);
```
kcudnik, line 1615: this is interesting, since array type was explicitly passed as argument, can you also add unittest for that scenario that was causing crash ? in this repo

rajkumar38: Added an UT case in file, json_ut.cpp.

kcudnik (approved):

rajkumar38:
Pls hold on this PR. 
Even with this work-around, we are seeing same crash after several hours.

kcudnik:
> Pls hold on this PR.
> Even with this work-around, we are seeing same crash after several hours.

please diagnose and add those crashes to unittests here also when fixed

kcudnik:
this one also after verification, could be closed, since issue is most likely related to cross compiler issues on armhf

kcudnik:
closing, as this is no longer needed since issue 801 was resolved

PR #469, opened 2021-03-30
sonic-swss-common changes for MPLS support

qbdwlr:
Refer to HLD:  https://github.com/Azure/SONiC/pull/706

### DIFF
--- a/common/schema.h
+++ b/common/schema.h
@@ -34,6 +34,7 @@ namespace swss {
 #define APP_INTF_TABLE_NAME               "INTF_TABLE"
 #define APP_NEIGH_TABLE_NAME              "NEIGH_TABLE"
 #define APP_ROUTE_TABLE_NAME              "ROUTE_TABLE"
+#define APP_LABEL_ROUTE_TABLE_NAME        "LABEL_ROUTE_TABLE"
 #define APP_TUNNEL_DECAP_TABLE_NAME       "TUNNEL_DECAP_TABLE"
 #define APP_TUNNEL_ROUTE_TABLE_NAME       "TUNNEL_ROUTE_TABLE"
 #define APP_FDB_TABLE_NAME                "FDB_TABLE"


common/label.h
```diff
@@ -0,0 +1,77 @@
...
+            {
+                str += LABEL_DELIMITER;
+            }
+            str += std::to_string(*it);
+        }
+        return str;
+    }
+
+private:
+    std::set<Label> m_labelstack;
```
smaheshm, line 72: why set? Isn't there a need to maintain the order of "labels"?

common/label.h
```diff
@@ -0,0 +1,77 @@
...
+class LabelStack
+{
+public:
+    LabelStack() = default;
+    // A list of Labels separated by '/'
+    LabelStack(const std::string &str)
+    {
+        auto labels = swss::tokenize(str, LABEL_DELIMITER);
+        for (const auto &i : labels)
+            m_labelstack.insert(stoi(i));
```
smaheshm, line 24: check for valid label and throw an exception

common/label.h
```diff
@@ -0,0 +1,77 @@
...
+    {
+        return m_labelstack;
+    }
+
+    inline size_t getSize() const
+    {
+        return m_labelstack.size();
+    }
+
+    inline bool empty() const
```
smaheshm, line 37: when is this used? Wouldn't Label object be created only if there's a label

common/label.h
```diff
@@ -0,0 +1,77 @@
...
+        return m_labelstack.empty();
+    }
+
+    inline bool operator<(const LabelStack &o) const
+    {
+        return m_labelstack < o.m_labelstack;
+    }
+
+    inline bool operator==(const LabelStack &o) const
+    {
```
smaheshm, line 48: two label stacks with same labels and different sequence are not equal.

common/label.h
```diff
@@ -0,0 +1,77 @@
...
+typedef uint32_t Label;
+
+#define LABEL_DELIMITER '/'
+
+class LabelStack
+{
+public:
+    LabelStack() = default;
+    // A list of Labels separated by '/'
+    LabelStack(const std::string &str)
```
smaheshm, line 20: can also add the option of constructing using a label stack as std::vector<uint32_t>. This seems cleaner than providing a string of labels with "/" as a separater.

common/label.h
qbdwlr: No.  See changes in sonic-swss/src/nexthopkey.h.  LabelStack is always present in NextHopKey.  When NextHopKey represents IP-only NH, then the NextHopKey will contain an empty LabelStack.

common/label.h
qbdwlr: No.  LabelStack information comes from APPL_DB as a single string with token delimiters.

common/label.h
```diff
@@ -0,0 +1,77 @@
...
+#include <stdint.h>
+#include <vector>
+#include <string>
+#include "tokenize.h"
+
+namespace swss {
+
+typedef uint32_t Label;
+
+#define LABEL_DELIMITER '/'
```
lguohan, line 13: is this compliant with yang model?

common/label.h
qbdwlr: This is changed to vector<Label>.

common/label.h
qbdwlr: This is done, but I have moved label.h from sonic-swss-common into sonic-swss since it should be co-located with nexthopkey.h.  Please address further comments in https://github.com/Azure/sonic-swss/pull/1686

common/label.h
qbdwlr: label.h is moved to sonic-swss from sonic-swss-common.

qbdwlr:
@lguohan @smaheshm @prsunny Please review/approve this one-line sonic-swss-common PR, since the associated sonic-swss PR has build dependencies on it.

prsunny (approved):

PR #471, opened 2021-04-03
Make nlohmann::json namespace visibility hidden

kcudnik:
Potential fix for armhf architecture to make json symbols local

Original issue described here: https://github.com/Azure/sonic-sairedis/issues/801

### DIFF
--- a/common/json.hpp
+++ b/common/json.hpp
@@ -48,7 +48,7 @@ Licensed under the MIT License <http://opensource.org/licenses/MIT>.
 @see https://github.com/nlohmann
 @since version 1.0.0
 */
-namespace nlohmann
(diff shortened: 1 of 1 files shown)


kcudnik:
```
gearparserbase.h:33:7: error: 'GearParserBase' declared with greater visibility than the type of its field 'GearParserBase::m_root' [-Werror=attributes]
 class GearParserBase
```

since GearParserBase exposes json as return object, visibility is broken, and this workaround can't be used in this way, we could relax this by using pragmas or refactoring GearParseBase, but that's another workaround ...

kcudnik:
closing as this is no longer needed, was fixed in sonic build env, original issue: https://github.com/Azure/sonic-sairedis/issues/801

PR #472, opened 2021-04-06
[azp] Purge swss before installing the newly built deb package

daall:
Signed-off-by: Danny Allen <daall@microsoft.com>

### DIFF
--- a/.azure-pipelines/docker-sonic-vs/Dockerfile
+++ b/.azure-pipelines/docker-sonic-vs/Dockerfile
@@ -4,6 +4,15 @@ ARG docker_container_name
 
 ADD ["debs", "/debs"]
 
+RUN dpkg --purge python-swsscommon
+RUN dpkg --purge python3-swsscommon
+RUN dpkg --purge swss
+RUN dpkg --purge libsairedis
+RUN dpkg --purge libswsscommon
+RUN dpkg --purge libsaimetadata
+RUN dpkg --purge libsaivs
+RUN dpkg --purge syncd-vs
+
 RUN dpkg -i /debs/libswsscommon_1.0.0_amd64.deb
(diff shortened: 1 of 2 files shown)


.azure-pipelines/docker-sonic-vs/Dockerfile
```diff
@@ -13,4 +13,6 @@ RUN dpkg -i /debs/libsairedis_1.0.0_amd64.deb
 RUN dpkg -i /debs/libsaivs_1.0.0_amd64.deb
 RUN dpkg -i /debs/syncd-vs_1.0.0_amd64.deb
 
+RUN md5sum /usr/bin/orchagent
```
kcudnik, line 16: why this would be there if you install it in next step :P

qiluo-msft: Also interested in the theory. @daall do you have a good one?

daall: Honestly, I don't. 😆 I just know that in the past we have encountered issues with debs not being installed properly via the dockerfile so I thought this might be similar.

.azure-pipelines/docker-sonic-vs/Dockerfile
```diff
@@ -13,4 +13,5 @@ RUN dpkg -i /debs/libsairedis_1.0.0_amd64.deb
 RUN dpkg -i /debs/libsaivs_1.0.0_amd64.deb
 RUN dpkg -i /debs/syncd-vs_1.0.0_amd64.deb
 
+RUN dpkg --purge swss
```
lguohan, line 16: please purge every package as abishek did in sairedis pr.

qiluo-msft (approved):

qiluo-msft:
I think it is also required on 201911. @abdosi Please check.

PR #473, opened 2021-04-06
Modify the hardcode separator ":" to variable.

PJHsieh:
This variable comes from getTableNameSeparator() method.


qiluo-msft (approved):

PJHsieh:
retest

PR #477, opened 2021-04-16
[swig] allow threads

qiluo-msft:
By default, SWIG does not enable support for multithreaded Python applications. More specifically, the Python wrappers generated by SWIG will not release the Python's interpreter's Global Interpreter Lock (GIL) when wrapped C/C++ code is entered. Hence, while any of the wrapped C/C++ code is executing, the Python interpreter will not be able to run any other threads, even if the wrapped C/C++ code is waiting in a blocking call for something like network or disk IO.

ref: http://www.swig.org/Doc4.0/Python.html#Python_multithreaded
But the solution is for swig 4.0. The PR fix it in swig 3.0.

### DIFF
--- a/pyext/swsscommon.i
+++ b/pyext/swsscommon.i
@@ -46,6 +46,7 @@
 %include <exception.i>
 
 %exception {
+    Py_BEGIN_ALLOW_THREADS
     try
     {
         $action
@@ -55,6 +56,7 @@
     {
         SWIG_exception(SWIG_UnknownError, "unknown exception");
     }
+    Py_END_ALLOW_THREADS
 }
 
 %template(FieldValuePair) std::pair<std::string, std::string>;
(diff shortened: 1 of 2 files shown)


kcudnik:
some description why this is needed?

jleveque:
> some description why this is needed?

I'll let @qiluo-msft  update the PR description, but basically, if swsscommon is imported in a threaded Python application, blocking operations in swsscommon in any thread will cause the blockage of other threads, also. A main problem we have encountered is with `Selectable.select()`.

I found this reference, also: https://stackoverflow.com/questions/2510696/allowing-threads-from-python-after-calling-a-blocking-i-o-code-in-a-python-exten

jleveque:
@qiluo-msft: Question -- can we not simply compile with the `-threads` flag?

See http://www.swig.org/Doc4.0/Python.html#Python_multithreaded

qiluo-msft:
@jleveque I did not try this feature, and I notice it is for swig 4.0. We are using swig 3.0.

jleveque (approved):

jleveque (approved):

jleveque (approved):

PR #478, opened 2021-04-17
[swig] Implement TableEntryPoppable.pops()

qiluo-msft:
Due to the SWIG limitation on tuples, it's not straightforward to wrap the C++ `TableEntryPoppable::pops()`. Here I implement a workaround `TableEntryPoppable.pops()` which returns a tuple of 3 arrays, and another helper function to transpose the result to emulate the behavior of C++ `TableEntryPoppable::pops()`.

Tested in unit test.

### DIFF
--- a/pyext/swsscommon.i
+++ b/pyext/swsscommon.i
@@ -79,6 +79,15 @@
     }
 }
 
+%template(FieldValuePair) std::pair<std::string, std::string>;
+%template(FieldValuePairs) std::vector<std::pair<std::string, std::string>>;
+%template(FieldValuePairsList) std::vector<std::vector<std::pair<std::string, std::string>>>;
+%template(FieldValueMap) std::map<std::string, std::string>;
+%template(VectorString) std::vector<std::string>;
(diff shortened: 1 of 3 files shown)


pyext/swsscommon.i
```diff
@@ -124,11 +125,16 @@ T castSelectableObj(swss::Selectable *temp)
     %template(hgetall) hgetall<std::map<std::string, std::string>>;
 }
 
+%ignore swss::TableEntryPoppable::pops(std::deque<KeyOpFieldsValuesTuple> &, const std::string &);
 %apply std::vector<std::string>& OUTPUT {std::vector<std::string> &keys};
+%apply std::vector<std::string>& OUTPUT {std::vector<std::string> &ops};
```
tahmed-dev, line 130: am not a SWIG expert, however is there a way to concatenate the KFV vectors here and eliminate the use of transpose?

qiluo-msft: I could not figure out a solution here.

lguohan (approved):

PR #481, opened 2021-04-20
Add support for config_db subscribe and unsubscribe python apis

arlakshm:
Signed-off-by: Arvindsrinivasan Lakshmi Narasimhan <arlakshm@microsoft.com>

The PR https://github.com/Azure/sonic-buildimage/pull/6920 adds the support for swsscommon in `hostcfgd`. The PR test failure are seen because the `config_db.subscribe` doesnt work with swsscommon. This PR adds the support for the following python APIs
 - `config_db.subscribe`
 - `config_db.unsubscribe`

### DIFF
--- a/common/configdb.h
+++ b/common/configdb.h
@@ -27,18 +27,26 @@ public:
     virtual std::map<std::string, std::map<std::string, std::map<std::string, std::string>>> get_config();
 
     std::string getKeySeparator() const;
+    std::string getTableNameSeparator() const;
+    std::string getDbName() const;
 
 #ifdef SWIG
     %pythoncode %{
         __swig_getmethods__["KEY_SEPARATOR"] = getKeySeparator
         __swig_setmethods__["KEY_SEPARATOR"] = None
+        __swig_getmethods__["TABLE_NAME_SEPARATOR"] = getTableNameSeparator
+        __swig_setmethods__["TABLE_NAME_SEPARATOR"] = None
+        __swig_getmethods__["db_name"] = getDbName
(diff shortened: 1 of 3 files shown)


common/configdb.h
```diff
@@ -126,6 +134,14 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
             if table in self.handlers:
                 handler = self.handlers[table]
                 handler(table, key, data)
+        
+        def subscribe(self, table, handler):
```
qiluo-msft, line 138: >subscribe [](http://example.com/codeflow?start=12&length=9)

Could you add unit test?

arlakshm: added in latest commit

tests/test_redis_ut.py
```diff
@@ -437,3 +437,24 @@ def test_multidb_ConfigDBConnector():
     SonicDBConfig.load_sonic_global_db_config(global_db_config)
     config_db = ConfigDBConnector(use_unix_socket_path=True, namespace='asic1')
     assert config_db.namespace == 'asic1'
+
+def test_ConfigDBSubscribe():
+        table_name = 'Test_table'
+        test_key = 'key1'
+        test_data = {'field1': 'value1'}
+
+        def test_handler(key, data):
```
qiluo-msft, line 446: >test_handler [](http://example.com/codeflow?start=12&length=12)

Could you also test this handler is really called at least once?

arlakshm: Fixed the test case in the latest commit

qiluo-msft (approved):

PR #483, opened 2021-04-22
[swig] fix ConfigDBConnector.db_name

qiluo-msft:
Originally this property is not working.
Add unit test.

### DIFF
--- a/common/configdb.h
+++ b/common/configdb.h
@@ -27,11 +27,12 @@ public:
     virtual std::map<std::string, std::map<std::string, std::map<std::string, std::string>>> get_config();
 
     std::string getKeySeparator() const;
+    std::string getTableNameSeparator() const;
     std::string getDbName() const;
 
 protected:
-    std::string TABLE_NAME_SEPARATOR = "|";
-    std::string KEY_SEPARATOR = "|";
+    std::string m_table_name_separator = "|";
+    std::string m_key_separator = "|";
 
     std::string m_db_name;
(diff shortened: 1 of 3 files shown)


common/configdb.h
```diff
@@ -27,18 +27,49 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
     virtual std::map<std::string, std::map<std::string, std::map<std::string, std::string>>> get_config();
 
     std::string getKeySeparator() const;
+    std::string getDbName() const;
+
+protected:
+    std::string TABLE_NAME_SEPARATOR = "|";
```
arlakshm, line 33: do we need to add property for `TABLE_NAME_SEPARATOR` as well ?

kcudnik: why those 2 members are capita letters like defines ? use common naming convention

qiluo-msft: @kcudnik I am trying to mimic py-swsssdk https://github.com/Azure/sonic-py-swsssdk/blob/9d019b0f330710588f37d112104c563087dc9529/src/swsssdk/configdb.py#L55

qiluo-msft: @kcudnik Agree with you and fixed.

qiluo-msft: @arlakshm Agree and fixed.

arlakshm (approved):
lgtm

kcudnik (approved):

PR #487, opened 2021-06-06
[ci]: Fix azure pipeline template branch reference error

xumia:
Fix azure pipeline template branch reference error

PR #488, opened 2021-06-06
[ci]: Fix azure pipeline template branch reference error

xumia:
Fix azure pipeline template branch reference error

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,9 +12,11 @@ resources:
   repositories:
   - repository: sonic-sairedis
     type: github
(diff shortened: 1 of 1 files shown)


kcudnik (approved):

xumia:
Looks like the parameter in resource ref not supported.
"A template expression is not allowed in the context"
https://dev.azure.com/mssonic/build/_build/results?buildId=18398&view=results

xumia:
Use https://github.com/Azure/sonic-swss-common/pull/487 instead.

PR #489, opened 2021-06-06
[ci]: Support azure pipelines for 201911 branch

xumia:
Merge azure pipelines templates to 201911 branch
Fix the pipelines branch reference error, change from master to 201911
Fix sonic-swss method not implementation error,  we need to used the patched teamd from sonic-buildimage 201911.
Fix default python 3.8 error, for 201911, need to use python2.7.


kcudnik (approved):

.azure-pipelines/build-docker-sonic-vs-template.yml
```diff
@@ -0,0 +1,76 @@
+parameters:
```
qiluo-msft, line 1: Cherry-pick master branch templates to stable branches will not scale in future. Can we keep the main template on master, and reference it from stable branches?

xumia: It may be depended on the azure pipeline script, it the script there are lots of hard code installation package version, such as libpython2.7-dev, it may be changed in different branches.

qiluo-msft (approved):

qiluo-msft (approved):

PR #497, opened 2021-06-16
Fix the DBname in the database_config.json file.

judyjoseph:
This commit fixes the DB name that came repeated in database_config.json file and resulted in UT failure.


qiluo-msft:
I will cherry-pick https://github.com/Azure/sonic-swss-common/pull/337 instead of this PR.

PR #498, opened 2021-06-17
modify the logger for more flexable change of log level

emilmih:
add extra parameter  to  linkToDbNative , which is default to "NOTICE" to allow  daemons to set their own log level easily.
Example:
 Logger::linkToDbNative("portsyncd"); // by default "NOTICE"

 Logger::linkToDbNative("portsyncd", "INFO"); //explicit "INFO"

### DIFF
--- a/common/logger.h
+++ b/common/logger.h
@@ -79,7 +79,7 @@ public:
     static void linkToDbWithOutput(const std::string &dbName, const PriorityChangeNotify& prioNotify, const std::string& defPrio, const OutputChangeNotify& outputNotify, const std::string& defOutput);
     static void linkToDb(const std::string &dbName, const PriorityChangeNotify& notify, const std::string& defPrio);
     // Must be called after all linkToDb to start select from DB
-    static void linkToDbNative(const std::string &dbName);
+    static void linkToDbNative(const std::string &dbName, const std::string& defPrio="NOTICE");
     void write(Priority prio, const char *fmt, ...)
 #ifdef __GNUC__
         __attribute__ ((format (printf, 3, 4)))
--- a/common/logger.h
+++ b/common/logger.h
@@ -79,7 +79,7 @@ public:
     static void linkToDbWithOutput(const std::string &dbName, const PriorityChangeNotify& prioNotify, const std::string& defPrio, const OutputChangeNotify& outputNotify, const std::string& defOutput);
     static void linkToDb(const std::string &dbName, const PriorityChangeNotify& notify, const std::string& defPrio);
     // Must be called after all linkToDb to start select from DB
-    static void linkToDbNative(const std::string &dbName, const std::string& defPrio="NOTICE");
+    static void linkToDbNative(const std::string &dbName, const char * defPrio="NOTICE");
     void write(Priority prio, const char *fmt, ...)
 #ifdef __GNUC__
         __attribute__ ((format (printf, 3, 4)))
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -141,11 +141,11 @@ void Logger::linkToDb(const std::string &dbName, const PriorityChangeNotify& pri
     linkToDbWithOutput(dbName, prioNotify, defPrio, swssOutputNotify, "SYSLOG");
 }
 
-void Logger::linkToDbNative(const std::string &dbName)
+void Logger::linkToDbNative(const std::string &dbName, const std::string& defPrio)
 {
     auto& logger = getInstance();
 
-    linkToDb(dbName, swssPrioNotify, "NOTICE");
+    linkToDb(dbName, swssPrioNotify, defPrio);
     logger.m_settingThread.reset(new std::thread(&Logger::settingThread, &logger));
 }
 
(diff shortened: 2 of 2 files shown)


emilmih (commented):
add extra parameter to linkToDbNative , which is default to "NOTICE" to allow daemons to set their own log level easily.
Example:
Logger::linkToDbNative("portsyncd"); // by default "NOTICE"

Logger::linkToDbNative("portsyncd", "INFO"); //explicit "INFO"

emilmih:
add extra parameter to linkToDbNative , which is default to "NOTICE" to allow daemons to set their own log level easily.
Example:
Logger::linkToDbNative("portsyncd"); // by default "NOTICE"

Logger::linkToDbNative("portsyncd", "INFO"); //explicit "INFO"

common/logger.h
```diff
@@ -79,7 +79,7 @@ class Logger
     static void linkToDbWithOutput(const std::string &dbName, const PriorityChangeNotify& prioNotify, const std::string& defPrio, const OutputChangeNotify& outputNotify, const std::string& defOutput);
     static void linkToDb(const std::string &dbName, const PriorityChangeNotify& notify, const std::string& defPrio);
     // Must be called after all linkToDb to start select from DB
-    static void linkToDbNative(const std::string &dbName);
+    static void linkToDbNative(const std::string &dbName, const std::string& defPrio="NOTICE");
```
qiluo-msft, line 82: >defPrio [](http://example.com/codeflow?start=77&length=7)

One minor performance issue, the default usage without a defPrio argument, you will construct a new string object from const char array. Better to use `const char *` as the type.

emilmih: Thanks , I will fix it . Can you be my reviewer so i can push it after that

qiluo-msft: Others looks good to me. I can approve if you fix this one.

emilmih (commented):
Modify the logger for more flexable change of log level:

Fix performance issue, change the default defPrio argument to const char *

common/logger.h
kcudnik: @qiluo-msft seems like this is not performance issue, how many times linkToDbNative is called in entire process ? once ? twice? I'm good with both approaches, also maybe "NOTICE" string should be standardized at some point, since used in many places probably

emilmih: please find the new fix

qiluo-msft:
Code LGTM. Could you check the test failure?

emilmih:
i looked at the fails . i don't think has anything to do with my  code  change ->
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python2.7/dist-packages/swsscommon/swsscommon.py:2195: in set_entry
    super(ConfigDBConnector, self).set_entry(table, key, raw_data)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <swsscommon.swsscommon.ConfigDBConnector; proxy of <Swig Object of type 'swss::ConfigDBConnector_Native *' at 0x7f40c949e180> >
table = 'TEST_TABLE', key = 'key1', data = {'field1': 'value1'}

    def set_entry(self, table, key, data):
>       return _swsscommon.ConfigDBConnector_Native_set_entry(self, table, key, data)
E       RuntimeError: RedisReply catches system_error: command: *4
E       $5
E       HMSET
E       $15
E       TEST_TABLE|key1
E       $6
E       field1
E       $6
E       value1
E       , reason: Expected to get redis type 5 got type 2, err: NON-STRING-REPLY: Input/output error: Input/output error

/usr/lib/python2.7/dist-packages/swsscommon/swsscommon.py:2034: RuntimeError
--------------------------- Captured stderr teardown ---------------------------
Exception in thread Thread-3:
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/__w/1/s/tests/test_redis_ut.py", line 494, in listen_thread_func
    data = config_db.raw_to_typed(client.hgetall(key))
  File "/usr/lib/python2.7/dist-packages/swsscommon/swsscommon.py", line 1601, in hgetall
    return _swsscommon.DBConnector_hgetall(self, key)
RuntimeError: RedisReply catches system_error: command: *2
$7
HGETALL
$15
TEST_TABLE|key1
, reason: Expected to get redis type 2 got type 5, err: NON-STRING-REPLY: Input/output error: Input/output error

===================== 1 failed, 19 passed in 17.60 seconds =====================
##[error]Bash exited with code '1'.
Finishing: Run swss common unit tests

qiluo-msft (approved):

emilmih:
Thank you for approving it.  
Do I have to merge the pull request. Or someone with write access ?

qiluo-msft:
The 'Required' PR checkers need pass, before we merge this PR.

emilmih:
The required checkers passed

common/logger.h
qiluo-msft: I agree that "NOTICE" string should be standardized

PR #502, opened 2021-07-09
[ip address] Add isV6() helper member function

wendani:
Used in https://github.com/Azure/sonic-swss/pull/1817

### DIFF
--- a/common/ipaddress.h
+++ b/common/ipaddress.h
@@ -29,6 +29,11 @@ public:
         return m_ip.family == AF_INET;
     }
 
+    inline bool isV6() const
+    {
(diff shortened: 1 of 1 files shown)


common/ipaddress.h
```diff
@@ -29,6 +29,11 @@ class IpAddress
         return m_ip.family == AF_INET;
     }
 
+    inline bool isV6() const
```
prsunny, line 32: so far, we've been using this as `!isV4()`. Looks like its needed for certain cases?

wendani: I see. Thanks for pointing out. This then is not needed.

PR #505, opened 2021-07-21
Fix DBInterface blocking operations

qiluo-msft:
Fixes https://github.com/Azure/sonic-buildimage/issues/8202

- Fix pubsub channel timeout unit
- Reset context err before next redisGetReply, otherwise later redisGetReply early return on no data, and burn CPU core in a loop.

ref: https://github.com/redis/hiredis/issues/332


kcudnik (approved):

dgsudharsan:
Required for 202106

PR #506, opened 2021-07-22
[netlink] refill netlink cache when failing to get the link object by name

wangshengjun:
**Why I did it**
The getLinkByName may fail because of the outdated netlink cache

**How I did it**
Refill the netlink cache when we fail to get link object.

**How to verify it**

**Description for the changelog**

**A picture of a cute animal (not mandatory but encouraged)**

### DIFF
--- a/common/linkcache.cpp
+++ b/common/linkcache.cpp
@@ -77,5 +77,13 @@ string LinkCache::ifindexToName(int ifindex)
 
 struct rtnl_link* LinkCache::getLinkByName(const char *name)
 {
-    return rtnl_link_get_by_name(m_link_cache, name);
+    struct rtnl_link* link = NULL;
+    link = rtnl_link_get_by_name(m_link_cache, name);
+    if(NULL == link)
+    {
+        /* Trying to refill cache */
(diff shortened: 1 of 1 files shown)


wangshengjun:
@qiluo-msft  Could you please review this PR?

common/linkcache.cpp
```diff
@@ -77,5 +77,13 @@ string LinkCache::ifindexToName(int ifindex)
 
 struct rtnl_link* LinkCache::getLinkByName(const char *name)
```
qiluo-msft, line 78: >getLinkByName [](http://example.com/codeflow?start=29&length=13)

The change looks good to me. I am wondering how is this function used in application? I could not find any use cases.

qiluo-msft (approved):

common/linkcache.cpp
wangshengjun: > > getLinkByName
> 
> The change looks good to me. I am wondering how is this function used in application? I could not find any use cases.


Hi qiluo
Yes, the function is not yet used in the code. I use it when trying to do the optimization of VLAN  interface creation through netlink.

PR #510, opened 2021-07-24
[logger] Make map access thread safe and proper terminate thread

kcudnik:
Signed-off-by: kcudnik <kcudnik@gmail.com>

Fixes https://github.com/Azure/sonic-buildimage/issues/8177

I also made some minor code style refactoring

### DIFF
--- a/common/concurentmap.h
+++ b/common/concurentmap.h
@@ -1,0 +1,73 @@
+#pragma once
+
+#include <map>
+#include <string>
+#include <mutex>
+
+namespace swss
+{
+    template <typename K, typename V>
+        class ConcurentMap
+        {
+            public:
+
+                ConcurentMap() = default;
+
+            public:
+
+                size_t size()
+                {
+                    std::lock_guard<std::mutex> _lock(m_mutex);
+
+                    return m_map.size();
+                }
+
+                bool contains(const K& key)
+                {
+                    std::lock_guard<std::mutex> _lock(m_mutex);
+
+                    return m_map.find(key) != m_map.end();
+                }
+
+                void insert(const std::pair<K,V>& pair)
+                {
+                    std::lock_guard<std::mutex> _lock(m_mutex);
+
+                    m_map.insert(pair);
+                }
+
+                void set(const K& key, const V& value)
+                {
+                    std::lock_guard<std::mutex> _lock(m_mutex);
+
+                    m_map[key] = value;
+                }
+
+                // return copy
+                V get(const K& key)
+                {
+                    std::lock_guard<std::mutex> _lock(m_mutex);
+
+                    return m_map[key];
+                }
+
+                // return copy
+                std::map<K,V> getCopy()
+                {
+                    std::lock_guard<std::mutex> _lock(m_mutex);
+
+                    return m_map;
+                }
+
+            private:
+
+                ConcurentMap(const ConcurentMap&);
+                ConcurentMap& operator=(const ConcurentMap&);
+
+            private:
+
+                std::map<K,V> m_map;
+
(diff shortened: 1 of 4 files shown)


kcudnik:
test error on test_ConfigDBSubscribe not related to PR

kcudnik:
@arlakshm can you check this test_ConfigDBSubscribe ?

common/logger.cpp
```diff
@@ -31,13 +33,34 @@ void err_exit(const char *fn, int ln, int e, const char *fmt, ...)
...
+
+    if (m_settingThread)
+    {
+        m_runSettingThread = false;
+
         m_settingThread->join();
+
+        m_settingThread = nullptr;
+
+        m_runSettingThread = true;
```
mikeberesford, line 53: Might consider putting m_runSettingThread = true; in restartSettingThread? Not sure which is cleaner.

common/logger.cpp
```diff
@@ -208,27 +236,30 @@ void Logger::settingThread()
...
+            auto& field = fvField(i);
+            auto& value = fvValue(i);
+
+            if ((field == DAEMON_LOGLEVEL) && (value != m_currentPrios.get(key)))
             {
-                m_currentPrios[key] = value;
-                m_settingChangeObservers[key].first(key, value);
+                m_currentPrios.set(key, value);
+
+                m_settingChangeObservers.get(key).first(key, value);
```
mikeberesford, line 255: nit: extra blank line?

common/logger.cpp
```diff
@@ -130,26 +158,26 @@ void Logger::linkToDbWithOutput(const std::string &dbName, const PriorityChangeN
...
 
-void Logger::linkToDbNative(const std::string &dbName, const char * defPrio)
+void Logger::linkToDbNative(const std::string& dbName, const char * defPrio)
 {
-    auto& logger = getInstance();
-
     linkToDb(dbName, swssPrioNotify, defPrio);
-    logger.m_settingThread.reset(new std::thread(&Logger::settingThread, &logger));
+
+    getInstance().restartSettingThread();
```
mikeberesford, line 177: Is restarting the thread actually required if the thread is already started and the data structures are thread-safe? Put another way: is there any reason to maintain a separate linkToDbNative and linkToDb vs consolidating them into one function which starts the thread if needed?

kcudnik: thats a good question, im just fixing this instance here, since previously if you call linkToDbNative twice, it will crash, and whether we should have only 1 thread, not sure about that, this PR is just to make those maps thread safe + plus crahs bug fix not changing behavior, @qiluo-msft can you have opinion?

common/logger.cpp
kcudnik: remove

common/logger.cpp
kcudnik: moved to restart

kcudnik:
ping

common/concurentmap.h
```diff
@@ -0,0 +1,73 @@
+#pragma once
+
+#include <map>
+#include <string>
+#include <mutex>
+
+namespace swss
+{
+    template <typename K, typename V>
+        class ConcurentMap
```
mikeberesford, line 10: nit: typo (ConcurrentMap)

mikeberesford (approved):

common/concurentmap.h
kcudnik: fixed

kcudnik:
ping

kcudnik:
ping

qiluo-msft (approved):

PR #511, opened 2021-07-25
Add support for swig 4.0 in configure.ac

kcudnik:


kcudnik:
test error on test_ConfigDBSubscribe not related to PR

kcudnik:
@arlakshm can you check this test_ConfigDBSubscribe ?

qiluo-msft:
I am working on fixing the test_ConfigDBSubscribe.

configure.ac
```diff
@@ -31,7 +31,7 @@ esac],[debug=false])
 AM_CONDITIONAL(DEBUG, test x$debug = xtrue)
 AM_CONDITIONAL(ARCH64, test `getconf LONG_BIT` = "64")
 
-AC_PATH_PROG(SWIG, [swig3.0])
+AC_PATH_PROGS(SWIG, [swig3.0 swig])
```
qiluo-msft, line 34: Is `swig` only for version 4.0? If it is a general one, do we still need `swig3.0` ?

kcudnik: not exactly sure, but latest 4.0 binary is "swig" but 3.0 is "swig3.0", this macro will check for both in this order

PR #512, opened 2021-07-26
Fix test_ConfigDBSubscribe timing risk

qiluo-msft:
test_ConfigDBSubscribe has timing risk in unit test, because multiple thread is using the same ConfigDBConnector instance, and reuse Redis connection. So there are possibility to interleave Redis transaction in between.

What I did:
1. swap the role of main thread and spawned thread
2. spawned thread uses a new ConfigDBConnector instance locally constructed

### DIFF
--- a/tests/test_redis_ut.py
+++ b/tests/test_redis_ut.py
@@ -491,28 +491,12 @@ def test_ConfigDBSubscribe():
         assert data == test_data
         output_data = test_data['field1']
 
-    # the config_db.listen() is a blocking function so could not use that in the tests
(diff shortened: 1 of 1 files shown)


kcudnik (approved):

liuh-80:
202012 branch have some build break caused by this UT.
https://dev.azure.com/mssonic/build/_build/results?buildId=231625&view=logs&j=70844059-bb1d-5ada-0b5d-9ae5f1383c57&t=a4192146-319f-5109-00e8-3fd38d4ebc1e&s=ff05ad62-bb9a-53b6-ce9f-72f329a63e7c

PR #515, opened 2021-07-29
Support building on Bullseye

saiarcot895:
Add support for building on Bullseye with the following changes:

1. Add support for conditionally building Python 2 bindings. Python 2 is deprecated.
2. Use either swig 4.0 or swig 3.0, whichever is found. Bullseye has swig 4.0, Buster has 3.0.
3. Fix some compile errors by adding the necessary includes.


debian/rules
```diff
@@ -16,19 +16,33 @@ include /usr/share/dpkg/default.mk
 # package maintainers to append LDFLAGS
 #export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed
 
+DEBIAN_DIST_CODENAME := $(shell lsb_release -sc | sed -e 's/\#/ /g')
+
+ifneq (,$(findstring bullseye,$(DEBIAN_DIST_CODENAME)))
```
qiluo-msft, line 21: >bullseye [](http://example.com/codeflow?start=21&length=8)

Better to check old distributions: jessie, stretch, buster. All future versions should disable.

debian/python-swsscommon.install
```diff
@@ -1 +0,0 @@
-usr/lib/python2.7/dist-packages/swsscommon/*
```
qiluo-msft, line 1: Will you also delete the file `debian/python-swsscommon.dirs` ?

debian/rules
```diff
@@ -16,19 +16,33 @@ include /usr/share/dpkg/default.mk
...
+CONFIGURE_ARGS = --disable-python2
+else
+CONFIGURE_ARGS = --enable-python2
+endif
 
 # main packaging script based on dh7 syntax
 %:
-	dh $@  --with autotools-dev
+	dh $@  --with autoreconf
```
qiluo-msft, line 29: >autoreconf [](http://example.com/codeflow?start=15&length=10)

Why change this line?

debian/rules
```diff
@@ -16,19 +16,33 @@ include /usr/share/dpkg/default.mk
...
 
+override_dh_auto_configure:
+	dh_auto_configure -- $(CONFIGURE_ARGS)
+
 override_dh_auto_install:
 	dh_auto_install
+	if [ -d debian/tmp/usr/lib/python2.7/dist-packages/swsscommon ]; then \
+		dh_installdirs --package=python-swsscommon /usr/lib/python2.7/dist-packages/swsscommon/ ; \
+		cp -a debian/tmp/usr/lib/python2.7/dist-packages/swsscommon/* debian/python-swsscommon/usr/lib/python2.7/dist-packages/swsscommon/ ; \
+	fi
```
qiluo-msft, line 45: Is this section really needed?

debian/python-swsscommon.install
saiarcot895: I thought I did remove it, but looks like I didn't. I'll remove it and verify the build.

debian/rules
saiarcot895: `autotools-dev` is deprecated for the debhelper packages that are in Debian stretch and newer, and `autoreconf` is the replacement for it.

debian/rules
saiarcot895: Yes. For Buster and older, the python 2.7 bindings still need to be installed, but they shouldn't be installed for Bullseye, since they aren't built. Since the same set of files are used for both Buster and Bullseye, I had to have it work in both cases. If this is in the `.install` file, then if the files to copy in are missing, then debhelper errors out.

debian/rules
saiarcot895: I'll invert this check.

It doesn't look like Jessie is getting built anymore in master (at least for the virtual switch) build). Do you still want me to include Jessie in the check?

qiluo-msft: It does not hurt to add it. Or you can try and check the build.

Another concern: sonic-swss-common is also built in Ubuntu. I think you need to check Ubuntu versions numbers.

debian/rules
qiluo-msft: I am think about the motivation of the whole PR. If python2 package could be built in Bulleye, we just don't install it. What is wrong?

saiarcot895: I think building the python 2 package will still work in Bullseye (since the Python development headers are still available); I haven't tried this yet though. However, if we're not copying it in for installation, is it still worth it to build it?

It's entirely possible that the next version of Debian will not have any Python 2 at all. Then, we may need this kind of code, to conditionally build the Python 2 bindings, since our containers may be on an older version and the base image may be getting upgraded to a newer version. Also, there's already code in the sonic-buildimage repository for conditionally copying/installing the Python 2 packages.

configure.ac
```diff
@@ -28,10 +28,18 @@ AC_ARG_ENABLE(debug,
...
+	yes) python2=true ;;
+	no)  python2=false ;;
+	*) AC_MSG_ERROR(bad value ${enableval} for --enable-python2) ;;
+esac],[python2=true])
 AM_CONDITIONAL(DEBUG, test x$debug = xtrue)
+AM_CONDITIONAL(PYTHON2, test x$python2 = xtrue)
 AM_CONDITIONAL(ARCH64, test `getconf LONG_BIT` = "64")
 
-AC_PATH_PROG(SWIG, [swig3.0])
+AC_PATH_PROGS(SWIG, [swig4.0 swig3.0])
```
kcudnik, line 42: please add "swig also at the end: 
```
AC_PATH_PROGS(SWIG, [swig4.0 swig3.0 swig])
``` 
then my PR can be skipped: https://github.com/Azure/sonic-swss-common/pull/511

debian/python-swsscommon.install
saiarcot895: Done.

debian/rules
saiarcot895: Done.

configure.ac
saiarcot895: Done.

kcudnik (approved):
Wait for others to approve

qiluo-msft (approved):

saiarcot895:
Looks like `test_zero_cable_len_profile_update` and `test_SpeedAndBufferSet` are failing for some reason now. An earlier version of my changes passed all tests, and I don't see anything in my changes that would cause these tests to fail.

Any suggestions here?

PR #516, opened 2021-07-31
Fix: DBInterface::get() return nullable strings

qiluo-msft:


arlakshm (approved):

common/dbinterface.cpp
```diff
@@ -67,9 +67,9 @@ std::string DBInterface::get(const std::string& dbName, const std::string& hash,
             throw UnavailableDataError(message, hash);
         }
         const std::string& value = *pvalue;
-        return value == "None" ? "" : value;
+        return value == "None" ? shared_ptr<string>() : make_shared<string>(value);
     };
-    return blockable<std::string>(innerfunc, dbName, blocking);
+    return blockable<shared_ptr<string>>(innerfunc, dbName, blocking);
```
kcudnik, line 72: inconsistent namespace usage, std::shared_ptr in function return value vs shared_ptr in function body

kcudnik: same for std::string and string

qiluo-msft: Fixed

kcudnik (approved):

PR #526, opened 2021-09-06
Added support for setting TTL on a Table managed entry.

Cosmin-Jinga-MS:
[table.h/rediscommand.h] 
* Updated table.h with a new set() routine capable of also setting TTL on an entry.
* Updated rediscommand.h with the EXPIRE command

 Signed-off-by: Cosmin Jinga <v-cjinga@microsoft.com>

### DIFF
--- a/common/rediscommand.cpp
+++ b/common/rediscommand.cpp
@@ -88,6 +88,12 @@ void RedisCommand::formatHDEL(const std::string& key, const std::vector<std::str
     formatArgv(static_cast<int>(args.size()), args.data(), NULL);
 }
 
+/* Format EXPIRE key field command */
+void RedisCommand::formatEXPIRE(const std::string& key, const int32_t& ttl)
+{
+    return format("EXPIRE %s %d", key.c_str(), ttl);
+}
+
 const char *RedisCommand::c_str() const
 {
     return temp;
--- a/common/rediscommand.cpp
+++ b/common/rediscommand.cpp
@@ -94,6 +94,12 @@ void RedisCommand::formatEXPIRE(const std::string& key, const int64_t& ttl)
     return format("EXPIRE %s %lld", key.c_str(), ttl);
 }
 
+/* Format TTL key command */
+void RedisCommand::formatTTL(const std::string& key)
+{
+    return format("TTL %s", key.c_str());
+}
+
 const char *RedisCommand::c_str() const
 {
     return temp;
--- a/common/rediscommand.cpp
+++ b/common/rediscommand.cpp
@@ -89,7 +89,7 @@ void RedisCommand::formatHDEL(const std::string& key, const std::vector<std::str
 }
 
 /* Format EXPIRE key field command */
-void RedisCommand::formatEXPIRE(const std::string& key, const int& ttl)
+void RedisCommand::formatEXPIRE(const std::string& key, const int64_t& ttl)
 {
     return format("EXPIRE %s %d", key.c_str(), ttl);
 }
--- a/common/rediscommand.cpp
+++ b/common/rediscommand.cpp
@@ -91,7 +91,7 @@ void RedisCommand::formatHDEL(const std::string& key, const std::vector<std::str
 /* Format EXPIRE key field command */
 void RedisCommand::formatEXPIRE(const std::string& key, const int64_t& ttl)
 {
-    return format("EXPIRE %s %d", key.c_str(), ttl);
+    return format("EXPIRE %s %lld", key.c_str(), ttl);
 }
 
 const char *RedisCommand::c_str() const
--- a/common/rediscommand.cpp
+++ b/common/rediscommand.cpp
@@ -89,7 +89,7 @@ void RedisCommand::formatHDEL(const std::string& key, const std::vector<std::str
 }
 
 /* Format EXPIRE key field command */
-void RedisCommand::formatEXPIRE(const std::string& key, const int32_t& ttl)
+void RedisCommand::formatEXPIRE(const std::string& key, const int& ttl)
 {
     return format("EXPIRE %s %d", key.c_str(), ttl);
 }
--- a/common/rediscommand.h
+++ b/common/rediscommand.h
@@ -51,7 +51,7 @@ public:
     void formatHDEL(const std::string& key, const std::vector<std::string>& fields);
 
     /* Format EXPIRE key field command */
-    void formatEXPIRE(const std::string& key, const int& ttl);
+    void formatEXPIRE(const std::string& key, const int64_t& ttl);
 
     const char *c_str() const;
 
--- a/common/rediscommand.h
+++ b/common/rediscommand.h
@@ -51,7 +51,7 @@ public:
     void formatHDEL(const std::string& key, const std::vector<std::string>& fields);
 
     /* Format EXPIRE key field command */
-    void formatEXPIRE(const std::string& key, const int32_t& ttl);
+    void formatEXPIRE(const std::string& key, const int& ttl);
 
     const char *c_str() const;
 
--- a/common/rediscommand.h
+++ b/common/rediscommand.h
@@ -50,6 +50,9 @@ public:
     /* Format HDEL key multiple fields command */
     void formatHDEL(const std::string& key, const std::vector<std::string>& fields);
 
+    /* Format EXPIRE key field command */
+    void formatEXPIRE(const std::string& key, const int32_t& ttl);
+
     const char *c_str() const;
 
     size_t length() const;
--- a/common/rediscommand.h
+++ b/common/rediscommand.h
@@ -50,8 +50,11 @@ public:
     /* Format HDEL key multiple fields command */
     void formatHDEL(const std::string& key, const std::vector<std::string>& fields);
 
-    /* Format EXPIRE key field command */
+    /* Format EXPIRE key ttl command */
     void formatEXPIRE(const std::string& key, const int64_t& ttl);
+
+    /* Format TTL key command */
+    void formatTTL(const std::string& key);
 
     const char *c_str() const;
 
--- a/common/table.cpp
+++ b/common/table.cpp
@@ -138,6 +138,36 @@ void Table::set(const string &key, const vector<FieldValueTuple> &values,
     }
 }
 
+void Table::set(const string &key, const vector<FieldValueTuple> &values,
+                const int32_t &ttl, const string& /*op*/, const string& /*prefix*/)
+{
+    if (values.size() == 0)
+        return;
+
+    RedisCommand cmd;
+    cmd.formatHMSET(getKeyName(key), values.begin(), values.end());
+
+    m_pipe->push(cmd, REDIS_REPLY_STATUS);
+
+    if (!m_buffered)
+    {
+        m_pipe->flush();
+    }
+
+    if (ttl != DEFAULT_DB_TTL)
+    {
+      // Configure the expire time for the entry that was just added
+      cmd.formatEXPIRE(getKeyName(key), ttl);
+
+      m_pipe->push(cmd, REDIS_REPLY_INTEGER);
+
+      if (!m_buffered)
+      {
+          m_pipe->flush();
+      }
+    }
+}
+
 void Table::del(const string &key, const string& /* op */, const string& /*prefix*/)
 {
     RedisCommand del_key;
--- a/common/table.cpp
+++ b/common/table.cpp
@@ -132,10 +132,6 @@ void Table::set(const string &key, const vector<FieldValueTuple> &values,
     cmd.formatHMSET(getKeyName(key), values.begin(), values.end());
 
     m_pipe->push(cmd, REDIS_REPLY_STATUS);
-    if (!m_buffered)
-    {
-        m_pipe->flush();
-    }
 
     if (ttl != DEFAULT_DB_TTL)
     {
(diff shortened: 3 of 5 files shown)


TACappleman:
@qiluo-msft could you review this PR please?

common/table.cpp
```diff
@@ -138,6 +138,36 @@ void Table::set(const string &key, const vector<FieldValueTuple> &values,
     }
 }
 
+void Table::set(const string &key, const vector<FieldValueTuple> &values,
+                const int32_t &ttl, const string& /*op*/, const string& /*prefix*/)
```
kcudnik, line 142: why op and prefix are omitted ? set should work exactly the same as with or without ttl, otherwise should be named differently

common/table.h
```diff
@@ -170,11 +170,21 @@ class Table : public TableBase, public TableEntryEnumerable {
     Table(RedisPipeline *pipeline, const std::string &tableName, bool buffered);
     ~Table() override;
 
+    /* The default time to live for a DB entry is infinite */
+    static constexpr int DEFAULT_DB_TTL = -1;
```
kcudnik, line 174: inconsistent naming, int here, int32_t in function

common/rediscommand.h
```diff
@@ -50,6 +50,9 @@ class RedisCommand {
     /* Format HDEL key multiple fields command */
     void formatHDEL(const std::string& key, const std::vector<std::string>& fields);
 
+    /* Format EXPIRE key field command */
+    void formatEXPIRE(const std::string& key, const int32_t& ttl);
```
qiluo-msft, line 54: >const int32_t& [](http://example.com/codeflow?start=46&length=14)

`int` is good enough.
How do you know it's 32 bits? It seems supporting 64 bits.

Cosmin-Jinga-MS: Agreed, changing the type to int.

common/table.cpp
Cosmin-Jinga-MS: op and prefix are also omitted in the existing set() routine for Table
I've overloaded set() with the extra ttl argument to have a distinction between it and the existing one.
I'm adding a reworked version of the code with the ttl is set directly in the existing routine.

common/table.h
Cosmin-Jinga-MS: should be int, issue addressed in the updated version

Cosmin-Jinga-MS (commented):
Addressed comments;
Main update is removal of the new overloaded set() and moving of the ttl configuration inside the existing routine.

common/rediscommand.h
qiluo-msft: Is `int64_t` better?

Cosmin-Jinga-MS: Redis documentation on expire does not actually provide a range for the ttl values [https://redis.io/commands/expire](https://redis.io/commands/expire)
Having manually tried an in-range and an out-of-range int64 value it does indeed need too be int64.

> 127.0.0.1:6379> EXPIRE ISIS_ROUTER_LSP_STATE:localhost:1 9223372036854775801
(integer) 1
127.0.0.1:6379> EXPIRE ISIS_ROUTER_LSP_STATE:localhost:1 9223372036854775809
(error) ERR value is not an integer or out of range
I'll update the code momentarily.

Cosmin-Jinga-MS (commented):
Updated ttl to int64_type and changed its position in the set() params to fix overlooked compilation errors in existing calls for set()

common/rediscommand.cpp
```diff
@@ -88,6 +88,12 @@ void RedisCommand::formatHDEL(const std::string& key, const std::vector<std::str
     formatArgv(static_cast<int>(args.size()), args.data(), NULL);
 }
 
+/* Format EXPIRE key field command */
+void RedisCommand::formatEXPIRE(const std::string& key, const int64_t& ttl)
+{
+    return format("EXPIRE %s %d", key.c_str(), ttl);
```
qiluo-msft, line 94: >%d [](http://example.com/codeflow?start=29&length=2)

hiredis internally implement some int length. ref: [https://github.com/redis/hiredis/blob/2d9d77518d012a54ae34f9822e4b4d19823c4b75/hiredis.c#L434](https://github.com/redis/hiredis/blob/2d9d77518d012a54ae34f9822e4b4d19823c4b75/hiredis.c#L434)

Seems like `%lld` is for int64_t. Could you check and add unit test?

common/table.cpp
```diff
@@ -136,6 +136,19 @@ void Table::set(const string &key, const vector<FieldValueTuple> &values,
...
+    if (ttl != DEFAULT_DB_TTL)
+    {
+      // Configure the expire time for the entry that was just added
+      cmd.formatEXPIRE(getKeyName(key), ttl);
+
+      m_pipe->push(cmd, REDIS_REPLY_INTEGER);
+
+      if (!m_buffered)
+      {
+          m_pipe->flush();
```
qiluo-msft, line 149: >flush [](http://example.com/codeflow?start=18&length=5)

Is it possible to flush only once in this function?

Cosmin-Jinga-MS: Does RedisPipeline support enqueuing of more than one command without it being flushed inbetween?
i.e.:

>     cmd.formatHMSET(getKeyName(key), values.begin(), values.end());
>     m_pipe->push(cmd, REDIS_REPLY_STATUS);
>     cmd.formatEXPIRE(getKeyName(key), ttl);
>     m_pipe->push(cmd, REDIS_REPLY_INTEGER);
>     if (!m_buffered)
>     {
>         m_pipe->flush();
>     }
If it does the code can be changed to:

>     if (ttl != DEFAULT_DB_TTL)
>     {
>       // Configure the expire time for the entry that was just added
>       cmd.formatEXPIRE(getKeyName(key), ttl);
> 
>       m_pipe->push(cmd, REDIS_REPLY_INTEGER);
> 
>       if (!m_buffered)
>       {
>           m_pipe->flush();
>       }
>     }
>     else
>     {
>        if (!m_buffered)
>        {
>           m_pipe->flush();
>        }
>     }

common/rediscommand.cpp
Cosmin-Jinga-MS: yes, I will update it to %lld and add/extend an UT to also issue a set with configured TTL

Cosmin-Jinga-MS:
@qiluo-msft Updating set() with an extra param for ttl rather than overloading it has raised compilation dependencies in sonic-swss's mock_table.cpp which I've missed during the local build of swss-common but have been uncovered by the pipeline now.
What would be the best approach here? Update sonic-swss as well with the extra ttl param? Not sure how to integrate it in the same PR as it's from a different submodule.
Should I revert back to overloading set()?

common/table.cpp
qiluo-msft: By design, yes. In your proposed code, you can further improve by move `flush` out of if-else.

qiluo-msft:
I see this is painful. Seems overloading `set()` is better approach. If you choose this way, please make sure
1. reuse code as much as possible, for example, rewrite old function by calling your new function
2. Add TODO comment to remove the overloading and use default parameter


---
In reply to: [915814243](https://github.com/Azure/sonic-swss-common/pull/526#issuecomment-915814243) [](http://example.com/codeflow?ancestors=915814243,915814243)

common/table.cpp
kcudnik: @qiluo-msft what do you think here?

common/rediscommand.cpp
kcudnik: return in void function? formta() also is void?

common/table.cpp
```diff
@@ -132,6 +132,14 @@ void Table::set(const string &key, const vector<FieldValueTuple> &values,
     cmd.formatHMSET(getKeyName(key), values.begin(), values.end());
 
     m_pipe->push(cmd, REDIS_REPLY_STATUS);
+
+    if (ttl != DEFAULT_DB_TTL)
+    {
+      // Configure the expire time for the entry that was just added
+      cmd.formatEXPIRE(getKeyName(key), ttl);
```
qiluo-msft, line 139: >       [](http://example.com/codeflow?start=0&length=6)

Please indent with 4x spaces

Cosmin-Jinga-MS (commented):
Updated the code back to overloaded set() routine and added an UT to check the TTL setting (along with the new Table::ttl() routine to enable querying for said ttl)

common/table.cpp
Cosmin-Jinga-MS: updated indents in last commit

common/table.cpp
```diff
@@ -132,12 +132,58 @@ void Table::set(const string &key, const vector<FieldValueTuple> &values,
...
+//       to existing set() command once sonic-swss's mock_table.cpp and other
+//       dependencies can be updated to use the extended new default set())
+void Table::set(const string &key, const vector<FieldValueTuple> &values,
+                const int64_t &ttl, const string& /*op*/, const string& /*prefix*/)
+{
+    if (values.size() == 0)
+        return;
+
+    RedisCommand cmd;
+    // @Qi Luo not directly reusing default set() here as it will require we
```
qiluo-msft, line 151: >@Qi Luo [](http://example.com/codeflow?start=7&length=7)

Do not quote others in code comment...

qiluo-msft, line 151: >not directly reusing default set() here as it will require we [](http://example.com/codeflow?start=15&length=61)

You can treat this function as the full-fledge function, and let old `set` calling this one.

common/table.h
```diff
@@ -175,10 +178,20 @@ class Table : public TableBase, public TableEntryEnumerable {
                      const std::vector<FieldValueTuple> &values,
                      const std::string &op = "",
                      const std::string &prefix = EMPTY_PREFIX);
+
+    /* Set an entry in the DB directly and configure ttl for it (op not in use) */
+    virtual void set(const std::string &key,
```
qiluo-msft, line 183: >set [](http://example.com/codeflow?start=17&length=3)

You should treat this overload function as a mimic of original `set` function with another optional parameter. So please only add `ttl` paramter at the end. Then `op` and `prefix` will be non-optional in this overload function.

common/table.cpp
Cosmin-Jinga-MS: Not sure I understand what you meant here.
Do you mean I should use the old set() inside the new one or the other way around?
I'm assuming the first option, as doing it the other way around would imply having both set be extended with an extra parameter(and triggering the sonic-swss dependencies).
It was my initial intention to use the old set() there but I wanted to double check it's okay to call flush twice(as per a previous discussion we had).

common/table.h
Cosmin-Jinga-MS: ok, I will update it accordingly

Cosmin-Jinga-MS:
Updated code with fixes for latest comments

Cosmin-Jinga-MS (commented):
Addressed param ordering in new overloaded set()

common/table.cpp
qiluo-msft: Thanks for implement the first option! I agree it has one extra `flush` and hurt the performance a little bit.

Actually I mean the other option. To be clear, I am proposing
```c++
void Table::set(const string &key, const vector<FieldValueTuple> &values,
                const string& op, const string& prefix)
{
    set(key, values, op, prefix, DEFAULT_DB_TTL);
}
```

Cosmin-Jinga-MS: I see, thanks for the clarification.

Cosmin-Jinga-MS (commented):
Added new set() as the driving routine for set OPs

qiluo-msft (approved):

Cosmin-Jinga-MS:
Hi @qiluo-msft,
About the remaining failing UT from the current pipeline. https://dev.azure.com/mssonic/build/_build/results?buildId=35308&view=ms.vss-test-web.build-test-results-tab
 I'm not seeing anything wrong with the set operation in it(it calls the old set() implementation, nothing touches the new ttl code in this scenario) and it seems it's not a particularly stable UT(looking into its history)
image
Is there a chance a retry of this pipeline would fix it?(I don't have enough rights on my account to re-run failed UTs from the gitlab pipeline interface)

qiluo-msft:
The PR owner could use `/azpw run` to retrigger checkers.

Cosmin-Jinga-MS:
Looks like the failure is still present. 
I've ran the UT in question locally and managed to repro the issue both with my changes from swss-common in place and with a fresh build from master. Currently looking a bit at the failure to see if I can spot the problem but it does not look to be caused by my changes.
Syslog err output from the sonic-swss-common master run(without my changes):
image
Looks like a failure to program those interfaces due to confilct on the lane/ids to me which leads to a orchagent crash.

Cosmin-Jinga-MS:
@qiluo-msft 
I did not manage to get very far with the investigation into this as orchagent logs are rate-limited before the failure occurs once it's set to DEBUG level. Are there any overnight/on-demand runs we can check to confirm this tests failure reproduces on master as well?(as said before, I'm seeing this failure on my local master ut run too)
If it does repro on a master pipeline as well, it would be better to defer the investigation to someone more familiar with  recirc ports, in my opinion.

qiluo-msft:
You can check master branch, other branches or PR on https://dev.azure.com/mssonic/build/_build?definitionId=9&_a=summary&view=branches

I just trigged a new build on master branch.

TACappleman:
One of the ones that's failed on the latest run fails on a large percentage of pipelines, and the other one sometimes, so worth keeping on trying until it works

Cosmin-Jinga-MS:
@qiluo-msft I see the master run has also failed on the  test_recirc_port UT(it's now running a 2nd attempt). My PR retry also failed that UT again. Should I keep retrying too?

qiluo-msft:
I am trying to remove the failure test case in https://github.com/Azure/sonic-swss/pull/1911

PR #528, opened 2021-09-10
[schema] Add IS-IS and LDP config and alarms

TACappleman:


TACappleman:
@qiluo-msft Could you review this PR please?

common/schema.h
```diff
@@ -104,6 +104,18 @@ namespace swss {
 
 #define APP_NEIGH_RESOLVE_TABLE_NAME        "NEIGH_RESOLVE_TABLE"
 
+#define APP_REJECTED_CONFIG                 "REJECTED_CONFIG"
```
prsunny, line 107: This is too generic name. Please rename to a specific table that is intended

TACappleman: It's a table to report configuration that's been rejected by one of the SONiC containers due to being malformed - so REJECTED_CONFIG is the specific name for it

prsunny: Which sonic container report this configuration? can you provide an HLD/doc that captures this implementation?

prsunny (changes requested):

PR #529, opened 2021-09-15
Routed subinterface enhancements

preetham-singh:
What I did
Routed subinterfae enhancements HLD [#833](https://github.com/Azure/SONiC/pull/833)
Add support for long name and short name routed subinterfaces.
Add support for long name and short name routed subinterfaces.
This swss common library provides APIS for:
- Subinterface validation checks
- Get parent interface corresponding to subinterface(short/long name)
- Get subinterface index
 
Why I did it
Routed subinterface feature was broken for physical and portchannel subinterfaces for subinterface name exceeding 15 characters due to kernel limitation of netdev name length of 15.

How I verified it
Routed subinterface Unit tests


common/ifcommon.cpp
```diff
@@ -0,0 +1,120 @@
...
+bool subIntf::isShortName()
+{
+    if (isCompressed == true)
+    {
+        return true;
+    }
+    else
+    {
+        return false;
+    }
```
kcudnik, line 119: this could be single line return iscommpressed == true;

common/ifcommon.h
```diff
@@ -0,0 +1,25 @@
...
+        public:
+            subIntf(const std::string &ifName);
+            bool isValid();
+            std::string parentIntf();
+            int subIntfIdx();
+            std::string longName();
+            std::string shortName();
+            bool isShortName();
+        private:
+#define VLAN_SUB_INTERFACE_SEPARATOR   "."
```
kcudnik, line 16: define is not bound by class or namespace, ether put it on top of the file (and also use pragma once) or make it as constexpr

common/ifcommon.h
```diff
@@ -0,0 +1,25 @@
...
+            std::string alias = "";
+            std::string subIfIdx = "";
+            std::string parentIf = "";
+            std::string parentIfShortName = "";
```
kcudnik, line 20: initialization strings like this is not required, they by default are empty

common/ifcommon.h
```diff
@@ -0,0 +1,25 @@
...
+    class subIntf
+    {
+        public:
+            subIntf(const std::string &ifName);
+            bool isValid();
+            std::string parentIntf();
+            int subIntfIdx();
+            std::string longName();
+            std::string shortName();
+            bool isShortName();
```
kcudnik, line 14: all those functions should be const

common/ifcommon.h
```diff
@@ -0,0 +1,25 @@
+#ifndef __IFCOMMON_H__
+#define __IFCOMMON_H__
+
+namespace swss {
+    class subIntf
```
kcudnik, line 5: preffered cilename should be as class it defines so like subintf.h

common/ifcommon.cpp
preetham-singh: Updated as part of next commit.

common/ifcommon.h
preetham-singh: Updated as part of next commit.

common/ifcommon.h
preetham-singh: Updated file names.

kcudnik (approved):

lguohan:
@prsunny , why is this pr on swss-common? are you sponsoring this pr?

preetham-singh:
Closing this pull request since this code is moved to swss/lib which can be found as part of swss PR [2017](https://github.com/Azure/sonic-swss/pull/2017)

PR #531, opened 2021-09-17
[ci]: Support pull request diff coverage

xumia:
Support pull request diff coverage


kcudnik:
@xumia  @qiluo-msft  this will cause to only show diff coverage, and coverage for echa separate file like is shown by defaul azure pipeline will not be possible !: take a look and address here: https://github.com/Azure/sonic-sairedis/pull/944#issue-1006990443

PR #543, opened 2021-10-22
Add default route table name

prsunny:
1. Add default route table name to State DB
2. Fix to follow the naming convention for BFD


qiluo-msft:
This commit could not be cleanly cherry-picked to 202012. Please submit another PR.

PR #547, opened 2021-10-27
[PINS] Add ConsumerNotifier for PINS response path

donNewtonAlpha:


qiluo-msft:
Please change the PR title to more descriptive. People in the swss-common context do not know `Pr3`

common/dbconnector.cpp
```diff
@@ -627,11 +628,6 @@ int64_t DBConnector::del(const string &key)
 bool DBConnector::exists(const string &key)
 {
     RedisCommand rexists;
-    if (key.find_first_of(" \t") != string::npos)
```
qiluo-msft, line 630: >if (key.find_first_of(" \t") != string::npos) [](http://example.com/codeflow?start=4&length=45)

This is behavior change. Do you have unit test?

.bazelrc
```diff
@@ -0,0 +1 @@
+build --cxxopt="--std=c++17"
```
qiluo-msft, line 1: >std=c++17 [](http://example.com/codeflow?start=18&length=9)

The current build is using `std=c++11`. Could you also use it? Or will you change current build to `std=c++17`?

It will be tricky if we need to maintain 2 build env in future.

common/json.hpp
```diff
@@ -5394,7 +5394,7 @@ class basic_json
                 {
                     assert(lhs.m_value.array != nullptr);
                     assert(rhs.m_value.array != nullptr);
-                    return *lhs.m_value.array < *rhs.m_value.array;
+                    return (*lhs.m_value.array) < *rhs.m_value.array;
```
qiluo-msft, line 5397: >( [](http://example.com/codeflow?start=27&length=1)

Is it a necessary change?

qiluo-msft:
lease give a link to HLD for this change.

common/schema.h
```diff
@@ -45,14 +45,14 @@ namespace swss {
...
-#define APP_P4RT_IPV6_TABLE_NAME              "P4RT_IPV6_TABLE"
-#define APP_P4RT_ACL_TABLE_DEFINITION_NAME    "P4RT_ACL_TABLE_DEFINITION"
-#define APP_P4RT_MIRROR_SESSION_TABLE_NAME    "P4RT_MIRROR_SESSION_TABLE"
+#define APP_P4RT_ROUTER_INTERFACE_TABLE_NAME  "FIXED_ROUTER_INTERFACE_TABLE"
+#define APP_P4RT_NEIGHBOR_TABLE_NAME          "FIXED_NEIGHBOR_TABLE"
+#define APP_P4RT_NEXTHOP_TABLE_NAME           "FIXED_NEXTHOP_TABLE"
+#define APP_P4RT_WCMP_GROUP_TABLE_NAME        "FIXED_WCMP_GROUP_TABLE"
+#define APP_P4RT_IPV4_TABLE_NAME              "FIXED_IPV4_TABLE"
+#define APP_P4RT_IPV6_TABLE_NAME              "FIXED_IPV6_TABLE"
+#define APP_P4RT_ACL_TABLE_DEFINITION_NAME    "ACL_TABLE_DEFINITION_TABLE"
```
qiluo-msft, line 54: >ACL_ [](http://example.com/codeflow?start=47&length=4)

Do you want add a prefix like `P4RT_` or `FIXED_`?

kcudnik:
this seems to me to be too big change like for a single PR

common/schema.h
prsunny: @qiluo-msft , this is an overlap with another PR - https://github.com/Azure/sonic-swss-common/pull/548. This was discussed in subgroup and plan to have the without P4RT prefix.

BUILD
```diff
@@ -0,0 +1,31 @@
+package(default_visibility = ["//visibility:public"])
```
bocon13, line 1: Integrate Bazel build into PR checker (update the build pipeline) to help catch errors for breaking this in the future

mint570:
@qiluo-msft
About the performance impact we discussed in the meeting, this is only used in P4RT code currently. For existing SONiC codes, they are unchanged (They use the real class instead of the interface. So there shouldn't be performance impact.)

Also, the interface won't change the existing swss-common unit test. They still need to test with the Redis instance. The interface mocking is only helpful for testing codes that use the swss-common libraries. (Such as P4RT, so that P4RT does not need a Redis instance when doing unit test.)

bocon13:
in modern C++, there is no performance penalty for virtual functions when the compiler can guarantee the runtime type. In our case, the implementation class is inferred by the compiler, resulting in no additional instructions added by `virtual`.

The compiler option involved here is: `-fdevirtualize` which is automatically enabled at -O2.
https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

The function may be inlined or called directly without a vtable lookup.

Here's a relevant blog post on the topic of virtual functions and final classes:
https://devblogs.microsoft.com/cppblog/the-performance-benefits-of-final-classes/

Anyway, I ran an experiment with and without virtual functions, and my implementation class (DBConnector) was NOT final. The result is identical assembly code.

`gcc --std=c++17 -O2`

With virtual:
```
        call    std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string<std::allocator<char> >(char const*, std::allocator<char> const&)
        mov     rsi, rsp
        mov     rdi, r12
        call    DBConnector::get(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
        mov     rdi, rsp
        mov     ebp, eax
```

Without virtual:
```
        call    std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string<std::allocator<char> >(char const*, std::allocator<char> const&)
        mov     rsi, rsp
        mov     rdi, rbp
        call    DBConnector::get(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
        mov     rdi, rsp
        mov     r12d, eax
```

.bazelrc
bocon13: I think we can leave this blank and Bazel will use the system default. This only applies to builds directly in this repo. Other repos (sonic-pins) will set their own versions.

@rhalstea do you have any objections?

common/dbconnectorinterface.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
```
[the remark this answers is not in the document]
bocon13: Update: See comment on main PR regarding virtual functions

common/json.hpp
bocon13: This is a fix from upstream: https://github.com/nlohmann/json/issues/590

.bazelrc
rhalstea: no objection.

IIRC there may be a few places where we use c++17 features, but they should be easy to update. What comes to mind is: `for (const auto& [var,key] : container)`

bocon13: I think these are all in the sonic-pins repo, and that will have its own .bazelrc file. I think safe to remove.

bocon13: Removed in #557

common/json.hpp
bocon13: Addressed in #545

common/schema.h
bocon13: removed from this PR

BUILD
bocon13: Move to separate PR

common/producerstatetable.cpp
```diff
@@ -358,4 +358,9 @@ void ProducerStateTable::apply_temp_view()
     m_tempViewActive = false;
 }
 
+std::string ProducerStateTable::get_table_name() const
```
qiluo-msft, line 361: This function does not add value, can you remove?

bocon13: We use this in the P4RT server application.

@rhalstea what do you think?

common/dbconnector.cpp
```diff
@@ -750,6 +751,19 @@ int64_t DBConnector::decr(const string &key)
     return r.getContext()->integer;
 }
 
+#ifndef SWIG
+std::unordered_map<std::string, std::string> DBConnector::hgetall(
```
qiluo-msft, line 755: >hgetall [](http://example.com/codeflow?start=58&length=7)

The template version is a general solution for all STL container types. Why we need to specialize it to unordered_map?

common/consumernotifier.cpp
```diff
@@ -0,0 +1,41 @@
...
+        s.addSelectable(notification_consumer_.get());
+        Selectable *sel;
+
+        int error = s.select(&sel, timeout_ms);
+        if (error != Select::OBJECT)
+        {
+            return false;
+        }
+    }
+    notification_consumer_->pop(op, data, values);
```
qiluo-msft, line 37: This logic is applicable to any Selectable or Select, which contains multiple Selectable. Could you move it to their class?

common/producerstatetable.cpp
rhalstea: It's useful at various points in P4RT App to know the table name. Particularly around logging. The function was added here to make it available as part of the interface.

common/dbconnector.cpp
bocon13: @rhalstea what do you think?

common/consumernotifierinterface.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <vector>
+
+#include "table.h"
+
+namespace swss
+{
+
+// Wrapper class to use swss::NotificationConsumer.
+class ConsumerNotifierInterface
```
qiluo-msft, line 11: >ConsumerNotifierInterface [](http://example.com/codeflow?start=6&length=25)

I feel the `*Interface` classes just mark some functions as virtual. Can we remove all the classes?

qiluo-msft:
On the proposal of not adding `virtual`, I find a discussion [https://stackoverflow.com/a/45636400/2514803](https://stackoverflow.com/a/45636400/2514803). Will it help keep production code intact?

---
In reply to: [963804275](https://github.com/Azure/sonic-swss-common/pull/547#issuecomment-963804275) [](http://example.com/codeflow?ancestors=963804275,963804275)

common/dbconnector.cpp
rhalstea: The templates didn't work with the virtual interfaces.

Where this is needed in the P4RT App could be updated to the Table class. We can remove this once that happens, but that change may be difficult with the MVP timelines. However, internally we do want it planned for the next push.

common/producerstatetable.cpp
qiluo-msft: It's already defined in TableBase::getTableName(), and TableBase could be treated as an interface.

common/consumernotifierinterface.h
rhalstea: We can, but the other issue these interface classes address is the base class constructors requiring a redis db to be active. If we remove this do you have any objections to having a protected constructor that doesn't initialize a redis connection?

common/producerstatetable.cpp
```diff
@@ -358,4 +358,9 @@ void ProducerStateTable::apply_temp_view()
...
+{
+  return getTableName();
```
qiluo-msft, line 363: >return [](http://example.com/codeflow?start=2&length=6)

Please follow the code format convention in existing code. Such as 1. use 4-space indentation. 2. `m_` prefix for non-public member variables.

common/dbconnector.h
```diff
@@ -199,11 +200,14 @@ class DBConnector : public RedisContext
     ReturnType hgetall(const std::string &key);
 
 #ifndef SWIG
+    std::unordered_map<std::string, std::string> hgetall(
```
bocon13, line 203: Qi's suggestion: make this a standalone function in sonic-pins; first parameter is DBConnector that calls the template version.

common/consumernotifier.h
```diff
@@ -0,0 +1,27 @@
...
+
+// Wrapper class to use swss::NotificationConsumer.
+class ConsumerNotifier : public ConsumerNotifierInterface
+{
+  public:
+    ConsumerNotifier(const std::string &notifier_channel_name, DBConnector *db_connector);
+
+    bool WaitForNotificationAndPop(std::string &op, std::string &data, std::vector<FieldValueTuple> &values,
+                                   int64_t timeout_ms = 60000LL) override;
+
```
bocon13, line 21: Robert's proposal
```
protected:
  ConsumerNotifier();  // used for testing
```

Alternative:
- Provide a fake/mock DBConnector as a subclass and use that; need to create a testing only constructor there

rhalstea:
@bocon13 @qiluo-msft 

I've updated sonic-pins to include the fakes/mocks so this pull request shouldn't be needed anymore.

donNewtonAlpha:
Hi Robert can you put a comment on the PR and I can close

donNewtonAlpha:
Closing PER @rhalstea (Google)

PR #553, opened 2021-11-09
Exclude *.a files from python deb packages

qiluo-msft:
Cherry-pick https://github.com/Azure/sonic-swss-common/pull/554 to master branch

The file `_swsscommon.a` in the python deb packages takes > 33 MB, and it is not used.
Remove it to save disk space.


saiarcot895:
Can `--disable-static` be used to disable building the static libraries entirely?

qiluo-msft:
@saiarcot895 has concern that if `--disable-static` worked, so that the static files wouldn't be built in the first place

saiarcot895:
Changes in this PR were done in #700.

PR #554, opened 2021-11-09
Exclude *.a files from python deb packages

qiluo-msft:
The file `_swsscommon.a` in the python deb packages takes > 33 MB, and it is not used.
Remove it to save disk space.

### DIFF
--- a/debian/python-swsscommon.install
+++ b/debian/python-swsscommon.install
@@ -1,1 +1,3 @@
-usr/lib/python2.7/dist-packages/swsscommon/*
+usr/lib/python2.7/dist-packages/swsscommon/*.py*
+usr/lib/python2.7/dist-packages/swsscommon/*.so*
+usr/lib/python2.7/dist-packages/swsscommon/*.la*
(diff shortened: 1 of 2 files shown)


debian/python-swsscommon.install
```diff
@@ -1 +1,3 @@
-usr/lib/python2.7/dist-packages/swsscommon/*
+usr/lib/python2.7/dist-packages/swsscommon/*.py*
+usr/lib/python2.7/dist-packages/swsscommon/*.so*
+usr/lib/python2.7/dist-packages/swsscommon/*.la*
```
lguohan, line 3: what is la, do we need them?

lguohan (approved):

debian/python-swsscommon.install
qiluo-msft: It is a textual file that includes a description of the library. It allows libtool to create platform-independent names.
ref: https://stackoverflow.com/a/1238079/2514803

It is a quite small text file and describe other files, we can keep it.

saiarcot895 (approved):

PR #558, opened 2021-11-12
Use build profiles instead of distro version for Python 2 binding build

saiarcot895:
Instead of using the distro version to determine if the Python 2
bindings should be built, use deb profiles instead to specify if they
should be built or not. The main reason of this is that normal .install
and .dirs files can be used for both the Python 2 and Python 3 bindings,
instead of .install/.dir files for Python 3 and debian/rules for Python
2.

More details on build profiles are [here](https://wiki.debian.org/BuildProfileSpec).

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -64,7 +64,7 @@ jobs:
   - script: |
       set -ex
       ./autogen.sh
-      fakeroot debian/rules DEB_CONFIGURE_EXTRA_FLAGS='--enable-code-coverage' CFLAGS="" CXXFLAGS="--coverage -fprofile-abs-path" LDFLAGS="--coverage -fprofile-abs-path" DEB_BUILD_PROFILES=python2 binary && cp ../*.deb .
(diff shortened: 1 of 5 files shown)


.azure-pipelines/build-template.yml
```diff
@@ -64,7 +64,7 @@ jobs:
   - script: |
       set -ex
       ./autogen.sh
-      fakeroot debian/rules DEB_CONFIGURE_EXTRA_FLAGS='--enable-code-coverage' CFLAGS="" CXXFLAGS="--coverage -fprofile-abs-path" LDFLAGS="--coverage -fprofile-abs-path" binary && cp ../*.deb .
+      fakeroot debian/rules DEB_CONFIGURE_EXTRA_FLAGS='--enable-code-coverage' CFLAGS="" CXXFLAGS="--coverage -fprofile-abs-path" LDFLAGS="--coverage -fprofile-abs-path" DEB_BUILD_PROFILES=python2 binary && cp ../*.deb .
```
qiluo-msft, line 67: >DEB_BUILD_PROFILES [](http://example.com/codeflow?start=170&length=18)

Thanks for the fix!
Do we need to change the build process in buildimage/slave.mk?

qiluo-msft (approved):

.azure-pipelines/build-template.yml
saiarcot895: There would need to be a change in `rules/swss-common.mk`, but `slave.mk` doesn't need to be fixed.

qiluo-msft (dismissed):

qiluo-msft (approved):

PR #556, opened 2021-11-10
[PINS] Add AppDB ACL schema values shared between the P4RT app and P4Orch

bocon13:
Submission containing materials of a third party:
    Copyright Google LLC; Licensed under Apache 2.0

Co-authored-by: Akarsh Gupta <akarshgupta@google.com>
Co-authored-by: Ashish Singh <ashishksingh@google.com>
Co-authored-by: Jay Hu <jiarenhu@google.com>
Co-authored-by: Manali Kumar <manalikumar@google.com>
Co-authored-by: Richard Yu <richardyu@google.com>
Co-authored-by: Robert J. Halstead <rhalstea@google.com>
Co-authored-by: Runming Wu <runmingwu@google.com>
Co-authored-by: Srikishen Pondicherry Shanmugam <kishanps@google.com>
Co-authored-by: Vivek Ramamoorthy <vivekmoorthy@google.com>
Co-authored-by: Yilan Ji <yilanji@google.com>

Signed-off-by: Don Newton don@opennetworking.org

PR #561, opened 2021-11-16
Added sanitizer testing for swss-common unit test.

mint570:


mint570:
@qiluo-msft
This is inspired by the review in https://github.com/Azure/sonic-swss-common/pull/556. Adding sanitizer testing could help in catching certain bugs like memory leak.
This pr is independent from PINS. It does not block the PINS prs.
Please take a look when you have time. Thanks.

tests/Makefile.am
```diff
@@ -10,6 +10,9 @@ endif
 
 CFLAGS_GTEST =
 LDADD_GTEST = -L/usr/src/gtest -lgtest -lgtest_main
+CFLAGS_ASAN = -fsanitize=address -static-libasan
+CFLAGS_TSAN = -fsanitize=thread
+CFLAGS_USAN = -fsanitize=undefined
```
qiluo-msft, line 15: Why not combine them together? Even further, just integrate all the options to the original tests?

mint570: It doesn't seem to work if I combine them.

mint570:
@qiluo-msft 
Looks like I don't have permission to restart the test.
I am not sure why the test got stuck.

tests/Makefile.am
qiluo-msft: Could you share the error message?

mint570: Address sanitizer cannot combine with thread sanitizer: https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html
(Check "-fsanitize=address" or "-fsanitize=thread".)

PR #566, opened 2021-12-01
Add option to suppress logging

TACappleman:
In some circumstances calls are made to query the Redis database without checking if a given key (or key/field pair) exists, and with code to handle the case where it does not. For example many show commands query all possible relevant data, without checking if it exists. In these cases it would be useful to be able to prevent warning logs being written for every missing field, and so we add that option in this PR.

### DIFF
--- a/common/dbinterface.h
+++ b/common/dbinterface.h
@@ -38,9 +38,9 @@ public:
     // Delete all keys which match %pattern from DB
     void delete_all_by_pattern(const std::string& dbName, const std::string& pattern);
     bool exists(const std::string& dbName, const std::string& key);
-    std::shared_ptr<std::string> get(const std::string& dbName, const std::string& hash, const std::string& key, bool blocking = false);
+    std::shared_ptr<std::string> get(const std::string& dbName, const std::string& hash, const std::string& key, bool blocking = false, bool verbose = true);
     bool hexists(const std::string& dbName, const std::string& hash, const std::string& key);
-    std::map<std::string, std::string> get_all(const std::string& dbName, const std::string& hash, bool blocking = false);
+    std::map<std::string, std::string> get_all(const std::string& dbName, const std::string& hash, bool blocking = false, bool verbose = true);
     std::vector<std::string> keys(const std::string& dbName, const char *pattern = "*", bool blocking = false);
     std::pair<int, std::vector<std::string>> scan(const std::string& db_name, int cursor, const char *match, uint32_t count);
     int64_t publish(const std::string& dbName, const std::string& channel, const std::string& message);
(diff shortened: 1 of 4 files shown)


common/sonicv2connector.cpp
```diff
@@ -74,19 +74,19 @@ std::pair<int, std::vector<std::string>> SonicV2Connector_Native::scan(const std
     return m_dbintf.scan(db_name, cursor, match, count);
 }
 
-std::shared_ptr<std::string> SonicV2Connector_Native::get(const std::string& db_name, const std::string& _hash, const std::string& key, bool blocking)
```
qiluo-msft, line 77: >SonicV2Connector_Native [](http://example.com/codeflow?start=29&length=23)

This change is not consistent with the python version you are working on [https://github.com/Azure/sonic-py-swsssdk/pull/115/files](https://github.com/Azure/sonic-py-swsssdk/pull/115/files).

Could you make them aligned?

common/dbinterface.h
```diff
@@ -38,9 +38,9 @@ class DBInterface
     // Delete all keys which match %pattern from DB
     void delete_all_by_pattern(const std::string& dbName, const std::string& pattern);
     bool exists(const std::string& dbName, const std::string& key);
-    std::shared_ptr<std::string> get(const std::string& dbName, const std::string& hash, const std::string& key, bool blocking = false);
+    std::shared_ptr<std::string> get(const std::string& dbName, const std::string& hash, const std::string& key, bool blocking = false, bool verbose = true);
```
qiluo-msft, line 41: >blocking [](http://example.com/codeflow?start=118&length=8)

For python version, `blocking` is always the last parameter. The order changes here.

I think it impacts the usage if caller provide all the parameters in positional fashion.

common/sonicv2connector.cpp
TACappleman: I'm not quite clear what you mean here. Are you referring to the fact that the py-swssdk code doesn't have the blocking parameter, and so I should add that at the same time so they match?

common/dbinterface.h
TACappleman: Don't we want additional parameters at the end, so that if for example someone is currently calling it with e.g. ("db_name", "hash", "key", true) then that will still work?

common/sonicv2connector.cpp
qiluo-msft: You does not change the class `SonicV2Connector` in that PR, but change it here. We need to make two library behavior the same way.

common/dbinterface.h
qiluo-msft: I believe all the existing code has `blocking=True` as parameter. My concern is that the behavior of python and C++ SWIG implementations are different.

qiluo-msft (dismissed):

qiluo-msft (approved):

PR #568, opened 2021-12-17
[TableBase] Make channel name from both table name and database ID

jimmyzhai:
1. To overload member function getChannelName with parameter tag
2. To use dbId as tag and call getChannelName in producer/consumertable and producer/consumerstatetable

### DIFF
--- a/common/notificationproducer.cpp
+++ b/common/notificationproducer.cpp
@@ -1,7 +1,7 @@
 #include "notificationproducer.h"
 
 swss::NotificationProducer::NotificationProducer(swss::DBConnector *db, const std::string &channel):
-    m_db(db), m_channel(channel)
+    m_db(db), m_channel(channel + "@" + db->getDbName())
 {
 }
 
--- a/common/notificationproducer.cpp
+++ b/common/notificationproducer.cpp
@@ -1,7 +1,7 @@
 #include "notificationproducer.h"
 
 swss::NotificationProducer::NotificationProducer(swss::DBConnector *db, const std::string &channel):
-    m_db(db), m_channel(channel + "@" + std::to_string(db->getDbId()))
+    m_db(db), m_channel(channel)
 {
 }
 
--- a/common/notificationproducer.cpp
+++ b/common/notificationproducer.cpp
@@ -1,7 +1,7 @@
 #include "notificationproducer.h"
 
 swss::NotificationProducer::NotificationProducer(swss::DBConnector *db, const std::string &channel):
-    m_db(db), m_channel(channel + "@" + db->getDbName())
+    m_db(db), m_channel(channel + "@" + std::to_string(db->getDbId()))
 {
 }
 
--- a/common/producertable.cpp
+++ b/common/producertable.cpp
@@ -19,7 +19,7 @@ ProducerTable::ProducerTable(DBConnector *db, const string &tableName)
 }
 
 ProducerTable::ProducerTable(RedisPipeline *pipeline, const string &tableName, bool buffered)
-    : TableBase(tableName, SonicDBConfig::getSeparator(pipeline->getDbName()))
+    : TableBase(pipeline->getDbId(), tableName)
     , TableName_KeyValueOpQueues(tableName)
     , m_buffered(buffered)
     , m_pipeowned(false)
--- a/common/producertable.cpp
+++ b/common/producertable.cpp
@@ -19,7 +19,7 @@ ProducerTable::ProducerTable(DBConnector *db, const string &tableName)
 }
 
 ProducerTable::ProducerTable(RedisPipeline *pipeline, const string &tableName, bool buffered)
-    : TableBase(pipeline->getDbId(), tableName)
+    : TableBase(tableName, SonicDBConfig::getSeparator(pipeline->getDbName()))
     , TableName_KeyValueOpQueues(tableName)
     , m_buffered(buffered)
     , m_pipeowned(false)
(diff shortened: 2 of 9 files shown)


common/table.h
```diff
@@ -34,34 +31,18 @@ typedef std::map<std::string,TableMap> TableDump;
 
 class TableBase {
 public:
-#ifndef SWIG
-    __attribute__((deprecated))
```
qiluo-msft, line 38: >deprecated [](http://example.com/codeflow?start=19&length=10)

Please don't remove `deprecated` attribute. The reason is we are using tableName to identify a table. And the deprecated ctor is only used in SWIG, and then used by old test code, which we did not migrate completely.

common/table.h
```diff
@@ -77,12 +58,13 @@ class TableBase {
         return m_tableSeparator;
     }
 
-    std::string getChannelName() { return m_tableName + "_CHANNEL"; }
+    std::string getChannelName() { return m_tableName + "_CHANNEL"
```
qiluo-msft, line 61: >getChannelName [](http://example.com/codeflow?start=16&length=14)

The dbId is a concept out of TableBase class.
Could you add new parameter `dbId`, and do not use a member?

qiluo-msft (changes requested):
As comments

common/table.h
jimmyzhai: Looks `dbId` is widely used to identify a database, and both `Id` and `Name` are primary properties of database.  In database domain hierarchy, table always belongs to one database. To add `dbId` member in class TableBase is meaningful to indicate in which database the table is. In ctor internal, table separator is initialized to the separator of  database `dbId`.

common/table.h
jimmyzhai: As above reply,  I thinking adding `dbId` as a member is much reasonable.

common/table.h
```diff
@@ -34,34 +31,18 @@ typedef std::map<std::string,TableMap> TableDump;
...
-            m_tableSeparator = it->second;
-        }
-        else
-        {
-            SWSS_LOG_NOTICE("Unrecognized database ID. Using default table name separator ('%s')", TABLE_NAME_SEPARATOR_VBAR.c_str());
-            m_tableSeparator = TABLE_NAME_SEPARATOR_VBAR;
-        }
-    }
-
-    TableBase(const std::string &tableName, const std::string &tableSeparator)
```
qiluo-msft, line 57: >TableBase [](http://example.com/codeflow?start=4&length=9)

In the design of `database_config.json`, the mapping is dbName -> (dbId, tableSeparator, dbInstance). So there is possible usage that multiple dbNames mapped to the same dbId, but with different tableSeparator.

This ctor is used by many C++ subclasses. In all use cases, the separator are deduced from dbName. The function `getSeparator(int dbId, const string &netns)` is only used in SWIG and python vstest. In future, the vstest code will not use dbId directly, and this function will be deprecated.

common/table.cpp
```diff
@@ -18,28 +18,14 @@ using json = nlohmann::json;
 const std::string TableBase::TABLE_NAME_SEPARATOR_COLON = ":";
 const std::string TableBase::TABLE_NAME_SEPARATOR_VBAR = "|";
 
-// NOTE: this map is deprecated and will be removed in the future, and new DBs should instead be added to database_config.json
```
qiluo-msft, line 21: >deprecated [](http://example.com/codeflow?start=21&length=10)

As the comment, this logic is used by some application depending on SWIG swsscommon library. As far as I know, some python vstest are still not using database_config.json and use old dbId. To migrate all of them to use dbName instead of dbId is a long effort.

So if you are stick with this solution, please check the vstest failures.

common/table.h
jimmyzhai: Sure, need to consider the case of 'multiple dbNames mapped to the same dbId, but with different tableSeparator'

common/table.cpp
jimmyzhai: I'll use the approach to adding a new overload function of getChannelName with param `tag`.

common/producertable.cpp
```diff
@@ -73,7 +73,7 @@ void ProducerTable::enqueueDbChange(const string &key, const string &value, cons
         "EVALSHA %s 2 %s %s %s %s %s %s",
         m_shaEnque.c_str(),
         getKeyValueOpQueueTableName().c_str(),
-        getChannelName().c_str(),
+        getChannelName(m_pipe->getDbName()).c_str(),
```
qiluo-msft, line 76: >getDbName [](http://example.com/codeflow?start=31&length=9)

Better to use `getDbId()` as a tag. It is possible that a `DBConnector` has empty dbName. And this is true with old vstest code written in python, using SWIG swsscommon.

jimmyzhai: Yes, correct it

common/notificationproducer.cpp
```diff
@@ -1,7 +1,7 @@
 #include "notificationproducer.h"
 
 swss::NotificationProducer::NotificationProducer(swss::DBConnector *db, const std::string &channel):
```
qiluo-msft, line 3: >NotificationProducer [](http://example.com/codeflow?start=28&length=20)

Do you need to change NotificationConsumer accordingly?

qiluo-msft (approved):

common/notificationproducer.cpp
jimmyzhai: Updated. Hold on the change of channel name tagged dbId for Notification Producer/Consumer. Some lua scripts like pfc_restore.lua, pfe_detect are using redis `PUBLISH` command directly with hardcode channel name.

qiluo-msft (approved):

PR #575, opened 2022-01-17
Fix package reference to master branch in BuildDocker stage

liushilongbuaa:


qiluo-msft (approved):

.azure-pipelines/build-docker-sonic-vs-template.yml
```diff
@@ -48,10 +48,10 @@ jobs:
     inputs:
       source: specific
       project: build
-      pipeline: 1
+      pipeline: 142
```
qiluo-msft, line 51: Why change this line, I checked 202012/master using pipeline 1?

liushilongbuaa: 142 is vs platform official build. We will change pipeline 1 to 142 job by job.
pipeline 1 uses cache while 142 don't use cache.

liushilongbuaa: And official build will run every day. Pipeline 1 is triggered by code change.
It will fail if no code change last for a month. Because artifact for pipeline 1 is stored for one month but pipeline 142 will be stored for three months.

qiluo-msft: Then please fix also on other branches (eg. 202012/master/etc) and keep them in sync.

qiluo-msft (approved):

qiluo-msft:
Will override and merge this obvious bug fix, in order to unblock other repo's job.

PR #579, opened 2022-01-26
Add a new constructor to NotificationConsumer class.

oleksandrivantsiv:
Add a new constructor that takes a shared pointer to DBConnector as a parameter.
The constructor will be used when the DBConnector object is shared across
different consumers objects. Or when DBConnector is created exclusively for one
consumer and consumer is responsible for the destruction of DBConnector object.


dgsudharsan (approved):

dgsudharsan (approved):

dgsudharsan (approved):

liat-grozovik:
@qiluo-msft could you please help to review?

liat-grozovik:
@kcudnik , @qiluo-msft kindly reminder to review and signoff.

kcudnik (approved):

qiluo-msft:
>             delete m_subscribe;

if m_db == m_db_shrd, they point to the same object. We should never delete it, right?

---
Refers to: common/notificationconsumer.cpp:45 in 603c458. [](commit_id = 603c4580277228dc60f4b010b897993edb7e877d, deletion_comment = False)

qiluo-msft:
>     delete m_subscribe;

The same question.

---
Refers to: common/notificationconsumer.cpp:54 in 603c458. [](commit_id = 603c4580277228dc60f4b010b897993edb7e877d, deletion_comment = False)

qiluo-msft (changes requested):
As comments

oleksandrivantsiv:
> if m_db == m_db_shrd, they point to the same object. We should never delete it, right?

This PR modifies behavior of **m_db** pointer. **m_db** should never be deleted. It is controlled via **m_db_shrd**
Here you are referring to **m_subscribe** pointer which is an internal member that is controlled by NotificationConsumer. It is completely independent of **m_db** and should always be deleted.

oleksandrivantsiv:
> > ```
> > delete m_subscribe;
> > ```
> 
> The same question.
> 
> Refers to: common/notificationconsumer.cpp:54 in [603c458](https://github.com/Azure/sonic-swss-common/commit/603c4580277228dc60f4b010b897993edb7e877d). [](commit_id = [603c458](https://github.com/Azure/sonic-swss-common/commit/603c4580277228dc60f4b010b897993edb7e877d), deletion_comment = False)

**m_subscribe** is owned by NotificationConsumer and should always be deleted

oleksandrivantsiv:
@qiluo-msft kindly ask to take a look to my answer

common/notificationconsumer.cpp
```diff
@@ -14,6 +14,22 @@ swss::NotificationConsumer::NotificationConsumer(swss::DBConnector *db, const st
     m_db(db),
     m_subscribe(NULL),
     m_channel(channel)
+{
+    init();
+}
+
+swss::NotificationConsumer::NotificationConsumer(std::shared_ptr<swss::DBConnector> db, const std::string &channel, int pri, size_t popBatchSize):
```
qiluo-msft, line 21: >NotificationConsumer [](http://example.com/codeflow?start=28&length=20)

I am wondering why this is needed in common library. You are good to use original constructor as, and it behaves the same.
```
swss::NotificationConsumer nc(dbNtf.get(), "NOTIFICATIONS", 100, (size_t)messages);
```

oleksandrivantsiv:
Closing this PR as we found a better way to trigger address sanitizer to report only memory that is definitely lost.

PR #582, opened 2022-02-15
[VS] Increase test timout to 360min

liushilongbuaa:
Changing the docker pool, VS test is taking slightly more than 4hrs.


diff.log
```diff
@@ -0,0 +1,2317 @@
+diff --git a/.azure-pipelines/build-docker-sonic-vs-template.yml b/.azure-pipelines/build-docker-sonic-vs-template.yml
```
prsunny, line 1: Please remove this file

liushilongbuaa: fixed. It's strange that file will impact vstest.

qiluo-msft (approved):

PR #584, opened 2022-03-02
Improve SonicDBConfig initialize multiple time with same file will throw exception issue.

liuh-80:
When swich sonic-py-common from use swss-sdk to use swss-common, some service crash, for example snmp agent.
This is because those the code of those service initialize SonicDBConfig multiple times in their code and sonic-py-common.

So add new data member to SonicDBConfig, when config already initialized with same file, not throw exception.

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -80,6 +80,12 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
 
     if (m_global_init)
     {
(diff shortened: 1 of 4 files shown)


liuh-80:
Will add UT to cover the multiple initialize with same file case.

liuh-80:
After discussion offline, we found the behavior is by design, a DB should never initialize multiple times. will abandon this PR.

PR #586, opened 2022-03-04
[ConfigDBPipeConnector]: Added set_entry API

nazariig:
Signed-off-by: Nazarii Hnydyn <nazariig@nvidia.com>

**What I did**
* Reworked `set_entry` to support Redis transactions
* Added PBH Edit Flows State DB schema

**Why I did it**
* Implementation is done according to the PBH HLD

**How I verified it**
* N/A

**Details if related**
* N/A


qiluo-msft:
Suggest change title. This change is a general feature, no need to mention PBH. "Edit Flows" is also not clear to readers.

Checking https://dev.azure.com/mssonic/build/_build/results?buildId=78095&view=codecoverage-tab, could you also add unit test to cover corner case branch?

nazariig:
> Suggest change title. This change is a general feature, no need to mention PBH. "Edit Flows" is also not clear to readers.
> 
> Checking https://dev.azure.com/mssonic/build/_build/results?buildId=78095&view=codecoverage-tab, could you also add unit test to cover corner case branch?

@qiluo-msft review comments are handled

nazariig:
@anish-n we need to have this PR merged before moving forward with swss changes. Please review it with a high prio.

nazariig:
@anish-n just a kind reminder

common/configdb.h
```diff
@@ -17,8 +17,8 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
     void db_connect(std::string db_name, bool wait_for_init = false, bool retry_on = false);
     void connect(bool wait_for_init = true, bool retry_on = false);
 
-    void set_entry(std::string table, std::string key, const std::map<std::string, std::string>& data);
-    void mod_entry(std::string table, std::string key, const std::map<std::string, std::string>& data);
+    virtual void set_entry(std::string table, std::string key, const std::map<std::string, std::string>& data);
+    virtual void mod_entry(std::string table, std::string key, const std::map<std::string, std::string>& data);
```
anish-n, line 21: I don't see this implemented in the cpp file as a new change, is that expected?

common/configdb.cpp
```diff
@@ -326,6 +326,63 @@ void ConfigDBPipeConnector_Native::_delete_table(DBConnector& client, RedisTrans
...
 }
 
+// Write a table entry to config db
+//    Remove extra fields in the db which are not in the data
+// Args:
+//     table: Table name
+//     key: Key of table entry, or a tuple of keys if it is a multi-key table
+//     data: Table row data in a form of dictionary {'column_key': 'value', ...}
+//           Pass {} as data will delete the entry
+void ConfigDBPipeConnector_Native::set_entry(string table, string key, const map<string, string>& data)
```
anish-n, line 336: Is ConfigDBPipeConnector_Native vs ConfigDBConnector_Native sufficient to imply that this set_entry function uses redis transactions and the ConfigDBConnector_Native::set_entry does not use transactions. I am wondering if its better to name this ConfigDBPipeConnector_Native::set_entry as ConfigDBPipeConnector_Native::set_entry_transactional

common/configdb.cpp
```diff
@@ -326,6 +326,63 @@ void ConfigDBPipeConnector_Native::_delete_table(DBConnector& client, RedisTrans
...
+{
+    auto& client = get_redis_client(m_db_name);
+    DBConnector clientPipe(client);
+    RedisTransactioner pipe(&clientPipe);
+
+    pipe.multi();
+    _set_entry(pipe, table, key, data);
+    pipe.exec();
```
anish-n, line 344: Do we need error handling. I was looking through the redis docs https://redis.io/topics/transactions it says that exec can fail on any of the individual commands in multi/exec, and its upto the client to provide the error back to the user.

common/configdb.h
nazariig: @anish-n in `ConfigDBPipeConnector` we have a different behaviour so basically we do override the base class implementation:
```cpp
void set_entry(std::string table, std::string key, const std::map<std::string, std::string>& data) override;
void mod_config(const std::map<std::string, std::map<std::string, std::map<std::string, std::string>>>& data) override;
```
P.S: no need to have `override` in `.cpp` file

common/configdb.cpp
nazariig: @anish-n basically `ConfigDBPipeConnector` is an extension of `ConfigDBConnector`. IMHO, it seems reasonable to have the same API but with a different behaviour (transactional support). Ideally, each API here should be reworked with transactional approach. That's the original aim to have this separate type of class. So far only `set_entry` has been reworked.

common/configdb.cpp
nazariig: @anish-n the short answer is no. Here, i was following the existing approach which doesn't assume using some sort of intermediate error handling. Also, i don't think that we even have an infra for that (i.e., reporting errors back to user). 

https://github.com/Azure/sonic-swss-common/blob/master/common/configdb.cpp#L363
```cpp
// Write multiple tables into config db.
//    Extra entries/fields in the db which are not in the data are kept.
// Args:
//     data: config data in a dictionary form
//     {
//         'TABLE_NAME': { 'row_key': {'column_key': 'value', ...}, ...},
//         'MULTI_KEY_TABLE_NAME': { ('l1_key', 'l2_key', ...) : {'column_key': 'value', ...}, ...},
//         ...
//     }
void ConfigDBPipeConnector_Native::mod_config(const map<string, map<string, map<string, string>>>& data)
{
    auto& client = get_redis_client(m_db_name);
    DBConnector clientPipe(client);
    RedisTransactioner pipe(&clientPipe);
    pipe.multi();
    for (auto const& id: data)
    {
        auto& table_name = id.first;
        auto& table_data = id.second;
        if (table_data.empty())
        {
            _delete_table(client, pipe, table_name);
            continue;
        }
        for (auto const& it: table_data)
        {
            auto& key = it.first;
            _mod_entry(pipe, table_name, key, it.second);
        }
    }
    pipe.exec();
}
```

anish-n: Ok I see, if this is not something we generally see or need to incorporate for then I am good.

common/configdb.cpp
anish-n: Is ConfigDBPipeConnector  designed to be a transactional implementation itself, ie if someone were to use ConfigDBPipeConnector  would it automatically imply they are want transactional behavior?

common/configdb.h
anish-n: Ok understand, thanks

anish-n:
I have reviewed the PR and provided feedback as per my understanding, since I haven't contributed to configdb.cpp/h in the past, request that you get review from someone who typically contributes to these files as well! :)

qiluo-msft (approved):

common/configdb.cpp
nazariig: @anish-n looking at the `mod_config` implementation in `ConfigDBPipeConnector` i am pretty confident that pipe stands for transactions support:
https://github.com/Azure/sonic-swss-common/blob/master/common/configdb.cpp#L363
```bash
// Write multiple tables into config db.
//    Extra entries/fields in the db which are not in the data are kept.
// Args:
//     data: config data in a dictionary form
//     {
//         'TABLE_NAME': { 'row_key': {'column_key': 'value', ...}, ...},
//         'MULTI_KEY_TABLE_NAME': { ('l1_key', 'l2_key', ...) : {'column_key': 'value', ...}, ...},
//         ...
//     }
void ConfigDBPipeConnector_Native::mod_config(const map<string, map<string, map<string, string>>>& data)
{
    auto& client = get_redis_client(m_db_name);
    DBConnector clientPipe(client);
    RedisTransactioner pipe(&clientPipe);
    pipe.multi();
    for (auto const& id: data)
    {
        auto& table_name = id.first;
        auto& table_data = id.second;
        if (table_data.empty())
        {
            _delete_table(client, pipe, table_name);
            continue;
        }
        for (auto const& it: table_data)
        {
            auto& key = it.first;
            _mod_entry(pipe, table_name, key, it.second);
        }
    }
    pipe.exec();
}
```

nazariig:
> I have reviewed the PR and provided feedback as per my understanding, since I haven't contributed to configdb.cpp/h in the past, request that you get review from someone who typically contributes to these files as well! :)

@anish-n thanks a lot for your feedback! Your comments are very valuable - a lot of interesting questions were answered.

PR #587, opened 2022-03-07
Add API endpoints to ConfigDBConnector to support pre-loading data without blackout

alexrallen:
Currently if someone wishes to use ConfigDBConnector to operate on updates coming from CONFIG_DB AND they wish to pull all the initial data from the table to initialize their only option is to pull the data and then run the `listen()` method on ConfigDBConnector.

This creates a "blackout" period where between the time the tables are downloaded and the `listen()` method is called, any new updates to CONFIG_DB will not be handled.

To resolve this I have changed two things in ConfigDBConnector. All of which are 100% backwards compatible...
1. I split `listen()` into `listen()` and `process()` where `listen()` may now be called with `start=False` as an argument which does not start immediately processing updates from CONFIG_DB and calling handlers. This allows you to call `listen()` *first* and then download table data and then call `process()` to start processing updates. This way no updates will be missed as they will be queued by redis while the table data is being downloaded by the consumer. 
2. I added a `cache` argument to `process` in which you pass any initial table data your system is operating on. When a table update is processed the system checks if that keys value is different than what is stored in the cache (if it is present) and only fires the callback handler in the case that the data has changes (added / deleted / updated). This prevents the same data from being processed multiple times.

### DIFF
--- a/common/configdb.h
+++ b/common/configdb.h
@@ -57,6 +57,7 @@ protected:
 
             ## Note: callback is difficult to implement by SWIG C++, so keep in python
             self.handlers = {}
+            self.fire_init_data = {}
 
         @property
         def KEY_SEPARATOR(self):
@@ -71,15 +72,27 @@ protected:
             return self.getDbName()
 
         ## Note: callback is difficult to implement by SWIG C++, so keep in python
-        def listen(self, init=None):
+        def listen(self, init_data_handler=None):
             ## Start listen Redis keyspace event. Pass a callback function to `init` to handle initial table data.
             self.pubsub = self.get_redis_client(self.db_name).pubsub()
             self.pubsub.psubscribe("__keyspace@{}__:*".format(self.get_dbid(self.db_name)))
+        
+            # Build a cache of data for all subscribed tables that will recieve the initial table data so we dont send duplicate event notifications
+            init_data = {tbl: self.get_table(tbl) for tbl in self.handlers if init_data_handler or self.fire_init_data[tbl]}
 
-            init_data = {}
-            if init:
-                init_data = {tbl: self.get_table(tbl) for tbl, cb in self.handlers.items()}
-                init(init_data)
+            # Function to send initial data as series of updates through individual table callback handlers
+            def load_data(tbl, data):
+                if self.fire_init_data[tbl]:
+                    for row, x in data.items():
+                        self.__fire(tbl, row, x)
+                    return False
+                return True
+
+            init_callback_data = {tbl: data for tbl, data in init_data.items() if load_data(tbl, data)}	
+
+            # Pass all initial data that we DID NOT send as updates to handlers through the init callback if provided by caller
+            if init_data_handler:
+                init_data_handler(init_callback_data)
 
             while True:
                 item = self.pubsub.listen_message()
@@ -164,8 +177,9 @@ protected:
                 handler = self.handlers[table]
                 handler(table, key, data)
 
-        def subscribe(self, table, handler):
+        def subscribe(self, table, handler, fire_init_data=False):
             self.handlers[table] = handler
+            self.fire_init_data[table] = fire_init_data
 
         def unsubscribe(self, table):
             if table in self.handlers:
--- a/common/configdb.h
+++ b/common/configdb.h
@@ -71,16 +71,16 @@ protected:
             return self.getDbName()
 
         ## Note: callback is difficult to implement by SWIG C++, so keep in python
-        def listen(self, start=True):
-            ## Start listen Redis keyspace event. Use start=False if you need to load in initial data to prevent blackout.
-            ## Then call process(cache) where cache is a dict of {table_name: data} of the initialized data to prevent
-            ## duplicate calls to the callback if the data has not changed. 
+        def listen(self, init=None):
+            ## Start listen Redis keyspace event. Pass a callback function to `init` to handle initial table data.
             self.pubsub = self.get_redis_client(self.db_name).pubsub()
             self.pubsub.psubscribe("__keyspace@{}__:*".format(self.get_dbid(self.db_name)))
 
-            if start: self.process()
+            init_data = {}
+            if init:
+                init_data = {tbl: self.get_table(tbl) for tbl, cb in self.handlers.items()}
+                init(init_data)
 
-        def process(self, cache={}):
             while True:
                 item = self.pubsub.listen_message()
                 if item['type'] == 'pmessage':
@@ -90,11 +90,11 @@ protected:
                         if table in self.handlers:
                             client = self.get_redis_client(self.db_name)
                             data = self.raw_to_typed(client.hgetall(key))
-                            if table in cache and row in cache[table]:
-                                if cache[table][row] == data:
+                            if table in init_data and row in init_data[table]:
+                                if init_data[table][row] == data:
                                     continue
                                 else:
-                                    del cache[table][row]
+                                    del init_data[table][row]
                             self.__fire(table, row, data)
                     except ValueError:
                         pass    #Ignore non table-formated redis entries
--- a/common/configdb.h
+++ b/common/configdb.h
@@ -72,7 +72,9 @@ protected:
 
         ## Note: callback is difficult to implement by SWIG C++, so keep in python
         def listen(self, start=True):
-            ## Start listen Redis keyspace events and will trigger corresponding handlers when content of a table changes.
+            ## Start listen Redis keyspace event. Use start=False if you need to load in initial data to prevent blackout.
+            ## Then call process(cache) where cache is a dict of {table_name: data} of the initialized data to prevent
+            ## duplicate calls to the callback if the data has not changed. 
             self.pubsub = self.get_redis_client(self.db_name).pubsub()
             self.pubsub.psubscribe("__keyspace@{}__:*".format(self.get_dbid(self.db_name)))
 
--- a/common/configdb.h
+++ b/common/configdb.h
@@ -71,10 +71,14 @@ protected:
             return self.getDbName()
 
         ## Note: callback is difficult to implement by SWIG C++, so keep in python
-        def listen(self):
+        def listen(self, start=True):
             ## Start listen Redis keyspace events and will trigger corresponding handlers when content of a table changes.
             self.pubsub = self.get_redis_client(self.db_name).pubsub()
             self.pubsub.psubscribe("__keyspace@{}__:*".format(self.get_dbid(self.db_name)))
+
+            if start: self.process()
+
+        def process(self, cache={}):
             while True:
                 item = self.pubsub.listen_message()
                 if item['type'] == 'pmessage':
@@ -84,6 +88,11 @@ protected:
                         if table in self.handlers:
                             client = self.get_redis_client(self.db_name)
                             data = self.raw_to_typed(client.hgetall(key))
+                            if table in cache and row in cache[table]:
+                                if cache[table][row] == data:
+                                    continue
+                                else:
+                                    del cache[table][row]
                             self.__fire(table, row, data)
                     except ValueError:
                         pass    #Ignore non table-formated redis entries
--- a/common/configdb.h
+++ b/common/configdb.h
@@ -106,8 +106,9 @@ protected:
                             if table in init_data and row in init_data[table]:
                                 if init_data[table][row] == data:
                                     continue
-                                else:
-                                    del init_data[table][row]
+                                del init_data[table][row]
+                                if not init_data[table]:
+                                    del init_data[table]
                             self.__fire(table, row, data)
                     except ValueError:
                         pass    #Ignore non table-formated redis entries
--- a/common/configdb.h
+++ b/common/configdb.h
@@ -104,11 +104,13 @@ protected:
                             client = self.get_redis_client(self.db_name)
                             data = self.raw_to_typed(client.hgetall(key))
                             if table in init_data and row in init_data[table]:
+                                cache_hit = False
                                 if init_data[table][row] == data:
-                                    continue
+                                    cache_hit = True
                                 del init_data[table][row]
                                 if not init_data[table]:
                                     del init_data[table]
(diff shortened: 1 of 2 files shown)


dgsudharsan (approved):

renukamanavalan (dismissed):

common/configdb.h
```diff
@@ -71,10 +71,16 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
...
+        def listen(self, start=True):
+            ## Start listen Redis keyspace event. Use start=False if you need to load in initial data to prevent blackout.
+            ## Then call process(cache) where cache is a dict of {table_name: data} of the initialized data to prevent
+            ## duplicate calls to the callback if the data has not changed. 
             self.pubsub = self.get_redis_client(self.db_name).pubsub()
             self.pubsub.psubscribe("__keyspace@{}__:*".format(self.get_dbid(self.db_name)))
+
+            if start: self.process()
+
+        def process(self, cache={}):
```
qiluo-msft, line 83: >process [](http://example.com/codeflow?start=12&length=7)

Please add new test case for the new workflow.

qiluo-msft, line 83: >{} [](http://example.com/codeflow?start=32&length=2)

Do not use mutable default argument in python: [https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)

common/configdb.h
```diff
@@ -84,6 +90,11 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
                         if table in self.handlers:
                             client = self.get_redis_client(self.db_name)
                             data = self.raw_to_typed(client.hgetall(key))
+                            if table in cache and row in cache[table]:
```
qiluo-msft, line 93: >if table in cache and row in cache[table]: [](http://example.com/codeflow?start=28&length=42)

If there is no pmessage coming at all, the cache has no chance to fire. I think you expect them to fire.

common/configdb.h
```diff
@@ -71,10 +71,16 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
             return self.getDbName()
 
         ## Note: callback is difficult to implement by SWIG C++, so keep in python
-        def listen(self):
-            ## Start listen Redis keyspace events and will trigger corresponding handlers when content of a table changes.
+        def listen(self, start=True):
+            ## Start listen Redis keyspace event. Use start=False if you need to load in initial data to prevent blackout.
```
qiluo-msft, line 75: >start=False if you need to load in initial data to prevent blackout [](http://example.com/codeflow?start=54&length=67)

The statement is different from your code. You just do not process if start==False.

common/configdb.h
```diff
@@ -84,6 +90,11 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
                         if table in self.handlers:
                             client = self.get_redis_client(self.db_name)
                             data = self.raw_to_typed(client.hgetall(key))
+                            if table in cache and row in cache[table]:
+                                if cache[table][row] == data:
```
qiluo-msft, line 94: >cache [](http://example.com/codeflow?start=35&length=5)

cache is not a proper name.

common/configdb.h
```diff
@@ -71,10 +71,16 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
             return self.getDbName()
 
         ## Note: callback is difficult to implement by SWIG C++, so keep in python
-        def listen(self):
-            ## Start listen Redis keyspace events and will trigger corresponding handlers when content of a table changes.
+        def listen(self, start=True):
+            ## Start listen Redis keyspace event. Use start=False if you need to load in initial data to prevent blackout.
+            ## Then call process(cache) where cache is a dict of {table_name: data} of the initialized data to prevent
```
qiluo-msft, line 76: >prevent [](http://example.com/codeflow?start=111&length=7)

The purpose is not to "prevent duplicate calls"

common/configdb.h
```diff
@@ -71,10 +71,16 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
             return self.getDbName()
 
         ## Note: callback is difficult to implement by SWIG C++, so keep in python
-        def listen(self):
-            ## Start listen Redis keyspace events and will trigger corresponding handlers when content of a table changes.
+        def listen(self, start=True):
```
qiluo-msft, line 74: >listen [](http://example.com/codeflow?start=12&length=6)

This solution does not solve the blackout issue. As long as you initialize the `cache` before `psubscribe`, you will have a short period of time which you may miss some events.

I understand you want to make the time extremely short, but it is still there.

Check the code in `SubscriberStateTable::SubscriberStateTable()`, it fetches table immediately after `psubscribe`. If there is any events happened in between, it is like duplicated on `m_buffer`, which is safe.

common/configdb.h
qiluo-msft, line 83: >cache [](http://example.com/codeflow?start=26&length=5)

Suggest not to expose `cache` to caller, and use it as private member. The initial `cache` and future events should be processed in the same way. If caller want to differentiate (hostcfgd is an example, which is weird practice) should treat the first `fire` with its `load` function.

alexrallen: I can fix this but the problem will never exist here as the cache is never appended to so an empty default won't leak across calls.

common/configdb.h
alexrallen: The purpose of the cache is to prevent duplicate calls. 

The caller may pass in any initialized data it has back into this system in order to prevent any updates which happened in the time period between calling `listen()` and getting the initial table data from firing a redundant callback.

common/configdb.h
alexrallen: This solution does solve the blackout issue.

The cache is initialized *after* the `psubscribe` the flow of the caller is the following:

1. Call `listen`
2. Get table data
3. Operate on initial table data
4. Call `process()` with initial table data
5. Callbacks are issued when updates which differ from the initial table data are received from redis

There is no blackout period here. In fact the issue we are *now* faced with is one of potentially redundant callbacks if the CONFIG_DB has an update between the time that `listen` is called and that the caller gets the initial table data, which the cache is used to resolve (we can easily prevent duplicate calls, we cannot reconstruct them if we were not listening to them in the first place so this scenario is clearly better).

common/configdb.h
alexrallen: The cache must be exposed to the caller, please see my other comments on the usage of `cache`. This is a variable provided by the caller to inform ConfigDBConnector of the *callers* understanding of the state of CONFIG_DB in order to optimize the callbacks delivered to it. 

In this design the ConfigDBConnector is not providing the initial table data, the caller is now responsible for that. 

I chose this route because it is much more flexible for the callers to be allowed to handle their initialization how they need (in this case if I have a single initial callback to hostcfgd for each table I will not be able to utilize some of the existing optimization as multiple table updates often only need one `systemctl restart` to commit the changes.)

With this in mind, I believe `cache` is the appropriate term for this information. At each update the validity of the cache is checked and if the data in the cache is stale then a callback to update the caller is fired and the stale data is removed from the cache. 

If you feel strongly I can call it `initdata` as I think that is valid as well although I don't think it implies how the data is used as well as `cache` does.

common/configdb.h
qiluo-msft: Thanks for explaining the flow! I double checked your hostcfgd [PR](https://github.com/Azure/sonic-buildimage/pull/10168), and agree with your analysis.

common/configdb.h
```diff
@@ -71,10 +71,16 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
...
         ## Note: callback is difficult to implement by SWIG C++, so keep in python
-        def listen(self):
-            ## Start listen Redis keyspace events and will trigger corresponding handlers when content of a table changes.
+        def listen(self, init=None):
+            ## Start listen Redis keyspace event. Pass a callback function to `init` to handle initial table data.
             self.pubsub = self.get_redis_client(self.db_name).pubsub()
             self.pubsub.psubscribe("__keyspace@{}__:*".format(self.get_dbid(self.db_name)))
+
+            init_data = {}
+            if init:
```
qiluo-msft, line 80: >init [](http://example.com/codeflow?start=15&length=4)

Add a `else` branch, and `__fire` the init_data similar to coming pmessage.

common/configdb.h
```diff
@@ -84,6 +90,11 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
                         if table in self.handlers:
                             client = self.get_redis_client(self.db_name)
                             data = self.raw_to_typed(client.hgetall(key))
+                            if table in init_data and row in init_data[table]:
+                                if init_data[table][row] == data:
+                                    continue
+                                else:
+                                    del init_data[table][row]
```
qiluo-msft, line 97: I understand you are trying to prevent duplicated callback due to the small period of time between `psubscribe` and `get_table`.

1. I am thinking it is safe to duplicate them.
2. There may be other use case: after that period of time, client code write the same value to the existing table/row just for event purpose. We don't want to the this logic to skip them.

So suggest remove this block.

common/configdb.h
```diff
@@ -71,10 +71,16 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
             return self.getDbName()
 
         ## Note: callback is difficult to implement by SWIG C++, so keep in python
-        def listen(self):
-            ## Start listen Redis keyspace events and will trigger corresponding handlers when content of a table changes.
+        def listen(self, init=None):
```
qiluo-msft, line 74: >init [](http://example.com/codeflow?start=25&length=4)

rename to `init_data_handler` or `init_data_loader` ?

common/configdb.h
qiluo-msft: Add new test case in tests/test_redis_ut.py ?

alexrallen: I'll add a test checking the init workflow.

common/configdb.h
```diff
@@ -71,10 +71,16 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
...
+            ## Start listen Redis keyspace event. Pass a callback function to `init` to handle initial table data.
             self.pubsub = self.get_redis_client(self.db_name).pubsub()
             self.pubsub.psubscribe("__keyspace@{}__:*".format(self.get_dbid(self.db_name)))
+
+            init_data = {}
+            if init:
+                init_data = {tbl: self.get_table(tbl) for tbl, cb in self.handlers.items()}
```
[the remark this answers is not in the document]
alexrallen: Sure.

common/configdb.h
alexrallen: I can do this but I only want to if the caller subscribes with some form of `init` argument as well. I do not want to change the default behavior of this library for backwards compatibility reasons.

qiluo-msft: Could you explain "the caller subscribes with some form of init argument as well"?

common/configdb.h
alexrallen: > I understand you are trying to prevent duplicated callback due to the small period of time between `psubscribe` and `get_table`.
> 
>     1. I am thinking it is safe to duplicate them.

It may be safe but is is inefficient and potentially unpredictable. A user will not expect they may receive duplicate callbacks, this is not intuitive and could easily lead to bugs. I don't see a problem with this code so I don't see an argument to remove it. 

>     2. There may be other use case: after that period of time, client code write the same value to the existing table/row just for event purpose. We don't want to the this logic to skip them.

Strongly disagree here. Can you show me anywhere in SONiC where this functionality is utilized? This feels like an anti-pattern and undefined behavior. 

If a developer is motivated to re-submit requests to redis for the purpose of firing events then we most likely have an architectural problem somewhere else that is triggering that desire. 
 
> So suggest remove this block.

common/configdb.h
alexrallen: Sure `init_data_handler` is fine

common/configdb.h
alexrallen:
```
self.config_db.subscribe('AAA', make_callback(self.aaa_handler))
```
This is default behavior.

```
self.config_db.subscribe('AAA', make_callback(self.aaa_handler), handle_init=True)
```
This is behavior where initial data comes through init callbacks. 

We can call `handle_init` anything but do you see where I am going with this?

qiluo-msft: Your proposal is even more fancy. I am okay with either one.

common/configdb.h
qiluo-msft: I see your point. If your intention is to prevent duplicated callback due to the small period of time between `psubscribe` and `get_table`, let's only enable this block only for that purpose. So I suggest:

1. Not matter `init_data[table][row] == data` or not equal, we can delete the initial data. So future events will be not impact by this block
2. After `del init_data[table][row]`, if the table is empty, we can further delete the table from `init_data`.

alexrallen: Sure.

qiluo-msft (approved):

tests/test_redis_ut.py
```diff
@@ -552,6 +553,50 @@ def thread_coming_entry():
...
+
+        config_db.listen(init_data_handler=lambda data: test_init_handler(data, ret))
+
+    config_db = ConfigDBConnector()
+    config_db.connect(wait_for_init=False)
+    client = config_db.get_redis_client(config_db.CONFIG_DB)
+    client.flushdb()
+
+    # Init table data
+    config_db.set_entry(table_name_1, test_key, test_data_1)
```
qiluo-msft, line 589: >test_data_1 [](http://example.com/codeflow?start=48&length=11)

Please check unit test failure:
NameError: global name 'test_data_1' is not defined

alexrallen:
@qiluo-msft Checkers are passing please take a look. Thanks.

dgsudharsan (approved):

dgsudharsan:
@qiluo-msft  Can you please help to merge this PR?

qiluo-msft (approved):

PR #589, opened 2022-03-18
Populating m_dbName in dbConnector constructor.

ntoorchi:
Currently getDbName() returns back an empty string as m_dbName is not populated when constructing dbConnector based on dbId. Here SonicDBConfig::getDbName method is added to find db name based on dbId and populating m_dbName in dbConnector constructor. 

Signed-off-by: Niloofar Toorchi niloofar@opennetworking.org


ntoorchi:
@bocon13 @rhalstea

common/dbconnector.cpp
```diff
@@ -287,6 +287,41 @@ int SonicDBConfig::getDbId(const string &dbName, const string &netns)
     return getDbInfo(dbName, netns).dbId;
 }
 
+string SonicDBConfig::getDbName(int dbId, const string &netns)
```
bocon13, line 290: This seems like a mostly copy-paste of `SonicDBConfig::getSeparator`. Can we refactor the two methods to call a common `getDbInfo` method that contains the common code?

bocon13 (commented):
Overall, this looks good to me. Thanks for putting it together!

ntoorchi:
@qiluo-msft please review

common/dbconnector.cpp
```diff
@@ -510,6 +529,7 @@ void DBConnector::select(DBConnector *db)
 DBConnector::DBConnector(const DBConnector &other)
     : RedisContext(other)
     , m_dbId(other.m_dbId)
+    , m_dbName(other.m_dbName)
```
bocon13, line 532:
```suggestion
    , m_dbName(other.m_dbName, other.m_namespace)
```

Do we want the namespace too?

common/dbconnector.cpp
```diff
@@ -209,7 +209,7 @@ void SonicDBConfig::validateNamespace(const string &netns)
     }
 }
 
-SonicDBInfo& SonicDBConfig::getDbInfo(const std::string &dbName, const std::string &netns)
+std::unordered_map<std::string, SonicDBInfo> SonicDBConfig::getDbEntry(const std::string &netns)
```
bocon13, line 212:
```suggestion
const std::unordered_map<std::string, SonicDBInfo>& SonicDBConfig::getDbEntry(const std::string &netns)
```

We need to avoid copying the map.

common/dbconnector.cpp
```diff
@@ -232,7 +232,12 @@ SonicDBInfo& SonicDBConfig::getDbInfo(const std::string &dbName, const std::stri
         SWSS_LOG_ERROR("%s", msg.c_str());
         throw out_of_range(msg);
     }
-    auto& infos = foundNetns->second;
+    return foundNetns->second;
+}
+
+SonicDBInfo& SonicDBConfig::getDbInfo(const std::string &dbName, const std::string &netns)
+{
+    auto infos = getDbEntry(netns);
```
bocon13, line 240:
```suggestion
    auto const& infos = getDbEntry(netns);
```
We want to make sure that we are taking a read-only reference.

common/dbconnector.cpp
```diff
@@ -232,7 +232,12 @@ SonicDBInfo& SonicDBConfig::getDbInfo(const std::string &dbName, const std::stri
         SWSS_LOG_ERROR("%s", msg.c_str());
         throw out_of_range(msg);
     }
-    auto& infos = foundNetns->second;
+    return foundNetns->second;
+}
+
+SonicDBInfo& SonicDBConfig::getDbInfo(const std::string &dbName, const std::string &netns)
+{
```
bocon13, line 239:
```suggestion
{
    std::lock_guard<std::recursive_mutex> guard(m_db_info_mutex);
    
    SWSS_LOG_ENTER();
    
```

Make sure to hold the lock while we are accessing the `m_db_info` structure. Also, the convention is to log entry into the method.

common/dbconnector.cpp
```diff
@@ -287,6 +292,20 @@ int SonicDBConfig::getDbId(const string &dbName, const string &netns)
     return getDbInfo(dbName, netns).dbId;
 }
 
+string SonicDBConfig::getDbName(int dbId, const string &netns)
+{
+    string db_name;
```
bocon13, line 297: Same as above. 
1. hold the log
2. log entry into the function
3. make sure we have a reference to the entry map

common/dbconnector.h
```diff
@@ -96,6 +97,7 @@ class SonicDBConfig
                                     std::unordered_map<std::string, RedisInstInfo> &inst_entry,
                                     std::unordered_map<std::string, SonicDBInfo> &db_entry,
                                     std::unordered_map<int, std::string> &separator_entry);
+    static std::unordered_map<std::string, SonicDBInfo> getDbEntry(const std::string &netns = EMPTY_NAMESPACE);
```
bocon13, line 100: Suggest renaming this to `getDbEntryMap`?

Also make sure the signature matches whatever we put in the .cc file

common/dbconnector.cpp
```diff
@@ -287,6 +292,20 @@ int SonicDBConfig::getDbId(const string &dbName, const string &netns)
     return getDbInfo(dbName, netns).dbId;
 }
 
+string SonicDBConfig::getDbName(int dbId, const string &netns)
```
qiluo-msft, line 295: >getDbName [](http://example.com/codeflow?start=22&length=9)

It is possible that one dbId is mapped to multiple dbName. Like `FLEX_COUNTER_DB` and `PFC_WD_DB` are all 5.

Why we need this function?

ntoorchi: `getDbName(dbId)` is needed to populate `db_name` when creating a `DBConnector` using a `dbId` (`DBConnector::DBConnector(int dbId, const RedisContext& ctx)`), which we call it in pins-infra at `p4rt_app/p4rt.cc`. Otherwise, DBConnector-> getDbName() returns an empty string. We might need to get the db name for logging purposes. 
The majority of dbs have unique ID (as it is expected). Not sure why `FLEX_COUNTER_DB` and `PFC_WD_DB` have the same id, but I think their `DBConnector` can be created through `dbName` instead of `dbId`, and then there is no issue.

PR #591, opened 2022-03-23
Add SAI copy action to ACL schema

daniele-moro:


bocon13:
@prsunny Can you please take a look? This fixes a bug in PINS.

prsunny (approved):

PR #594, opened 2022-03-31
[azp]: Disable optimization when building

theasianpianist:
Signed-off-by: Lawrence Lee <lawlee@microsoft.com>


kcudnik:
why to disable optimizations when building ,and you are actually enabling -O1 so you are enabling them, -O0 is disable, why you need this change/

theasianpianist:
> why to disable optimizations when building ,and you are actually enabling -O1 so you are enabling them, -O0 is disable, why you need this change/

I am trying to check if changing the optimization level affects how code coverage is calculated for SWSS VS tests. -O0 did not improve it, so I'm checking now if -O1 will improve it

PR #595, opened 2022-04-01
[POC] POC code for show default value with yang model

liuh-80:


common/dbconnector.cpp
```diff
@@ -780,7 +785,21 @@ shared_ptr<string> DBConnector::hget(const string &key, const string &field)
 
     if (reply->type == REDIS_REPLY_NIL)
     {
-        return shared_ptr<string>(NULL);
+        if (!withDefaultValue || this->getDbId() != CONFIG_DB)
+        {
+            return shared_ptr<string>(NULL);
+        }
+
+        size_t pos = key.find("|");
```
qiluo-msft, line 793: >"|" [](http://example.com/codeflow?start=30&length=3)

The separator ground truth is from SonicDBConfig. Do not hard code.

common/dbconnector.cpp
```diff
@@ -772,6 +772,11 @@ shared_ptr<string> DBConnector::get(const string &key)
 }
 
 shared_ptr<string> DBConnector::hget(const string &key, const string &field)
+{
+    return hget(key, field, false);
+}
+
+shared_ptr<string> DBConnector::hget(const string &key, const string &field, bool withDefaultValue)
```
qiluo-msft, line 779: >DBConnector [](http://example.com/codeflow?start=19&length=11)

From the design point of view, DBConnector is in the Redis concept domain and it has no knowledge of ConfigDB, table, row and default value. So it has no requirement to engaged into this new feature.

The same applies to DBInterface, SonicV2Connector.

common/configdb.h
```diff
@@ -19,12 +19,24 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
 
     virtual void set_entry(std::string table, std::string key, const std::map<std::string, std::string>& data);
     virtual void mod_entry(std::string table, std::string key, const std::map<std::string, std::string>& data);
+#ifndef SWIG
+    [[deprecated("Please use get_entry(std::string table, std::string key, bool withDefaultValue) instead.")]]
+#endif
     std::map<std::string, std::string> get_entry(std::string table, std::string key);
+    std::map<std::string, std::string> get_entry(std::string table, std::string key, bool withDefaultValue);
```
qiluo-msft, line 26: >withDefaultValue [](http://example.com/codeflow?start=90&length=16)

I see you keeps backward-compatible to help gradually migration. I support the idea.

Instead of customize per function call, I think customize the constructor will be good enough. So the migration is instance by instance.

lguohan:
this is one of key feature. do we have a design doc to review?

liuh-80:
> this is one of key feature. do we have a design doc to review?

Here is design document PR: https://github.com/Azure/SONiC/pull/989/files

common/configdb.h
liuh-80: Will update this PR and design document according to your comments and discussion.

common/dbconnector.cpp
liuh-80: Fixed, now get separator from DBConnector.

common/dbconnector.cpp
liuh-80: Fixed by create a ConfigDBDecorator class and move this code to this class.

common/configdb.h
liuh-80: Fixed, remove all new parameter and revert UT change.

liuh-80:
Build failed because we need libyang, however sonic version of patched libyang is in sonic-buildimage repo.
Will check if can install original libyang in pipeline and pass build.

common/dbconnector.cpp
```diff
@@ -780,7 +783,15 @@ shared_ptr<string> DBConnector::hget(const string &key, const string &field)
 
     if (reply->type == REDIS_REPLY_NIL)
     {
-        return shared_ptr<string>(NULL);
+        auto dbdecortor = this->getDBDecorator();
```
liuh-80, line 786: The reason does not create ConfigDBConnector but add code here is because DBinterface has following data member:

    std::unordered_map<std::string, DBConnector> m_redisClient;

So, create a new derived class ConfigDBConnector need code change in this member and related code.

Will discussion if this is acceptable.

liuh-80:
Close this PR, POC will move to: https://github.com/Azure/sonic-swss-common/pull/625

PR #596, opened 2022-04-02
[ci]: Fix gmock.h not found issue

xumia:
Fix gmock.h not found issue

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -60,6 +60,7 @@ jobs:
         libnl-route-3-dev \
         libnl-nf-3-dev \
         swig3.0
+      sudo apt-get install -y libgmock-dev
     displayName: "Install dependencies"
   - script: |
       set -ex
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -17,7 +17,7 @@ parameters:
 
 - name: timeout
   type: number
-  default: 60
+  default: 180
 
 - name: sonic_slave
   type: string


liat-grozovik (approved):

.azure-pipelines/build-template.yml
```diff
@@ -60,6 +60,7 @@ jobs:
         libnl-route-3-dev \
         libnl-nf-3-dev \
         swig3.0
+      sudo apt-get install -y libgmock-dev
```
qiluo-msft, line 63: It should be already installed in the docker image `sonicdev-microsoft.azurecr.io:443/${{ parameters.sonic_slave }}:latest`. However it seems some wrong image were pushed. Suggest dig into the push log and audit which job is pushing the wrong images.

qiluo-msft (approved):

.azure-pipelines/build-template.yml
xumia: @qiluo-msft , you are right, this PR should be not necessary.

It should be caused by the latest tag added by the slave buster build in 202012 branch which does not contain libgmock-dev. 

https://dev.azure.com/mssonic/build/_build/results?buildId=86989&view=results

See the template:
https://github.com/Azure/sonic-buildimage/blob/e1c36dbfc111dc9dded6453bd33e29458daa6e83/.azure-pipelines/docker-sonic-slave-template.yml#L106

@liushilongbuaa , I have disabled the CI/Cron triggers in the build for release branches.

qiluo-msft:
This PR should be not necessary.

PR #599, opened 2022-04-07
[ci] Parameterize pipeline and improve azure pipeline

liushilongbuaa:
1. Parameterize pipeline. When checkout new release branch, we don't need manually setup azp.
2. Use common lib pipeline instead of buildimage pipeline to get higher success rate.
3. Clean workspace when downloading external packages. This will ensure package version matching.

### DIFF
--- a/.azure-pipelines/build-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/build-docker-sonic-vs-template.yml
@@ -34,7 +34,6 @@ jobs:
   steps:
   - checkout: self
     clean: true
-  - template: template-get-source-branch.yml
   - task: DownloadPipelineArtifact@2
     inputs:
       artifact: ${{ parameters.swss_common_artifact_name }}
@@ -57,7 +56,7 @@ jobs:
       pipeline: Azure.sonic-buildimage.official.vs
       artifact: sonic-buildimage.vs
       runVersion: 'latestFromBranch'
-      runBranch: 'refs/heads/$(SOURCE_BRANCH)'
+      runBranch: 'refs/heads/$(BUILD_BRANCH)'
       path: $(Build.ArtifactStagingDirectory)/download
       patterns: '**/target/docker-sonic-vs.gz'
     displayName: "Download sonic-buildimage docker-sonic-vs"
--- a/.azure-pipelines/build-docker-sonic-vs-template.yml
+++ b/.azure-pipelines/build-docker-sonic-vs-template.yml
@@ -32,35 +32,44 @@ jobs:
     vmImage: 'ubuntu-20.04'
 
   steps:
+  - checkout: self
+    clean: true
+  - template: template-get-source-branch.yml
   - task: DownloadPipelineArtifact@2
     inputs:
       artifact: ${{ parameters.swss_common_artifact_name }}
-    displayName: "Download sonic swss common deb packages"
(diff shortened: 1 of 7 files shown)


.azure-pipelines/build-docker-sonic-vs-template.yml
```diff
@@ -32,35 +32,44 @@ jobs:
...
+      path: $(Build.ArtifactStagingDirectory)/download
+    displayName: "Download pre-stage built ${{ parameters.swss_artifact_name }}"
   - task: DownloadPipelineArtifact@2
     inputs:
       source: specific
       project: build
-      pipeline: 1
+      pipeline: Azure.sonic-buildimage.official.vs
       artifact: sonic-buildimage.vs
       runVersion: 'latestFromBranch'
```
xumia, line 59: If latestFromBranch is specified, we may need to specify the branchName, the default value is refs/heads/master.
See https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/download-build-artifacts?view=azure-devops

.azure-pipelines/test-docker-sonic-vs-template.yml
```diff
@@ -18,29 +18,36 @@ jobs:
       ls -A1 | xargs -I{} sudo rm -rf {}
     displayName: "Clean workspace"
   - checkout: self
+    clean: true
+    displayName: "Checkout sonic-swss-common"
+  - checkout: sonic-swss
```
xumia, line 23: Maybe we can use the Azure template extends to set the correct branch name in the template.
Currently, the repository resource does not access a dynamic branch name, not quite sure if we can move into the template.

.azure-pipelines/build-docker-sonic-vs-template.yml
liushilongbuaa: Default value is ok.
Although remove this line, pipeline still use master branch.

.azure-pipelines/test-docker-sonic-vs-template.yml
liushilongbuaa: resources can't appear after steps. we need to use script to get correct branch name.
So, using extends can't help dynamic branch name in resource.

azure-pipelines.yml
```diff
@@ -2,11 +2,26 @@
 # Build your C/C++ project with GCC using make.
 # Add steps that publish test results, save build artifacts, deploy, and more:
 # https://docs.microsoft.com/azure/devops/pipelines/apps/c-cpp/gcc
+pr:
+- master
+- 202???
+- 201???
```
xumia, line 8: Can we trigger for all branches?

azure-pipelines.yml
```diff
@@ -2,11 +2,26 @@
...
 
 trigger:
+- master
+- 202???
+- 201???
```
xumia, line 13: Can we trigger the ci build for all branches, and add batch=true?

azure-pipelines.yml
liushilongbuaa: There are some branches other than release branch.
It seems make sence to add some filter.

azure-pipelines.yml
liushilongbuaa: Add batch

xumia (approved):

Issue #603, opened 2022-04-16
There is a problem that hostcfgd blocks forever and does not react to SIGTERM causing a delay in warm boot.


qiluo-msft:
There is a problem that hostcfgd blocks forever and does not react to SIGTERM causing a delay in warm boot.
When hitting this line  https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-host-services/scripts/hostcfgd#L1241 .

Example:

```
>>> from swsscommon import swsscommon
>>> c=swsscommon.ConfigDBConnector()
>>> c.subscribe('A', lambda a: None)
>>> c.connect()
>>> c.listen()


^C^C^C^C
```
@qiluo-msft This looks like a swss-common issue.

This wasn't seen for some reason during 500 warm boot tests on 202012 since warm reboot is executed early enough before hostcfgd stucks in a listen().

_Originally posted by @stepanblyschak in https://github.com/Azure/sonic-buildimage/issues/10510#issuecomment-1100055929_

liuh-80:
ACK, will start investigation this issue.

liuh-80:
I found there are 2 place have infinite loop issue:

[std::map<std::string, std::string> PubSub::listen_message()](https://github.com/Azure/sonic-swss-common/blob/15c0f729648ee9f26ac9aea767d5712db34ca298/common/pubsub.cpp#L135)

https://github.com/Azure/sonic-swss-common/blob/36e1f61691df7aa44782a377e4082cc4380f1018/common/configdb.h#L97

We need exit these infinite loop when receive SIGTERM by add a SIGTERM handler and check SIGTERM  status in loop. also we may need check if other signal also need handle, for example SIGKILL

liat-grozovik:
@qiluo-msft any reason the issue is not opened on buildimage for tracking? i believe we only track it and not submodules.

liuh-80:
To minimize code change and make code scalable, my proposal is:
1.Add a CancellationToken class.
2.And add new parameter to these existed API.
3.Code change in hostcfgd to use the new API with a signal handler.

liuh-80:
Here is a draft PR for this issue:

https://github.com/Azure/sonic-swss-common/pull/606

>>> from swsscommon import swsscommon
>>>
>>> global_cancellation_token = swsscommon.SignalHandlerHelper.GetCancellationToken(2)
>>> swsscommon.SignalHandlerHelper.RegisterSignalHandler(2)
>>>
[...]

PR #606, opened 2022-04-19
Fix SIGTERM can't terminate PubSub::listen issue by add cancellation token support.

liuh-80:
#### Why I did it
    There are infinite loops inside PubSub::listen() method, so application using this method can't handle SIGTERM correctly.
    https://github.com/Azure/sonic-swss-common/issues/603

#### How I did it
    Add following class:
    1. CancellationToken: this class will help exist the infinite loops when SIGTERM or other signal happen.
    2. SignalHandlerHelper: Provide a native signal handler.

#### How to verify it
   1. manually test.
   2. Pass all test case.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


common/cancellationtoken.cpp
```diff
@@ -0,0 +1,29 @@
+#include "cancellationtoken.h"
+
+using namespace swss;
+
+CancellationToken::CancellationToken()
+:m_cancled(false)
+{
+}
+
+CancellationToken::~CancellationToken()
```
stepanblyschak, line 10: Suggestion: I am not sure if there is a need to define the destructor. Also, usually when there is an explicit destructor it is better to define copy semantics as well (rule of 3). I suggest to drop the explicit destructor.

common/cancellationtoken.h
```diff
@@ -0,0 +1,18 @@
...
+namespace swss {
+
+class CancellationToken
+{
+public:
+    CancellationToken();
+    ~CancellationToken();
+    bool IsCancled();
+    void Cancel();
+    void Reset();
```
stepanblyschak, line 12: Could you please add "const" and "noexcept" where neccessary.

common/signalhandlerhelper.cpp
```diff
@@ -0,0 +1,28 @@
...
+using namespace swss;
+
+std::map<int, CancellationToken*> SignalHandlerHelper::signalTokenMapping;
+
+void SignalHandlerHelper::RegisterSignalHandler(int signalNumber)
+{
+    signal(signalNumber, SignalHandlerHelper::OnSignal);
+}
+
+void SignalHandlerHelper::RegisterCancellationToken(int signalNumber, CancellationToken& token)
```
stepanblyschak, line 15: It might be easy to missuse this API by passing a token which lives on the stack which will make a pointer ConcellationToken* in the global map invalid. What if the map *owns* CancellationToken and this method will return a reference to it ```auto& token = SignalHandlerHelper::RegisterCancellationToken(2)```?

common/signalhandlerhelper.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "cancellationtoken.h"
+#include "signalhandlerhelper.h"
+
+using namespace swss;
+
+std::map<int, CancellationToken*> SignalHandlerHelper::signalTokenMapping;
+
+void SignalHandlerHelper::RegisterSignalHandler(int signalNumber)
+{
+    signal(signalNumber, SignalHandlerHelper::OnSignal);
```
stepanblyschak, line 12: Does it override the signal handler defined by the application? In case of python, does it mean that the user defined signal handler in python won't be executed?

common/select.cpp
```diff
@@ -96,8 +97,14 @@ int Select::poll_descriptors(Selectable **c, unsigned int timeout)
     do
     {
         ret = ::epoll_wait(m_epoll_fd, events.data(), sz_selectables, timeout);
+
+        // sleep here make python signal handler not be blocked.
+        sleep(0);
```
stepanblyschak, line 102: Could you please explain a bit more about this sleep, how does it make python signal handler not be blocked?

common/select.cpp
```diff
@@ -96,8 +97,14 @@ int Select::poll_descriptors(Selectable **c, unsigned int timeout)
...
     }
-    while(ret == -1 && errno == EINTR); // Retry the select if the process was interrupted by a signal
+    while(ret == -1 && errno == EINTR && !cancellationToken.IsCancled()); // Retry the select if the process was interrupted by a signal
```
stepanblyschak, line 104: If we would return immidiatelly on EINTR instead of retrying the loop would it solve the issue?

common/signalhandlerhelper.cpp
liuh-80: I consider this solution in earlier version, but this solution can only 1 signal for 1 cancellation token, so I can't call  API and make it cancelled by both SIGTERM and SIGINT.
I will try find a better solution for this.

common/signalhandlerhelper.cpp
liuh-80: Yes, because linux can only have 1 signal handler per-application.
If customer need define their signal handler, then customer also need cancel the token when signal happen, this class just a helper class.

common/select.cpp
liuh-80: I will remove this code, it's just a test code, it can't unblock python signal handler.

common/select.cpp
liuh-80: No. because there are 3 infinite loops in this case:
1. a python while loop in ConfigDBConnector::listen
2. a c++ loop in PubSub::listen_message
3. while loop here.

So, if we return here then this method will invoke again by PubSub::listen_message.
Also, I consider about that customer may not want SIGINT can break this API.
So, I add a cancellation token to parameter, and let customer decide the behavior.

common/cancellationtoken.h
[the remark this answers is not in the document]
liuh-80: Fixed.

common/cancellationtoken.cpp
liuh-80: Fixed.

common/cancellationtoken.h
liuh-80: Fixed.

common/signalhandlerhelper.cpp
liuh-80: Fixed:
Code change to use shared_ptr in this map and declare CancellationToken as a shared_ptr in swsscommon.i.

common/select.cpp
liuh-80: Fixed.

common/select.cpp
stepanblyschak: We could introduce a flag to Select::select, PubSub::listen_message and PubSub::get_message, ConfigDBConnector::listen that would tell whether to retry on EINTR or not and by default it will retry on EINTR in order to not break clients code relying on existing behaviour. For those clients that define a signal hanler it would be just a one line change to add a flag to ConfigDBConnector::listen. I find that the fact that python clients won't be able to define their own signal handlers is a limitation and would require a bit more changes to the python clients code.

liuh-80: Fixed, new I change to this solution:
1. code change in C++ part to return when epoll break by any signal.
    This is OK because python side will retry.
    Also, this change will unblock the python default signal handler.
2. when SIGTERM happen, python default signal handler will break the python side retry.

Here is my test result:

>>> from swsscommon import swsscommon
>>>
>>> c=swsscommon.ConfigDBConnector()
>>> c.subscribe('A', lambda a: None)
>>> c.connect()
>>> c.listen(None)
**^C**Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/swsscommon/swsscommon.py", line 318, in <                          lambda>
    __setattr__ = lambda self, name, value: _swig_setattr(self, FieldValueMap, n                          ame, value)
KeyboardInterrupt

The above exception was the direct cause of the following exception:

  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/swsscommon/swsscommon.py", line 2181, in                           listen
    item = self.pubsub.listen_message()
  File "/usr/lib/python3/dist-packages/swsscommon/swsscommon.py", line 1781, in                           listen_message
    return _swsscommon.PubSub_listen_message(self)
SystemError: <built-in function PubSub_listen_message> returned a result with an                           error set
>>>

common/configdb.h
```diff
@@ -96,6 +96,10 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
 
             while True:
                 item = self.pubsub.listen_message()
+                if not item.has_key('type'):
+                    # When timeout or cancled, item will not contains 'type'
```
qiluo-msft, line 100: >cancled [](http://example.com/codeflow?start=38&length=7)

typo

common/configdb.h
```diff
@@ -96,6 +96,10 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
 
             while True:
                 item = self.pubsub.listen_message()
+                if not item.has_key('type'):
```
qiluo-msft, line 99: >if not item.has_key('type') [](http://example.com/codeflow?start=16&length=27)

```recommendation
                if 'type' not in item:
```
Not a big issue, but more readable.

qiluo-msft: Not a big issue, but more readable.

common/select.cpp
```diff
@@ -93,11 +94,9 @@ int Select::poll_descriptors(Selectable **c, unsigned int timeout)
     std::vector<struct epoll_event> events(sz_selectables);
     int ret;
 
-    do
-    {
-        ret = ::epoll_wait(m_epoll_fd, events.data(), sz_selectables, timeout);
-    }
-    while(ret == -1 && errno == EINTR); // Retry the select if the process was interrupted by a signal
```
qiluo-msft, line 100: >// [](http://example.com/codeflow?start=40&length=2)

The comment is good. Let's keep it or maybe tune it for your new code.

qiluo-msft:
>     ret = poll_descriptors(c, timeout);

Did you change old behavior?

---
Refers to: common/select.cpp:166 in 037283c. [](commit_id = 037283c33e84a4a7e1404065a76dfe718db11166, deletion_comment = False)

common/select.h
```diff
@@ -30,6 +30,7 @@ class Select
         OBJECT = 0,
         ERROR = 1,
         TIMEOUT = 2,
+        ERRINTR = 3,
```
qiluo-msft, line 33: >ERRINTR [](http://example.com/codeflow?start=8&length=7)

What is the relation ERROR vs ERRINTR? Need some code comments.

liuh-80:
> > ```
> > ret = poll_descriptors(c, timeout);
> > ```
> 
> Did you change old behavior?
> 
> Refers to: common/select.cpp:166 in [037283c](https://github.com/Azure/sonic-swss-common/commit/037283c33e84a4a7e1404065a76dfe718db11166). [](commit_id = [037283c](https://github.com/Azure/sonic-swss-common/commit/037283c33e84a4a7e1404065a76dfe718db11166), deletion_comment = False)

No, I didn't. when timeout happen, current code still will try read again.

common/select.h
liuh-80: Fixed, rename to SIGNALINTR and add comments.

common/select.cpp
liuh-80: Fixed.

common/configdb.h
liuh-80: Fixed.

common/configdb.h
liuh-80: Fixed

qiluo-msft:
>     ret = poll_descriptors(c, timeout);

For non terminating signals (like siguser), original behavior is keeping while loop, but the new behavior is return SIGNALINTR. I think this is unexpected.

---
In reply to: [1106076032](https://github.com/Azure/sonic-swss-common/pull/606#issuecomment-1106076032) [](http://example.com/codeflow?ancestors=1106076032,1106076032)

---
Refers to: common/select.cpp:166 in 037283c. [](commit_id = 037283c33e84a4a7e1404065a76dfe718db11166, deletion_comment = False)

liuh-80:
> > ```
> > ret = poll_descriptors(c, timeout);
> > ```
> 
> For non terminating signals (like siguser), original behavior is keeping while loop, but the new behavior is return SIGNALINTR. I think this is unexpected.
> 
> In reply to: [1106076032](https://github.com/Azure/sonic-swss-common/pull/606#issuecomment-1106076032)
> 
> Refers to: common/select.cpp:166 in [037283c](https://github.com/Azure/sonic-swss-common/commit/037283c33e84a4a7e1404065a76dfe718db11166). [](commit_id = [037283c](https://github.com/Azure/sonic-swss-common/commit/037283c33e84a4a7e1404065a76dfe718db11166), deletion_comment = False)

Will discussion offline, the reason is:
1. without a customize signal handler we don't know which signal break the epoll_wait.
2. When poll_descriptors break by siguser, it will return to python code and python code will re-run poll_descriptors method, so from the python code side, the behavior not change.

However. there are some UT failed, seems related with this change, will check reason.

liuh-80:
Add SignalHandlerHelper class back because:
1. There are UT break because in last version any signal will break poll_descriptors() method.
2. So the solution should only allow SIGTERM to break poll_descriptors().
3. However, according to linux signal system, it's impossible to check which signal happening when there is no signal handler.

So we have to add this class back.

common/pubsub.cpp
```diff
@@ -79,20 +79,29 @@ bool PubSub::hasCachedData()
...
 map<string, string> PubSub::get_message(double timeout)
 {
-    map<string, string> ret;
+    return get_message_internal(timeout).second;
+}
+
+pair<int, map<string, string> > PubSub::get_message_internal(double timeout)
+{
+    pair<int, map<string, string> > ret;
+    ret.first = Select::OBJECT;
```
qiluo-msft, line 88: This line could be moved into next if-block.

common/pubsub.cpp
```diff
@@ -79,20 +79,29 @@ bool PubSub::hasCachedData()
...
+
     if (!m_subscribe)
     {
         return ret;
```
qiluo-msft, line 92: treat it as Select::ERROR?

common/pubsub.cpp
```diff
@@ -124,12 +133,14 @@ std::map<std::string, std::string> PubSub::listen_message()
     const double GET_MESSAGE_INTERVAL = 600.0; // in seconds
     for (;;)
     {
-        auto ret = get_message(GET_MESSAGE_INTERVAL);
-        if (!ret.empty())
+        auto ret = get_message_internal(GET_MESSAGE_INTERVAL);
+        if (!ret.second.empty() || ret.first == Select::SIGNALINT)
```
qiluo-msft, line 137: >SIGNALINT [](http://example.com/codeflow?start=56&length=9)

The function is to mimic redis-py PubSub.get_message() and only used in python through SWIG.

Could you check the sigint behavior the original one?

common/pubsub.h
```diff
@@ -29,6 +30,7 @@ class PubSub : protected RedisSelect
 private:
     /* Pop keyspace event from event buffer. Caller should free resources. */
     std::shared_ptr<RedisReply> popEventBuffer();
+    std::pair<int, std::map<std::string, std::string> > get_message_internal( double timeout = 0.0);
```
qiluo-msft, line 33: >  [](http://example.com/codeflow?start=77&length=1)

Check code style in similar functions.

common/signalhandlerhelper.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "signalhandlerhelper.h"
+
+using namespace swss;
+
+std::map<int, bool> SignalHandlerHelper::signalStatusMapping;
+
+void SignalHandlerHelper::registerSignalHandler(int signalNumber)
+{
+    signalStatusMapping[signalNumber] = false;
+    signal(signalNumber, SignalHandlerHelper::onSignal);
```
qiluo-msft, line 12: >signal [](http://example.com/codeflow?start=4&length=6)

Use sigaction(). ref: [https://stackoverflow.com/a/232711/2514803](https://stackoverflow.com/a/232711/2514803)

common/select.cpp
```diff
@@ -96,8 +98,17 @@ int Select::poll_descriptors(Selectable **c, unsigned int timeout)
     do
     {
         ret = ::epoll_wait(m_epoll_fd, events.data(), sz_selectables, timeout);
+
+        // sleep here make python signal handler not be blocked.
+        sleep(0);
```
qiluo-msft, line 103: >sleep(0) [](http://example.com/codeflow?start=8&length=8)

Introduce more context switch. Is it really necessary?

common/pubsub.cpp
liuh-80: The sigint behavior is not the original one:
1. before this change, sigint will no break the loop in get_message_internal, so this method will not return when sigint happen.
2. After this change, sigint will break the loop in get_message_internal, and this method will return when sigint happen.

common/pubsub.cpp
liuh-80: Fixed.

common/pubsub.cpp
liuh-80: Fixed.

common/pubsub.h
liuh-80: Fixed with typedef

common/signalhandlerhelper.cpp
liuh-80: Fixed with sigaction.

common/select.cpp
liuh-80: Removed.

common/pubsub.cpp
liuh-80: This change seems cause 4 UT failed, will revert and try again: 
2022-05-10T04:30:38.5956920Z test_route.py::TestRoute::test_RouteAddRemoveIpv4Route FAILED            [ 76%]
2022-05-10T04:31:01.5159070Z test_route.py::TestRoute::test_RouteAddRemoveIpv4RouteUnresolvedNeigh FAILED [ 77%]
2022-05-10T04:31:21.9768409Z test_route.py::TestRoute::test_RouteAddRemoveIpv4RouteWithVrf FAILED     [ 77%]
2022-05-10T04:31:28.2277393Z test_route.py::TestRoute::test_RouteAddRemoveIpv4BlackholeRoute FAILED   [ 78%]

liuh-80:
Code coverage does not reach target, will add UT to cover new code.

common/pubsub.cpp
liuh-80: Fixed.

liuh-80:
@stepanblyschak, could you please check this PR again? all issue fixed.

common/pubsub.h
qiluo-msft: Sorry, I mean the blank before `double`.

common/pubsub.cpp
```diff
@@ -124,12 +133,14 @@ std::map<std::string, std::string> PubSub::listen_message()
...
         {
-            return ret;
+            return ret.second;
         }
     }
+
+    return map<string, string>();
```
qiluo-msft, line 143: Seems code is not reachable, so it will decrease code coverage. How about add `break` into the for-loop and `return`  here?

common/pubsub.cpp
qiluo-msft: My comment is still applicable.

common/signalhandlerhelper.cpp
```diff
@@ -0,0 +1,72 @@
...
+    auto result = m_sigActionMapping.find(signalNumber);
+    if (result != m_sigActionMapping.end())
+    {
+        // signal action already registered
+        SWSS_LOG_WARN("sigaction for %d already registered.", signalNumber);
+        return;
+    }
+
+    m_signalStatusMapping[signalNumber] = false;
+    auto *old_action = new struct sigaction();
```
qiluo-msft, line 21: >new [](http://example.com/codeflow?start=23&length=3)

Possible to prevent new/delete? How about smart pointers?

qiluo-msft:
>     ret = poll_descriptors(c, timeout);

Could you fix for non terminating signals (like siguser)?

---
In reply to: [1109033970](https://github.com/Azure/sonic-swss-common/pull/606#issuecomment-1109033970) [](http://example.com/codeflow?ancestors=1109033970,1106076032)

---
Refers to: common/select.cpp:166 in 037283c. [](commit_id = 037283c33e84a4a7e1404065a76dfe718db11166, deletion_comment = False)

common/signalhandlerhelper.cpp
liuh-80: Fixed

common/pubsub.cpp
liuh-80: Fixed

common/pubsub.h
liuh-80: Fixed

common/pubsub.cpp
liuh-80: Fixed, all UT passed.

liuh-80:
> > ```
> > ret = poll_descriptors(c, timeout);
> > ```
> 
> Could you fix for non terminating signals (like siguser)?
> 
> In reply to: [1109033970](https://github.com/Azure/sonic-swss-common/pull/606#issuecomment-1109033970)
> 
> Refers to: common/select.cpp:166 in [037283c](https://github.com/Azure/sonic-swss-common/commit/037283c33e84a4a7e1404065a76dfe718db11166). [](commit_id = [037283c](https://github.com/Azure/sonic-swss-common/commit/037283c33e84a4a7e1404065a76dfe718db11166), deletion_comment = False)

This already fixed, currently only sigint can break the while loop in select.cpp, all other signal will not break while loop.

qiluo-msft (approved):

PR #614, opened 2022-04-28
selectabletimer: add mutex to start() and stop()

ds952811:
Add mutex protection to avoid having the timer expiration reset due to
subsequent calls of start()

Avoid having the timer expiration reset due to subsequent calls of start()

Add mutex protection to start() and stop()

Signed-off-by: Dante Su <dante.su@broadcom.com>

### DIFF
--- a/common/selectabletimer.cpp
+++ b/common/selectabletimer.cpp
@@ -21,6 +21,7 @@ SelectableTimer::SelectableTimer(const timespec& interval, int pri)
         SWSS_LOG_THROW("failed to create timerfd, errno: %s", strerror(errno));
     }
     setInterval(interval);
+    m_alive = false;
 }
 
 SelectableTimer::~SelectableTimer()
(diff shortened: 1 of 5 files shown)


adyeung:
@zhangyanzhao @yxieca @qiluo-msft pls help signoff and merge

common/selectabletimer.cpp
```diff
@@ -21,6 +21,7 @@ SelectableTimer::SelectableTimer(const timespec& interval, int pri)
         SWSS_LOG_THROW("failed to create timerfd, errno: %s", strerror(errno));
     }
     setInterval(interval);
+    m_alive = false;
```
qiluo-msft, line 24: >m_alive [](http://example.com/codeflow?start=4&length=7)

better name `m_running`.

common/selectabletimer.cpp
```diff
@@ -36,22 +37,40 @@ SelectableTimer::~SelectableTimer()
 
 void SelectableTimer::start()
```
qiluo-msft, line 38: >start [](http://example.com/codeflow?start=22&length=5)

Could you add new test case which will fail with original code?

common/selectabletimer.cpp
ds952811: Done

common/selectabletimer.cpp
ds952811: Done

ds952811: Please refer to timer_ut.cpp: line 52

tests/redis_ut.cpp
```diff
@@ -653,42 +653,6 @@ TEST(DBConnector, selectableevent)
     EXPECT_EQ(value, 2);
 }
 
-TEST(DBConnector, selectabletimer)
```
ds952811, line 656: Moved to timer_ut.cpp

tests/timer_ut.cpp
```diff
@@ -0,0 +1,52 @@
...
+    ASSERT_EQ(result, Select::OBJECT);
+    ASSERT_EQ(sel, &timer);
+
+    // Reset and wait long enough so we got timer notification first
+    timer.reset();
+    result = s.select(&sel, 10000);
+    ASSERT_EQ(result, Select::OBJECT);
+    ASSERT_EQ(sel, &timer);
+
+    // Check if the timer gets reset by subsequent timer.start()
```
qiluo-msft, line 43: >Check if the timer gets reset by subsequent timer.start() [](http://example.com/codeflow?start=7&length=57)

Could you help me understand what is the behavior of original code in this testcase? How does it fail?

ds952811: In the case of original code, since the timer is created with flag=0, a relative timer is started, and everytime that timer.start() gets invoked, timerfd_settime() will reset the timeout based on the current timestamp, hence it will cause unexpected delay here.

qiluo-msft: Do you mean original code will fail on this line?
```
ASSERT_EQ(result, Select::OBJECT);
```

qiluo-msft (approved):

tests/timer_ut.cpp
ds952811: Yes, it'll fail at that line without the changes in the PR

PR #632, opened 2022-06-14
[portcounter] Check if counter ID exists before arithmetic operation 

jimmyzhai:
In local function `get_gbcounter` of portcounter.lua, need to check the return value of redis HGET. We can do addition operation only if counter ID exists and returns its value number.

### DIFF
--- a/common/portcounter.lua
+++ b/common/portcounter.lua
@@ -22,10 +22,14 @@ local gb_counter_list= {
 local function get_gbcounter(counter_id)
     local r = 0
     local counter = gb_counter_list[counter_id]
+    local num
     if counter then
         for i,id in ipairs(counter) do
             if #id > 0 then
-                r = r + redis.call('HGET', counters_table .. separator .. KEYS[i+1], id)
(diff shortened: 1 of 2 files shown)


common/portcounter.lua
```diff
@@ -22,10 +22,14 @@ local gb_counter_list= {
 local function get_gbcounter(counter_id)
     local r = 0
     local counter = gb_counter_list[counter_id]
+    local num
     if counter then
         for i,id in ipairs(counter) do
             if #id > 0 then
-                r = r + redis.call('HGET', counters_table .. separator .. KEYS[i+1], id)
+                num = redis.call('HGET', counters_table .. separator .. KEYS[i+1], id)
+                if num then
```
qiluo-msft, line 30: >if [](http://example.com/codeflow?start=16&length=2)

Add unit test for this case?

jimmyzhai: unit test is added

qiluo-msft (approved):

PR #635, opened 2022-06-20
Add thread safe guard to throw exception when share DBConnection cross threads

liuh-80:
#### Why I did it
For performance erason, swsscommon is not thread safe.
Code using swsscommon may make mistake to share db connection cross thread, and it's difficult to debug.
To slove this issue, add ThreadSafeGuard class to detect and throw exception when db connection shared by multiple thread.

#### How I did it
Add ThreadSafeCheck class to check and throw exception when DBConnect shared by multiple thread.

#### How to verify it
Pass all test case.
Add new test case to cover new code.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Add ThreadSafeCheck class to check and throw exception when DBConnect shared by multiple thread.

### DIFF
--- a/common/redisreply.cpp
+++ b/common/redisreply.cpp
@@ -33,6 +33,7 @@ inline void guard(FUNC func, const char* command)
 
 RedisReply::RedisReply(RedisContext *ctx, const RedisCommand& command)
 {
+    ThreadSafeGuard safeguard(ctx->getRunningFlag(), string(command.c_str()));
     int rc = redisAppendFormattedCommand(ctx->getContext(), command.c_str(), command.length());
     if (rc != REDIS_OK)
     {
@@ -51,6 +52,7 @@ RedisReply::RedisReply(RedisContext *ctx, const RedisCommand& command)
 
 RedisReply::RedisReply(RedisContext *ctx, const string& command)
 {
+    ThreadSafeGuard safeguard(ctx->getRunningFlag(), command);
     int rc = redisAppendCommand(ctx->getContext(), command.c_str());
     if (rc != REDIS_OK)
     {
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -69,7 +69,8 @@ libswsscommon_la_SOURCES = \
     warm_restart.cpp          \
     luatable.cpp              \
     countertable.cpp          \
-    redisutility.cpp
+    redisutility.cpp          \
+    threadsafeguard.cpp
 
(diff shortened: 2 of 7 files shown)


common/redisreply.cpp
```diff
@@ -33,6 +33,7 @@ inline void guard(FUNC func, const char* command)
 
 RedisReply::RedisReply(RedisContext *ctx, const RedisCommand& command)
 {
+    ThreadSafeGuard safeguard(ctx->getRunningFlag(), string(command.c_str()));
```
kcudnik, line 36: why not use std::mutex ? and mutex guard ? instead of implementing yourself

liuh-80: As I know the libswsscommon is not thread safe by design, so we add this class to catch the multi-thread scenario and throw exception.

If we add lock here, it will impact the critical performance, like route sync or warm reboot.

kcudnik: i dont think std::mutext will impact performace here, mutext is really quick, implementation internal is on atomic exchange, its 1 assembly instruction, so this is not performance critical here, and using already given tools make it simpler and better understanding for someone that is new to the code

if you are throwing here then how this will be handled in production scenario? just crash ?

liuh-80: As my understand mutex itself will not be the performance issue, here is a scenario after we add mutex:

Thread 1, run a lua script which take few minutes.
Thread 2, update a critical config.

And these 2 threads share same RedisContext.

So if thread 1 start first, because there is a mutex, thread 2 will wait untill thread 1 finish the script.

kcudnik: that depends on the mutex, you could have potentially spin lock since this operation could be very fast, and you dont need to use recursive_mutex as in showed PR, since you will not have recursive call here, still for me using std::mutex here is better than implementing it from scratch, also why crash on multiple thread access? and not wait ?

liuh-80: The ThreadSafeGuard is not a lock, it's just check race condition and throw exception when it happens.
Also, the code inside RedisContext ctor sometimes very slowly, for example run a LUA script.

Here is an example how crash happen:
caclmgrd currently share same RedisContext with multiple thread:
Thread 1: run 'keys' command
Thread 2: run 'hgetall' command

Then when thread 1 read command result, it read the 'keys' result, however the thread 1 process the result as 'hgetall' result, so the code crash.

ZhaohuiS:
@liuh-80 Can this PR fix the issue https://github.com/sonic-net/sonic-buildimage/issues/10883?
If multiple threads run different commands for same DBConnection at same time in caclmgrd, will this change avoid crash and just return exception?

PR #648, opened 2022-07-11
Add SonicV2Connector::set method for int value.

liuh-80:
#### Why I did it
Following config command in sonic-utility using an API does not exist in swss-common:

sudo config rate smoothing-interval 100

https://github.com/Azure/sonic-utilities/blob/e089b964ef0d38d8c691eb3e551e479c0c68dd96/config/main.py#L6548
counters_db.set('COUNTERS_DB', 'RATES:PORT', 'PORT_SMOOTH_INTERVAL', interval) <== interval is an int value.

Currently swss-common only support string value:
int64_t SonicV2Connector_Native::set(const std::string& db_name, const std::string& _hash, const std::string& key, const std::string& val, bool blocking)

#### How I did it
Add SonicV2Connector::set method for int value.

#### How to verify it
Pass all test case.
Add UT to cover new API.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Add SonicV2Connector::set method for int value.

### DIFF
--- a/common/sonicv2connector.cpp
+++ b/common/sonicv2connector.cpp
@@ -102,7 +102,7 @@ int64_t SonicV2Connector_Native::set(const std::string& db_name, const std::stri
 
 int64_t SonicV2Connector_Native::set(const std::string& db_name, const std::string& _hash, const std::string& key, const int val, bool blocking)
 {
-    return m_dbintf.set(db_name, _hash, key, to_string(val), blocking);
+    return m_dbintf.set(db_name, _hash, key, std::to_string(val), blocking);
 }
 
 int64_t SonicV2Connector_Native::del(const std::string& db_name, const std::string& key, bool blocking)
--- a/common/sonicv2connector.cpp
+++ b/common/sonicv2connector.cpp
@@ -99,11 +99,6 @@ int64_t SonicV2Connector_Native::set(const std::string& db_name, const std::stri
     return m_dbintf.set(db_name, _hash, key, val, blocking);
 }
 
-int64_t SonicV2Connector_Native::set(const std::string& db_name, const std::string& _hash, const std::string& key, const int val, bool blocking)
-{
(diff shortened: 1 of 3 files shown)


liuh-80:
Another PR depends on this change: https://github.com/Azure/sonic-utilities/pull/2251

liuh-80:
This PR need improve, because following error:

NotImplementedError: Wrong number or type of arguments for overloaded function 'SonicV2Connector_Native_set'.
  Possible C/C++ prototypes are:
    swss::SonicV2Connector_Native::set(std::string const &,std::string const &,std::string const &,std::string const &,bool)
    swss::SonicV2Connector_Native::set(std::string const &,std::string const &,std::string const &,int const,bool)

liuh-80:
> This PR need improve, because following error:
> 
> NotImplementedError: Wrong number or type of arguments for overloaded function 'SonicV2Connector_Native_set'. Possible C/C++ prototypes are: swss::SonicV2Connector_Native::set(std::string const &,std::string const &,std::string const &,std::string const &,bool) swss::SonicV2Connector_Native::set(std::string const &,std::string const &,std::string const &,int const,bool)



> This PR need improve, because following error:
> 
> NotImplementedError: Wrong number or type of arguments for overloaded function 'SonicV2Connector_Native_set'. Possible C/C++ prototypes are: swss::SonicV2Connector_Native::set(std::string const &,std::string const &,std::string const &,std::string const &,bool) swss::SonicV2Connector_Native::set(std::string const &,std::string const &,std::string const &,int const,bool)

Fixed by using python method.

qiluo-msft (approved):

liuh-80:
Build sai redis failed because UT failed, all build failed after sunday, so not related with this PR. will keep retry and merge after it passed.

PR #649, opened 2022-07-13
APIs to support streaming structured events

renukamanavalan:
Refer: [HLD #954](https://github.com/sonic-net/SONiC/pull/954)

1) APIs to publish & receive
2) Service APIs to cache events offline
3) C-binding for APIs to facilitate use from golang.


tests/Makefile.am
```diff
@@ -37,9 +37,11 @@ tests_SOURCES = redis_ut.cpp                \
...
                 timer_ut.cpp                \
-                cli_ut.cpp                  \
+                events_common_ut.cpp        \
+                events_service_ut.cpp       \
+                events_ut.cpp               \
                 main.cpp
 
 tests_CFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_GTEST) $(LIBNL_CFLAGS)
 tests_CPPFLAGS = $(DBGFLAGS) $(AM_CFLAGS) $(CFLAGS_COMMON) $(CFLAGS_GTEST) $(LIBNL_CFLAGS)
-tests_LDADD = $(LDADD_GTEST) -lpthread -L$(top_srcdir)/common -lswsscommon $(LIBNL_LIBS) $(CODE_COVERAGE_LIBS) -L$(top_srcdir)/sonic-db-cli -lsonicdbcli
```
qiluo-msft, line 45: >sonicdbcli [](http://example.com/codeflow?start=142&length=10)

Why remove `sonicdbcli`?

common/logger.cpp
```diff
@@ -186,6 +186,8 @@ Logger& Logger::getInstance()
 void Logger::setMinPrio(Priority prio)
 {
     getInstance().m_minPrio = prio;
+    SWSS_LOG_ERROR("loglevel Set=%d", prio);
```
qiluo-msft, line 189: >SWSS_LOG_ERROR [](http://example.com/codeflow?start=4&length=14)

This is not an error.

renukamanavalan: removed.

renukamanavalan:
@qiluo-msft, @zbud-msft, @lguohan, @liuh-80, @kcudnik,
Can you please provide your reviews, ASAP ?
We are currently blocked. This will help us progress.

tests/Makefile.am
renukamanavalan: mistake. Adding back

common/events.cpp
```diff
@@ -0,0 +1,611 @@
...
+#include "events_wrap.h"
+
+/*
+ *  Track created publishers to avoid duplicates
+ *  As receivers track missed event count by publishing instances, avoiding
+ *  duplicates helps reduce load.
+ */
+
+typedef map <string, EventPublisher *> lst_publishers_t;
+lst_publishers_t s_publishers;
```
kcudnik, line 11: why this is glbal variable? use static member in class if you need that

common/events.cpp
```diff
@@ -0,0 +1,611 @@
...
+     * NOTE: We don't wait for response here, but read it upon first publish
+     * If the publisher init happened early at the start by caller, by the
+     * the first event is published, the echo response will be available locally.
+     */
+    rc = m_event_service.echo_send("hello");
+    RET_ON_ERR (rc == 0, "Failed to echo send in event service");
+
+    m_event_source = event_source;
+
+    {
```
kcudnik, line 46: why this needs separate scope?

common/events.cpp
```diff
@@ -0,0 +1,611 @@
...
+events_init_publisher(const string event_source)
+{
+    event_handle_t ret = NULL;
+    lst_publishers_t::const_iterator itc = s_publishers.find(event_source);
+    if (itc != s_publishers.end()) {
+        // Pre-exists
+        ret = itc->second;
+    }
+    else {
+        EventPublisher *p =  new EventPublisher();
```
kcudnik, line 154: use shared_ptr to hold all pointers

common/events.cpp
```diff
@@ -0,0 +1,611 @@
...
+static EventSubscriber *s_subscriber = NULL;
+
+event_handle_t
+events_init_subscriber(bool use_cache, int recv_timeout,
+        const event_subscribe_sources_t *sources)
+{
+    EventSubscriber *sub = NULL;
+
+    if (s_subscriber == NULL) {
+        sub = new EventSubscriber();
```
qiluo-msft, line 419: >new [](http://example.com/codeflow?start=14&length=3)

Recommend use smart pointer instead of manage memory by new/delete.

common/events.cpp
```diff
@@ -0,0 +1,611 @@
...
+
+out:
+    return rc;
+}
+
+event_handle_t
+events_init_publisher(const string event_source)
+{
+    event_handle_t ret = NULL;
+    lst_publishers_t::const_iterator itc = s_publishers.find(event_source);
```
qiluo-msft, line 148: >lst_publishers_t::const_iterator [](http://example.com/codeflow?start=4&length=32)

Recommend use `auto&` to simplify.

common/events.h
```diff
@@ -0,0 +1,206 @@
...
+ *      The YANG module name for the event source. All events published with the handle
+ *      returned by this call is tagged with this source, transparently. The receiver
+ *      could subscribe with this source as filter.
+ *
+ * Return 
+ *  Non NULL handle
+ *  NULL on failure
+ */
+
+event_handle_t events_init_publisher(const std::string event_source);
```
kcudnik, line 44: most if not all stuff in swss-common is OOP code, and here you are adding some global functions, maybe there is way to make this oop ?

common/events.cpp
renukamanavalan: Good catch. Fixed.

common/events.cpp
renukamanavalan: Functionality is a standalone. Now made it indeed standalone.

common/events.cpp
```diff
@@ -0,0 +1,611 @@
...
+    /*
+     * REQ socket is connected and a message is sent & received, more to
+     * ensure PUB socket had enough time to establish connection.
+     * Any message published before connection establishment is dropped.
+     * NOTE: We don't wait for response here, but read it upon first publish
+     * If the publisher init happened early at the start by caller, by the
+     * the first event is published, the echo response will be available locally.
+     */
+    rc = m_event_service.echo_send("hello");
+    RET_ON_ERR (rc == 0, "Failed to echo send in event service");
```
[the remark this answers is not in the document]
renukamanavalan: yes

common/events.cpp
renukamanavalan: good point

common/events.cpp
```diff
@@ -0,0 +1,611 @@
...
+    for(itc=s_publishers.begin(); itc != s_publishers.end(); ++itc) {
+        if (itc->second == handle) {
+            return itc->second->publish(tag, params);
+        }
+    }
+    return -1;
+}
+
+
+EventSubscriber::EventSubscriber() : m_zmq_ctx(NULL), m_socket(NULL),
```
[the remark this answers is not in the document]
renukamanavalan: Fixed

common/events.cpp
renukamanavalan: Done

common/events.cpp
renukamanavalan: const_iterator helps compiler spin out more efficient code. I am not sure, what auto would spin out.
Plus I prefer not using auto as it completely hides the data type.
Only in cases of complex cases, I go for auto.

common/events.cpp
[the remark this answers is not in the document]
kcudnik: this seems like a bad design here :(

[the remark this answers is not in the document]
renukamanavalan: What makes it bad ?
These are all in error cases. 
They all goto a specific label, which gives opportunity for any cleanup.
I don't like goto if it is used in a logic, like if condition go here, but here it is not. 
Here for any failure goto out (get out);

This makes the code pretty readable too, which otherwise will be inundated with if clauses before every statement.

Just found this from s[tackoverflow](https://stackoverflow.com/questions/788903/valid-use-of-goto-for-error-management-in-c)
 _While goto is a bad idea in general, it can be useful for error handling when done in a simple and uniform manner. In this situation, even though it's a goto, it's being used in well-defined and more or less structured manner_.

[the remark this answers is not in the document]
kcudnik: its in general bad idea to use goto, and imo even worse to hide it in macro like this, idk about this approach what would be better here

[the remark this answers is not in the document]
renukamanavalan: idk ???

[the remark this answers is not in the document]
renukamanavalan: MACRO is explicit "RET_ON_ERR"
If you prefer, I can rename it, "GOTO_OUT_ON_ERR"

common/events.cpp
```diff
@@ -19,7 +19,7 @@ EventPublisher::get_publisher(const string event_source)
         ret = itc->second;
     }
     else {
-        EventPublisher_ptr_t p(n.get()ew EventPublisher());
+        EventPublisher_ptr_t p(new EventPublisher());
```
kcudnik, line 22: why not use shared_ptr?

renukamanavalan: Eventpublisher_ptr_t is shared_ptr.
I could use new or make_shared to create the pointer. make_shared also uses new under the  hood.
Moreover make_shared demands constructor to be public, but I have it under private to avoid any new from outside. Hence using new instead of the make_shared macro

kcudnik:
please fix build issues

PR #682, opened 2022-09-09
Load Yang model default value to DefaultValueProvider

liuh-80:
#### Why I did it
Support sonic-swss-common read Yang model defaut value.

#### How I did it
Load Yang model default value to DefaultValueProvider.

#### How to verify it
Add new UT to cover DefaultValueProvider code.
Pass all UT.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Load Yang model default value to DefaultValueProvider.


.azure-pipelines/build-sairedis-template.yml
```diff
@@ -20,6 +20,10 @@ parameters:
 - name: sonic_slave
   type: string
```
liuh-80, line 22: All azure pipeline change is from another PR: https://github.com/sonic-net/sonic-swss-common/pull/681
That PR need merge first.

qiluo-msft:
> #include <fstream>

You removed the whole content and changed file attribute. Is it intended?

---
Refers to: common/dbconnector.cpp:7 in 00935a3. [](commit_id = 00935a38360083bcc5ea7bbbb073e947985e60e9, deletion_comment = True)

common/defaultvalueprovider.cpp
```diff
@@ -0,0 +1,544 @@
...
+#include "logger.h"
+#include "table.h"
+#include "json.h"
+
+#if defined(__arm__) || defined(__aarch64__)
+#define WARNINGS_NO_CAST_ALIGN \
+    _Pragma ("GCC diagnostic push") \
+    _Pragma ("GCC diagnostic ignored \"-Wcast-align\"")
+#define WARNINGS_RESET \
+   _Pragma ("GCC diagnostic pop")
```
qiluo-msft, line 23: >    [](http://example.com/codeflow?start=0&length=3)

Indent 4 spaces?

common/defaultvalueprovider.cpp
```diff
@@ -0,0 +1,544 @@
...
+    SWSS_LOG_ERROR("DefaultValueProvider: %s", message.c_str());
+    throw runtime_error(message);
+}
+
+TableInfoBase::TableInfoBase()
+{
+    // C++ need this empty ctor
+}
+
+shared_ptr<string> TableInfoBase::GetDefaultValue(const string &key, const string &field)
```
qiluo-msft, line 43: >shared_ptr<string> [](http://example.com/codeflow?start=0&length=18)

Just return a `string`.

common/defaultvalueprovider.h
```diff
@@ -0,0 +1,118 @@
...
+{
+public:
+    TableInfoBase();
+
+    void AppendDefaultValues(const std::string &row, std::map<std::string, std::string>& sourceValues, std::map<std::string, std::string>& targetValues);
+
+    std::shared_ptr<std::string> GetDefaultValue(const std::string &row, const std::string &field);
+
+protected:
+    virtual bool FindFieldMappingByKey(const std::string &row, FieldDefaultValueMapping ** foundedMappingPtr) = 0;
```
qiluo-msft, line 35: >row [](http://example.com/codeflow?start=58&length=3)

do you mean `key`?

common/defaultvalueprovider.h
```diff
@@ -0,0 +1,118 @@
...
+
+    bool FindFieldMappingByKey(const std::string &row, FieldDefaultValueMapping ** foundedMappingPtr);
+};
+
+class DefaultValueProvider
+{
+public:
+    static DefaultValueProvider& instance();
+
+    void appendDefaultValues(const std::string &table, const std::string &key, std::vector<std::pair<std::string, std::string>> &values);
```
qiluo-msft, line 79: >values [](http://example.com/codeflow?start=129&length=6)

Do you mean `fvs`? ie. field and value pairs

common/defaultvalueprovider.h
```diff
@@ -0,0 +1,118 @@
...
+
+    std::map<std::string, std::string> getDefaultValues(const std::string &table, const std::string &key);
+
+#ifndef UNITTEST
+private:
+#endif
+    DefaultValueProvider();
+    ~DefaultValueProvider();
+
+    //  libyang context
```
qiluo-msft, line 91: >   [](http://example.com/codeflow?start=6&length=2)

One blank?

common/defaultvalueprovider.h
```diff
@@ -0,0 +1,118 @@
...
+    struct ly_ctx *m_context = nullptr;
+
+    // The table name to table default value info mapping
+    std::map<std::string, std::shared_ptr<TableInfoBase> > m_defaultValueMapping;
+
+    void Initialize(char* modulePath = DEFAULT_YANG_MODULE_PATH);
+    
+    void LoadModule(const std::string name, const std::string path, struct ly_ctx *context);
```
qiluo-msft, line 99: >const std::string [](http://example.com/codeflow?start=20&length=17)

Normally use `string` or `const string&`.

tests/defaultvalueprovider_ut.cpp
```diff
@@ -0,0 +1,98 @@
...
+    EXPECT_EQ(values["interval"], "30");
+
+    // default value define in leaflist
+    EXPECT_EQ(values["domain-search"], "[\"test\"]");
+}
+
+TEST(DECORATOR, GetDefaultValues)
+{
+    DefaultValueProvider instance;
+    instance.Initialize("./tests/yang");
```
qiluo-msft, line 29: >./tests/yang [](http://example.com/codeflow?start=25&length=12)

The relative path is working only when the cwd is this "tests" folder.

common/defaultvalueprovider.cpp
```diff
@@ -0,0 +1,544 @@
...
+    {
+        // KeyInfo.first is key value
+        string keyValue = fieldMapping.first.first;
+        m_defaultValueMapping.emplace(keyValue, fieldMapping.second);
+
+        SWSS_LOG_DEBUG("TableInfoDict::TableInfoDict %s\n", keyValue.c_str());
+    }
+}
+
+bool TableInfoDict::FindFieldMappingByKey(const string &key, FieldDefaultValueMapping ** foundedMappingPtr)
```
qiluo-msft, line 104: >bool [](http://example.com/codeflow?start=0&length=4)

If return a bool, recommend the function name `FoundFieldMappingByKey`.

`FindXxx` normally return an iterator.

common/defaultvalueprovider.cpp
```diff
@@ -0,0 +1,544 @@
...
+    }
+
+    SWSS_LOG_DEBUG("Found default value for field %s=%s\n", field.c_str(), fieldData->second.c_str());
+    return make_shared<string>(fieldData->second);
+}
+
+// existedValues and targetValues can be same container.
+void TableInfoBase::AppendDefaultValues(const string &row, map<string, string>& existedValues, map<string, string>& targetValues)
+{
+    assert(!row.empty());
```
qiluo-msft, line 69: >assert [](http://example.com/codeflow?start=4&length=6)

This is a public function, you should not use `assert` to validate input because it is noop in release build.

common/defaultvalueprovider.cpp
qiluo-msft, line 104: >foundedMappingPtr [](http://example.com/codeflow?start=89&length=17)

typo: foundMappingPtr

common/defaultvalueprovider.cpp
```diff
@@ -0,0 +1,544 @@
...
+}
+
+bool TableInfoDict::FindFieldMappingByKey(const string &key, FieldDefaultValueMapping ** foundedMappingPtr)
+{
+    assert(!key.empty());
+    assert(foundedMappingPtr != nullptr);
+
+    SWSS_LOG_DEBUG("TableInfoDict::FindFieldMappingByKey %s\n", key.c_str());
+    auto keyResult = m_defaultValueMapping.find(key);
+    *foundedMappingPtr = keyResult->second.get();
```
qiluo-msft, line 111: >keyResult [](http://example.com/codeflow?start=25&length=9)

If not found, will it throw?

common/defaultvalueprovider.cpp
```diff
@@ -0,0 +1,544 @@
...
+bool TableInfoMultipleList::FindFieldMappingByKey(const string &key, FieldDefaultValueMapping ** foundedMappingPtr)
+{
+    assert(!key.empty());
+    assert(foundedMappingPtr != nullptr);
+
+    SWSS_LOG_DEBUG("TableInfoMultipleList::FindFieldMappingByKey %s\n", key.c_str());
+    int fieldCount = (int)count(key.begin(), key.end(), '|') + 1;
+    auto keyInfo = m_defaultValueMapping.find(fieldCount);
+    
+    // when not found, key_info still a valied iterator
```
qiluo-msft, line 149: >valied [](http://example.com/codeflow?start=40&length=6)

typo

common/defaultvalueprovider.cpp
```diff
@@ -0,0 +1,544 @@
...
+    *foundedMappingPtr = keyInfo->second.get();
+    
+    // return false when not found
+    return keyInfo != m_defaultValueMapping.end();
+}
+
+DefaultValueProvider& DefaultValueProvider::instance()
+{
+    static DefaultValueProvider instance;
+#ifndef UNITTEST
```
qiluo-msft, line 159: >UNITTEST [](http://example.com/codeflow?start=8&length=8)

Why you need this? Possible to remove unit test related conditional code?

common/defaultvalueprovider.cpp
```diff
@@ -0,0 +1,544 @@
...
+        {
+            SWSS_LOG_DEBUG("choice field: %s\n",field->name);
+            GetDefaultValueInfoForChoice(field, fieldMapping);
+        }
+        else if (field->nodetype == LYS_LEAFLIST)
+        {
+            SWSS_LOG_DEBUG("list field: %s\n",field->name);
+            GetDefaultValueInfoForLeaflist(field, fieldMapping);
+        }
+#ifdef DEBUG
```
qiluo-msft, line 468: >DEBUG [](http://example.com/codeflow?start=7&length=5)

Why need this check?

common/defaultvalueprovider.h
```diff
@@ -0,0 +1,118 @@
...
+#include <vector>
+#include "common/table.h"
+
+#define DEFAULT_YANG_MODULE_PATH "/usr/local/yang-models"
+#define EMPTY_STR ""
+
+struct ly_ctx;
+
+// Key information
+typedef std::pair<std::string, int> KeyInfo;
```
qiluo-msft, line 14: >KeyInfo [](http://example.com/codeflow?start=36&length=7)

I checked HLD and there is no document. Suggest add enough code comments to explain the struct. It is difficult to understand:
1. what is key's infor?
2. what is the first string?
3. what is the second int?

common/defaultvalueprovider.h
```diff
@@ -0,0 +1,118 @@
...
+
+// Field name to default value mapping
+typedef std::map<std::string, std::string> FieldDefaultValueMapping;
```
qiluo-msft, line 17: >string [](http://example.com/codeflow?start=35&length=6)

Is it true that the default value is not a string?
For example: [https://github.com/sonic-net/sonic-buildimage/blob/19155df148fbb49eecd2ce4a3b8f022a66beab76/src/sonic-yang-models/yang-models/sonic-buffer-pg.yang#L54](https://github.com/sonic-net/sonic-buildimage/blob/19155df148fbb49eecd2ce4a3b8f022a66beab76/src/sonic-yang-models/yang-models/sonic-buffer-pg.yang#L54)

common/defaultvalueprovider.h
```diff
@@ -0,0 +1,118 @@
...
+    std::map<std::string, std::string> getDefaultValues(const std::string &table, const std::string &key);
+
+#ifndef UNITTEST
+private:
+#endif
+    DefaultValueProvider();
+    ~DefaultValueProvider();
+
+    //  libyang context
+    struct ly_ctx *m_context = nullptr;
```
qiluo-msft, line 92: >m_context [](http://example.com/codeflow?start=19&length=9)

You are assuming there is only one `ly_ctx`. Is it always true? Can we relax?

common/defaultvalueprovider.cpp
```diff
@@ -0,0 +1,544 @@
...
+    {
+        const char** dfltValues = listNode->dflt;
+        //convert list default value to json string
+        string dfltValueJson = JSon::buildJson(dfltValues);
+        SWSS_LOG_DEBUG("list field: %s, default: %s\n",field->name, dfltValueJson.c_str());
+        fieldMapping->emplace(string(field->name), dfltValueJson);
+    }
+}
+
+FieldDefaultValueMappingPtr DefaultValueProvider::GetDefaultValueInfo(struct lys_node* tableChildNode)
```
qiluo-msft, line 444: >GetDefaultValueInfo [](http://example.com/codeflow?start=50&length=19)

This function has nothing todo with the concept of "default value". Seems a general help function related to yang data structure. Suggest move out of this class.

common/defaultvalueprovider.cpp
```diff
@@ -0,0 +1,544 @@
...
+    SWSS_LOG_DEBUG("DefaultValueProvider::GetDefaultValueInfo %s\n",tableChildNode->name);
+
+    auto field = tableChildNode->child;
+    auto fieldMapping = make_shared<FieldDefaultValueMapping>();
+    while (field)
+    {
+        if (field->nodetype == LYS_LEAF)
+        {
+            SWSS_LOG_DEBUG("leaf field: %s\n",field->name);
+            GetDefaultValueInfoForLeaf(field, fieldMapping);
```
qiluo-msft, line 456: >field [](http://example.com/codeflow?start=39&length=5)

You just checked the `nodetype` field, so better cast `field` to proper class in this function, and no need to cast in the lower level function such as `GetDefaultValueInfoForLeaf`.

common/defaultvalueprovider.cpp
```diff
@@ -0,0 +1,544 @@
...
+#include <stdio.h> 
+
+#include <libyang/libyang.h>
+
+#include "defaultvalueprovider.h"
+#include "logger.h"
+#include "table.h"
+#include "json.h"
+
+#if defined(__arm__) || defined(__aarch64__)
```
qiluo-msft, line 18: >__aarch64__ [](http://example.com/codeflow?start=32&length=11)

SONiC will be built on arm platforms. Is it safe to ignore the warning (as error)?

liuh-80:
> > #include
> 
> You removed the whole content and changed file attribute. Is it intended?
> 
> Refers to: common/dbconnector.cpp:7 in [00935a3](https://github.com/sonic-net/sonic-swss-common/commit/00935a38360083bcc5ea7bbbb073e947985e60e9). [](commit_id = [00935a3](https://github.com/sonic-net/sonic-swss-common/commit/00935a38360083bcc5ea7bbbb073e947985e60e9), deletion_comment = True)

Fixed, this PR not change any content in dbconnector.cpp, I only fix the incorrect file mode issue. seems a github bug.

common/defaultvalueprovider.cpp
liuh-80: Can't return string because following 2 scenario are different:
1. When Yang model set default value to a empty string, we need return a empty string.
2. When Yang model does not set default value, we need return a nullptr.

common/defaultvalueprovider.cpp
liuh-80: When not found, here will not throw exception.
When not found, the keyResult->second.get() will return a nullptr and the return value of this method will be False. the caller code will check return value to handle this case.

common/defaultvalueprovider.h
liuh-80: Yes, it's true that default value may not a string.

However, libyang always get default value as string. so in internal we store the default value as string.

Also currently swss-common will return field-value pair as a string to string dictionary.

common/defaultvalueprovider.h
liuh-80: It's true, there will only one ly_ctx, it's created by our code and this is a singleton.
Libyang only need 1 context, there is no scenario that we need multiple context.

common/defaultvalueprovider.cpp
liuh-80: New UT cover the warning code and passed, so I think it's safe.
Also I check this issue online and found there some discussion suggest ignore this warning is safe.

common/defaultvalueprovider.cpp
liuh-80: Fixed.

common/defaultvalueprovider.h
liuh-80: Fixed, also fix other similer cases.

common/defaultvalueprovider.h
liuh-80: Fixed.

common/defaultvalueprovider.h
liuh-80: Fixed.

common/defaultvalueprovider.h
liuh-80: Fixed, change to 'const string&'

tests/defaultvalueprovider_ut.cpp
liuh-80: There is no better solution, other C++ UT also using relative path for config file.

common/defaultvalueprovider.cpp
liuh-80: Fixed, change to FoundFieldMappingByKey

common/defaultvalueprovider.cpp
liuh-80: Fixed, assert removed.

common/defaultvalueprovider.cpp
liuh-80: Fixed.

common/defaultvalueprovider.cpp
liuh-80: Fixed.

common/defaultvalueprovider.cpp
liuh-80: Will check if gmock can fix this issue and update later.

common/defaultvalueprovider.cpp
liuh-80: Fixed.

common/defaultvalueprovider.h
liuh-80: Fixed, add more detail to comments.

common/defaultvalueprovider.cpp
liuh-80: Fixed, create a new DefaultValueHelper class and move these method to DefaultValueHelper and change to static method.

common/defaultvalueprovider.cpp
liuh-80: Fixed

common/defaultvalueprovider.cpp
liuh-80: Fixed.

common/defaultvalueprovider.cpp
```diff
@@ -0,0 +1,493 @@
...
+#define WARNINGS_NO_CAST_ALIGN \
+     _Pragma ("GCC diagnostic push") \
+     _Pragma ("GCC diagnostic ignored \"-Wcast-align\"")
```
qiluo-msft, line 18: >      [](http://example.com/codeflow?start=0&length=5)

Indent 4 spaces? 5 is weird.

qiluo-msft:
> #include <fstream>

I see non-critical comment change and still file attribute change. If it is intended change, suggest move to a standalone PR. This PR is complex and should be focused.

---
In reply to: [1242820539](https://github.com/sonic-net/sonic-swss-common/pull/682#issuecomment-1242820539) [](http://example.com/codeflow?ancestors=1242820539,1242820539)

---
Refers to: common/dbconnector.cpp:7 in 00935a3. [](commit_id = 00935a38360083bcc5ea7bbbb073e947985e60e9, deletion_comment = True)

common/defaultvalueprovider.h
qiluo-msft: Today the answer is yes. I am think relax the limitation may be future proof.

In future, ApplDB may have separate yang models, structure events may also has separate yang models. And these yang models are independent with each other. So I guess there will be use case that clients may use multiple `ly_ctx.`.

The lazy initialization is still useful since the client of swsscommon may not use yang features at all.

liuh-80: Fixed, now defaultvalue provider is not a singleton, so we can use multiple default value provider to load default value for different DB.

common/defaultvalueprovider.cpp
liuh-80: Fixed

liuh-80:
> > #include
> 
> I see non-critical comment change and still file attribute change. If it is intended change, suggest move to a standalone PR. This PR is complex and should be focused.
> 
> In reply to: [1242820539](https://github.com/sonic-net/sonic-swss-common/pull/682#issuecomment-1242820539)
> 
> Refers to: common/dbconnector.cpp:7 in [00935a3](https://github.com/sonic-net/sonic-swss-common/commit/00935a38360083bcc5ea7bbbb073e947985e60e9). [](commit_id = [00935a3](https://github.com/sonic-net/sonic-swss-common/commit/00935a38360083bcc5ea7bbbb073e947985e60e9), deletion_comment = True)

Fixed, reverted change in this file.

qiluo-msft (approved):

liuh-80:
Currently, there still 10 repos need update their azure pipeline to install libyang. 10 PRs already created.
This PR need wait for those PRs finish first.

PR #687, opened 2022-09-14
Make the loglevel persistent by moving the LOGGER table from the LOGLEVEL DB to the CONFIG DB

EdenGri:
This PR needs to be merged only after the following PR: https://github.com/sonic-net/sonic-buildimage/pull/11996

#### What I did
Add the ability to the user to save the loglevel and make it persistent to reboot.

#### How I did it
Move the logger tables from the LOGLEVEL DB to the CONFIG DB

#### How to verify it
1. Change the orchagent loglevel (for example) -> swssloglevel -c orchagent -l DEBUG
2. Save the loglevel -> run config save
3. Reboot
4. Verify that the orchagent log level is still DEBUG ->run run redis-cli -n 4 hgetall "LOGGER|orchagent"


common/logger.cpp
```diff
@@ -120,9 +121,12 @@ void Logger::linkToDbWithOutput(
     // Initialize internal DB with observer
     logger.m_settingChangeObservers.insert(std::make_pair(dbName, std::make_pair(prioNotify, outputNotify)));
 
-    DBConnector db("LOGLEVEL_DB", 0);
+    DBConnector db("CONFIG_DB", 0);
+    std::string key_prefix(CFG_LOGGER_TABLE_NAME);
+    key_prefix+="|";
```
qiluo-msft, line 126: >key_prefix [](http://example.com/codeflow?start=4&length=10)

You can use `Table` instead of `DBConnector` to access configDB table. Then you do not need to concat a key.

common/logger.h
```diff
@@ -46,6 +46,10 @@ void err_exit(const char *fn, int ln, int e, const char *fmt, ...)
 class Logger
 {
 public:
+
+#define DAEMON_LOGLEVEL "LOGLEVEL"
```
qiluo-msft, line 50: >DAEMON_LOGLEVEL [](http://example.com/codeflow?start=8&length=15)

you can use `static const char * const` or `static const std::string`

common/loglevel.cpp
```diff
@@ -177,11 +194,7 @@ int main(int argc, char **argv)
             keys.erase(std::remove_if(keys.begin(), keys.end(), filterOutSaiKeys), keys.end());
         }
 
-        for (const auto& key : keys)
-        {
-            setLoglevel(db, key, loglevel);
-        }
-
+        setAllLoglevel(logger_tbl,keys, loglevel);
```
qiluo-msft, line 197: >, [](http://example.com/codeflow?start=33&length=1)

Add a blank after `,`

tests/Makefile.am
```diff
@@ -28,6 +28,8 @@ tests_SOURCES = redis_ut.cpp                \
                 warm_restart_ut.cpp         \
                 redis_multi_db_ut.cpp       \
                 logger_ut.cpp               \
+                ../common/loglevel.cpp      \
```
qiluo-msft, line 31: >loglevel [](http://example.com/codeflow?start=26&length=8)

Is it needed? I think it should be linked.

common/loglevel.cpp
```diff
@@ -72,14 +68,35 @@ bool filterSaiKeys(const std::string& key)
...
+    return keys;
+}
+
+std::vector<std::string> get_no_sai_keys(std::vector<std::string> keys)
+{
+    keys.erase(std::remove_if(keys.begin(), keys.end(), filterOutSaiKeys), keys.end());
+    return keys;
+}
+
+void setAllLoglevel(swss::Table& logger_tbl, std::vector<std::string> components,std::string loglevel)
```
qiluo-msft, line 83: >, [](http://example.com/codeflow?start=80&length=1)

Add blank after `,`

tests/Makefile.am
EdenGri: It's not linked. 

Without this line I am getting the following error:

`
/usr/bin/ld: tests-loglevel_ut.o: in function 'LOGLEVEL_loglevel_Test::TestBody()':
./tests/loglevel_ut.cpp:31: undefined reference to 'swssloglevel(int, char**)'
collect2: error: ld returned 1 exit status
`

common/logger.cpp
EdenGri: Done, thanks

common/logger.h
EdenGri: Done, thanks

common/loglevel.cpp
EdenGri: Done, thanks

qiluo-msft (approved):

liat-grozovik:
merging as the PR as prev approved and only build failures fixed

liat-grozovik (approved):

PR #689, opened 2022-09-15
Check whether a pointer created by dynamic_cast is null before using it.

pettershao-ragilenetworks:
What I did
Check whether a pointer created by dynamic_cast is null before using it.
Why I did it
dynamic_cast<ConsumerStateTable *>(selectable) may return NULL

### DIFF
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -236,7 +236,10 @@ void Logger::settingThread()
         ConsumerStateTable *consumerStateTable = NULL;
         consumerStateTable = dynamic_cast<ConsumerStateTable *>(selectable);
         if (consumerStateTable == NULL)
-            continue;
+        {
+            SWSS_LOG_ERROR("dynamic_cast returns NULL.");
+            break;
+        }
         consumerStateTable->pop(koValues);
         std::string key = kfvKey(koValues), op = kfvOp(koValues);
 
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -1,5 +1,5 @@
 #include "logger.h"
-#include <assert.h>
+
 #include <algorithm>
 #include <stdarg.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


pettershao-ragilenetworks:
Hi~ @qiluo-msft  - could you please review and merge, thanks.

pettershao-ragilenetworks:
Hi~ could you please review, thanks. @prsunny @qiluo-msft

pettershao-ragilenetworks:
please review, thanks. @prsunny  @qiluo-msft

common/logger.cpp
```diff
@@ -233,7 +233,13 @@ void Logger::settingThread()
         }
 
         KeyOpFieldsValuesTuple koValues;
-        dynamic_cast<ConsumerStateTable *>(selectable)->pop(koValues);
```
qiluo-msft, line 236: >selectable [](http://example.com/codeflow?start=43&length=10)

I think it is always of the type `ConsumerStateTable`? Did you have a use case to trigger a negative case? If yes, could you add a unit test case?

If I understand it correctly, then you can add `assert`.

pettershao-ragilenetworks: The expression dynamic_cast(v) converts the expression v to type T. Type T must be a pointer or reference to a complete class type or a pointer to void. If T is a pointer and the dynamic_cast operator fails, the operator returns a null pointer of type T. If T is a reference and the dynamic_cast operator fails, the operator throws the exception std::bad_cast. You can find this class in the standard library header .

In C++ usage, it would be a problem.
It will not be triggered by the test case.

pettershao-ragilenetworks:
Please review, thanks. @qiluo-msft

common/logger.cpp
qiluo-msft: Totally understood what you are saying. My point is that it could not happen due to the function context. Then assert is the recommendation, no need to check in runtime.

qiluo-msft (dismissed):

common/logger.cpp
pettershao-ragilenetworks: All right, I have changed the error handling.

pettershao-ragilenetworks: I have changed. Please review. thanks.

qiluo-msft (approved):

pettershao-ragilenetworks:
Could you please merge, thanks. @qiluo-msft

qiluo-msft:
Pending on PR checkers passing.

pettershao-ragilenetworks:
I changed file to rerun the failed jobs, could you please review and merge. Thanks! @qiluo-msft

qiluo-msft (approved):

PR #691, opened 2022-09-19
[Fast/Warm restart] Implement helper class for waiting restart done

Junchao-Mellanox:
**Depends on:**

- https://github.com/sonic-net/sonic-sairedis/pull/1100
- https://github.com/sonic-net/sonic-utilities/pull/2286
- https://github.com/sonic-net/sonic-buildimage/pull/11594

**Why I did this?**

Daemons which are not related to warm/fast restart might affect the performance of warm/fast restart. A hardcoded start up delay is the current solution to avoid this.

This PR implements a function to wait warm/fast restart done. This function provided a efficiency and graceful way for daemons to wait warm/fast restart done.

**How I did it?**

Implement a utility function RestartWaiter::waitRestartDone. This function waits warm restart done flag in STATE DB and return true if the flag is set by warm restart finalizer. This function is also exposed as python extension so that python daemons can utilize it.

This PR depends on new fastboot design: https://github.com/sonic-net/SONiC/blob/master/doc/fast-reboot/Fast-reboot_Flow_Improvements_HLD.md

### DIFF
--- a/common/redis_table_waiter.cpp
+++ b/common/redis_table_waiter.cpp
@@ -1,0 +1,145 @@
+#include "redis_table_waiter.h"
+#include "select.h"
+#include "subscriberstatetable.h"
+
+using namespace swss;
+
+bool RedisTableWaiter::waitUntil(
+    DBConnector &db,
+    const std::string &tableName,
+    unsigned int maxWaitSec,
+    CheckFunc checkFunc)
+{
+    if (maxWaitSec == 0)
+    {
+        SWSS_LOG_ERROR("Error: invalid maxWaitSec value 0, must be larger than 0");
+        return false;
+    }
+
+    SubscriberStateTable table(&db, tableName);
+    Select s;
+    s.addSelectable(&table);
+
+    int maxWaitMs = static_cast<int>(maxWaitSec) * 1000;
+    int selectTimeout = maxWaitMs;
+    auto start = std::chrono::steady_clock::now();
+    while(1)
+    {
+        Selectable *sel = NULL;
+        int ret = s.select(&sel, selectTimeout, true);
+        if (ret == Select::OBJECT)
+        {
+            KeyOpFieldsValuesTuple kco;
+            table.pop(kco);
+            if (checkFunc(kco))
+            {
+                return true;
+            }
+        }
+        else if (ret == Select::ERROR)
+        {
+            SWSS_LOG_NOTICE("Error: wait redis table got error - %s!", strerror(errno));
+        }
+        else if (ret == Select::TIMEOUT)
+        {
+            SWSS_LOG_INFO("Timeout: wait redis table got select timeout");
+        }
+        else if (ret == Select::SIGNALINT)
+        {
+            return false;
+        }
+
+        auto end = std::chrono::steady_clock::now();
+        int delay = static_cast<int>(
+                    std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count());
+
+        if (delay >= maxWaitMs)
+        {
+            return false;
+        }
+
+        selectTimeout = maxWaitMs - delay;
+    }
+
+    return false;
+}
+
+bool RedisTableWaiter::waitUntilFieldSet(
+    DBConnector &db,
+    const std::string &tableName,
+    const std::string &key,
+    const std::string &fieldName,
+    unsigned int maxWaitSec,
+    ConditionFunc cond)
+{
+    auto sep = SonicDBConfig::getSeparator(&db);
+    auto value = db.hget(tableName + sep + key, fieldName);
+    if (value && cond(*value.get()))
+    {
+        return true;
+    }
+
+    auto checkFunc = [&](const KeyOpFieldsValuesTuple &kco) -> bool {
+        if (SET_COMMAND == kfvOp(kco))
+        {
+            if (key == kfvKey(kco))
+            {
+                auto& values = kfvFieldsValues(kco);
+                for (auto& fvt: values)
+                {
+                    if (fieldName == fvField(fvt))
+                    {
+                        return cond(fvValue(fvt));
+                    }
+                }
+            }
+        }
+
+        return false;
+    };
+    return waitUntil(db, tableName, maxWaitSec, checkFunc);
+}
+
+bool RedisTableWaiter::waitUntilKeySet(
+    DBConnector &db,
+    const std::string &tableName,
+    const std::string &key,
+    unsigned int maxWaitSec)
+{
+    auto sep = SonicDBConfig::getSeparator(&db);
+    if (db.exists(tableName + sep + key))
+    {
+        return true;
+    }
+
+    auto checkFunc = [&](const KeyOpFieldsValuesTuple &kco) -> bool {
+        if (SET_COMMAND == kfvOp(kco))
+        {
+            return key == kfvKey(kco);
+        }
+        return false;
+    };
+    return waitUntil(db, tableName, maxWaitSec, checkFunc);
+}
+
+bool RedisTableWaiter::waitUntilKeyDel(
+    DBConnector &db,
+    const std::string &tableName,
+    const std::string &key,
+    unsigned int maxWaitSec)
+{
+    auto sep = SonicDBConfig::getSeparator(&db);
+    if (!db.exists(tableName + sep + key))
+    {
+        return true;
+    }
+
+    auto checkFunc = [&](const KeyOpFieldsValuesTuple &kco) -> bool {
+        if (DEL_COMMAND == kfvOp(kco))
+        {
+            return key == kfvKey(kco);
+        }
+        return false;
+    };
+    return waitUntil(db, tableName, maxWaitSec, checkFunc);
+}
--- a/common/redis_table_waiter.cpp
+++ b/common/redis_table_waiter.cpp
@@ -8,7 +8,7 @@ bool RedisTableWaiter::waitUntil(
     DBConnector &db,
     const std::string &tableName,
     unsigned int maxWaitSec,
-    CheckFunc checkFunc)
+    CheckFunc &checkFunc)
 {
     if (maxWaitSec == 0)
     {
@@ -70,7 +70,7 @@ bool RedisTableWaiter::waitUntilFieldSet(
     const std::string &key,
     const std::string &fieldName,
     unsigned int maxWaitSec,
-    ConditionFunc cond)
+    ConditionFunc &cond)
 {
     auto sep = SonicDBConfig::getSeparator(&db);
     auto value = db.hget(tableName + sep + key, fieldName);
@@ -79,7 +79,7 @@ bool RedisTableWaiter::waitUntilFieldSet(
         return true;
     }
 
-    auto checkFunc = [&](const KeyOpFieldsValuesTuple &kco) -> bool {
+    CheckFunc checkFunc = [&](const KeyOpFieldsValuesTuple &kco) -> bool {
         if (SET_COMMAND == kfvOp(kco))
         {
             if (key == kfvKey(kco))
@@ -112,7 +112,7 @@ bool RedisTableWaiter::waitUntilKeySet(
         return true;
     }
 
-    auto checkFunc = [&](const KeyOpFieldsValuesTuple &kco) -> bool {
+    CheckFunc checkFunc = [&](const KeyOpFieldsValuesTuple &kco) -> bool {
         if (SET_COMMAND == kfvOp(kco))
         {
             return key == kfvKey(kco);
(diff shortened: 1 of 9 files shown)


stephenxs (dismissed):

common/restart_waiter.cpp
```diff
@@ -0,0 +1,130 @@
...
+        else if (ret == Select::SIGNALINT)
+        {
+            return false;
+        }
+
+        auto end = std::chrono::steady_clock::now();
+        int delay = static_cast<int>(
+                    std::chrono::duration_cast<std::chrono::seconds>(end - start).count());
+
+        selectTimeout -= delay;
```
qiluo-msft, line 112: >delay [](http://example.com/codeflow?start=25&length=5)

If `delay` is truncated to 0, infinite loop?

common/restart_waiter.cpp
```diff
@@ -0,0 +1,130 @@
...
+        return true;
+    }
+
+    return isWarmOrFastRestartInProgress(stateDb) ? doWait(stateDb, maxWaitSec) : true;
+}
+
+bool RestartWaiter::doWait(DBConnector &stateDb,
+                           unsigned int maxWaitSec)
+{
+    int selectTimeout = static_cast<int>(maxWaitSec);
```
qiluo-msft, line 57: >maxWaitSec [](http://example.com/codeflow?start=41&length=10)

Is 0 valid value? Will it burn CPU in a hot loop?

common/restart_waiter.cpp
Junchao-Mellanox: will fix

common/restart_waiter.cpp
Junchao-Mellanox: will fix

common/restart_waiter.cpp
qiluo-msft: Let me clarify my assumption. If `delay == 0` happen due to `end - start` is so small, `selectTimeout` will not change, and this is infinite loop.

Junchao-Mellanox: understood, please check again?

stephenxs (dismissed):

Junchao-Mellanox:
Hi @qiluo-msft @liuh-80, could you please review and sign-off?

common/restart_waiter.cpp
```diff
@@ -0,0 +1,136 @@
...
+        {
+            return false;
+        }
+    }
+}
+
+bool RestartWaiter::isWarmOrFastRestartInProgress(DBConnector &stateDb)
+{
+    auto ret = stateDb.hget(STATE_WARM_RESTART_ENABLE_TABLE_NAME + STATE_DB_SEPARATOR + RESTART_KEY, RESTART_ENABLE_FIELD);
+    return ret && *ret.get() == "true";
```
liuh-80, line 129: SAA, do we need handle "True" case?

common/restart_waiter.h
```diff
@@ -0,0 +1,32 @@
...
+    static bool waitFastRestartDone(unsigned int maxWaitSec = 180,
+                                    unsigned int dbTimeout = 0,
+                                    bool isTcpConn = false);
+
+private:
+    static bool doWait(swss::DBConnector &stateDb,
+                       unsigned int maxWaitSec);
+
+    static bool isWarmOrFastRestartInProgress(swss::DBConnector &stateDb);
+    static bool isFastRestartInProgress(swss::DBConnector &stateDb);
```
stephenxs, line 29: Can we expose it as a public method? I think we need a way to distinguish fast boot from warm boot.

stephenxs (dismissed):

Junchao-Mellanox:
Hi @liuh-80, I have fixed all comments, could you please review it？

stephenxs (approved):

Junchao-Mellanox:
Hi @qiluo-msft , could you please review and sign-off?

common/restart_waiter.cpp
qiluo-msft: Still applicable. It is possible that `delay==0`, and `selectTimeout>0` always.

common/restart_waiter.cpp
```diff
@@ -0,0 +1,157 @@
...
+        if (selectTimeout <= 0)
+        {
+            return false;
+        }
+    }
+}
+
+bool RestartWaiter::isWarmOrFastRestartInProgress(DBConnector &stateDb)
+{
+    auto ret = stateDb.hget(STATE_WARM_RESTART_ENABLE_TABLE_NAME + STATE_DB_SEPARATOR + RESTART_KEY, RESTART_ENABLE_FIELD);
```
qiluo-msft, line 139: >STATE_WARM_RESTART_ENABLE_TABLE_NAME [](http://example.com/codeflow?start=28&length=36)

Why not using `WarmStart` class?

common/restart_waiter.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+        selectTimeout -= delay;
+        if (selectTimeout <= 0)
+        {
+            return false;
+        }
+    }
+}
+
+bool RestartWaiter::isWarmOrFastRestartInProgress(DBConnector &stateDb)
```
qiluo-msft, line 137: >isWarmOrFastRestartInProgress [](http://example.com/codeflow?start=20&length=29)

The implementation is about warm reboot only?

common/restart_waiter.cpp
Junchao-Mellanox: To use warm restart class, it is required to execute` WarmStart::checkWarmStart` function before calling `WarmStart::isWarmStart`. `WarmStart::checkWarmStart` requires some input parameters:

```cpp
    static bool checkWarmStart(const std::string &app_name,
                               const std::string &docker_name,
                               const bool incr_restore_cnt = true);
```

For a daemon who only wants to know that whether warm restart is in progress, it is not necessary to force the daemon give `app_name` and `docker_name`. This is too much for a daemon.

liat-grozovik:
@qiluo-msft could you please help to review comments feedback?

common/restart_waiter.cpp
Junchao-Mellanox: How about change it to following:

```cpp
        if (delay == 0)
        {
            // avoid delay being truncated to 0
            delay = 1;
        }
```

qiluo-msft: This change is too random to me. Could you explore wait on a "monotonic clock"?

Junchao-Mellanox: Hi @qiluo-msft , the assumption that `delay == 0` might not be always true. As you can see, the variable `start` is defined out of the while loop, it will not be changed during iterations. So, even if `end - start` is too small for the first few iterations, it does not matter. `end` will increase as iteration goes on, `start` keeps unchanged. So, there will be a point that `end - start` is larger than 1ms.

qiluo-msft:
@ vaibhavhd could you also review?

qiluo-msft (dismissed):

common/restart_waiter.cpp
```diff
@@ -0,0 +1,158 @@
...
+        auto end = std::chrono::steady_clock::now();
+        int delay = static_cast<int>(
+                    std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count());
+
+        if (delay >= static_cast<int>(maxWaitSec) * 1000)
+        {
+            return false;
+        }
+
+        selectTimeout -= delay;
```
qiluo-msft, line 134: >delay [](http://example.com/codeflow?start=25&length=5)

Suspecting a bug: `delay` is measured from a fixed start time, you should not decrease by `delay` in a loop.

Junchao-Mellanox: Not sure I understand. Let's say `maxWaitTime` is 90 seconds, thus initial `selectTimeout` is 90000 ms. For first iteration, restart is not done, and delay is 10 seconds. To make sure `maxWaitTime` is 90,  the `selectTimeout` should be adjusted 90 - 10 = 80 seconds because we have already waited 10 seconds.

common/restart_waiter.h
```diff
@@ -0,0 +1,33 @@
...
+
+    static bool waitWarmBootDone(unsigned int maxWaitSec = 180,
+                                    unsigned int dbTimeout = 0,
+                                    bool isTcpConn = false);
+
+    static bool waitFastBootDone(unsigned int maxWaitSec = 180,
+                                    unsigned int dbTimeout = 0,
+                                    bool isTcpConn = false);
+
+    static bool isWarmOrFastBootInProgress(swss::DBConnector &stateDb);
```
qiluo-msft, line 24: >isWarmOrFastBootInProgress [](http://example.com/codeflow?start=16&length=26)

-> isAdvancedBootInProgress

common/redis_table_waiter.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include "select.h"
+#include "subscriberstatetable.h"
+
+using namespace swss;
+
+bool RedisTableWaiter::waitUntil(
+    DBConnector &db,
+    const std::string &tableName,
+    unsigned int maxWaitSec,
+    CheckFunc checkFunc)
```
qiluo-msft, line 11: >CheckFunc [](http://example.com/codeflow?start=4&length=9)

CheckFunc&

common/redis_table_waiter.cpp
```diff
@@ -0,0 +1,145 @@
...
+    return false;
+}
+
+bool RedisTableWaiter::waitUntilFieldSet(
+    DBConnector &db,
+    const std::string &tableName,
+    const std::string &key,
+    const std::string &fieldName,
+    unsigned int maxWaitSec,
+    ConditionFunc cond)
```
qiluo-msft, line 73: >ConditionFunc cond [](http://example.com/codeflow?start=4&length=18)

ConditionFunc &cond

common/redis_table_waiter.h
```diff
@@ -0,0 +1,43 @@
...
+                          const std::string &tableName,
+                          const std::string &key,
+                          unsigned int maxWaitSec);
+
+    static bool waitUntilKeyDel(DBConnector &db,
+                          const std::string &tableName,
+                          const std::string &key,
+                          unsigned int maxWaitSec);
+
+    static bool waitUntil(
```
qiluo-msft, line 35: >waitUntil [](http://example.com/codeflow?start=16&length=9)

private function?

Junchao-Mellanox: Hi, I was thinking to allow user providing the `checkFunc` to do something customize. Do you think that it is necessary?

common/redis_table_waiter.cpp
Junchao-Mellanox: Fixed

common/restart_waiter.h
Junchao-Mellanox: Fixed

common/redis_table_waiter.cpp
Junchao-Mellanox: Fixed

qiluo-msft (approved):

Junchao-Mellanox:
Hi @vaibhavhd @liuh-80 @stephenxs @arfeigin , could you please review and sign-off?

stephenxs (approved):

common/restart_waiter.cpp
```diff
@@ -0,0 +1,92 @@
...
+using namespace swss;
+
+static const std::string STATE_DB_NAME = "STATE_DB";
+static const std::string STATE_DB_SEPARATOR = "|";
+static const std::string RESTART_KEY = "system";
+static const std::string RESTART_ENABLE_FIELD = "enable";
+static const std::string FAST_REBOOT_TABLE_NAME = "FAST_REBOOT";
+
+// waitAdvancedBootDone
+bool RestartWaiter::waitAdvancedBootDone(
```
vaibhavhd, line 17: I have a generic comment: do we necessarily need to have 3 different wait mechanisms here?

I think just waiting on AdvancedBoot can be sufficient for both the usecases of delaying startup (for performance reasons) in fast OR warmboot.

Is there a usecase where services are required to wait differently during fast or warm start of the system?

If there is no such usecase, for simplicity purposes, I would suggest to have only one wait mechanism -- wait on `STATE_WARM_RESTART_ENABLE_TABLE_NAME` as that's the system wide flag enabled for both the cases.

Junchao-Mellanox: Hi @vaibhavhd , I was told that there is assumption that services could be relevant for fastboot, but not relevant for warmboot, or vice versa. Even if there is no use case for now, I suppose this assumption is reasonable because fastboot and warmboot have different requirement for traffic forwarding. So, I implemented 3 functions here.

vaibhavhd (approved):

PR #700, opened 2022-10-18
Non recursive automake and Debian packaging changes

saiarcot895:
This PR brings in three major changes:

1. Switch to using non-recursive automake. The primary advantage of this is that builds for source files across subdirectories can happen in parallel, which reduces the build time on systems with sufficient cores. On my dev VM, the build time went from about 2:55 to 2:28 for the Bullseye build.
2. Update the Debian packaging files to a more recent syntax/features supported in Debian Buster (and newer).
3. Disable building static libraries. The build time dropped to about 1:30 on my dev VM.


qiluo-msft (dismissed):
LGTM. Please wait for other reviewers.

qiluo-msft (approved):

saiarcot895:
Requesting cherry-pick to 202205 and 202211 for space reduction purposes (removal of static libraries).

yxieca:
@saiarcot895 please create separate PR for 202205 branch.

PR #703, opened 2022-10-26
Update sairedis build command

saiarcot895:
The build rules for sairedis have been simplified in sonic-net/sonic-sairedis#1116. As a result, build targets have changed. Update the build step here accordingly.


xumia (dismissed):

prsunny:
@saiarcot895 , seems I don't have more context on this change. Its good to get @xumia to signoff

PR #708, opened 2022-11-11
[NotificationProducer] add pipeline support

stepanblyschak:
Added a pipeline support for *NotificationProducer*. *NotificationProducer* created with *RedisPipeline* will buffer published notifications untill a *flush* is called on *RedisPipeline* or *RedisPipeline*'s command buffer is filled.

The motivation for this change is to optimize redis usage when using *ResponsePublisher* from sonic-swss for a lot of entries.


stepanblyschak:
Fix for sairedis tests - https://github.com/sonic-net/sonic-sairedis/pull/1169

liat-grozovik:
@marian-pritsak @prsunny could you please help to review?

BUILD
```diff
@@ -10,6 +10,7 @@ cc_library(
         "common/*.hpp",
     ]),
     copts = [
+        "-std=c++14",
```
qiluo-msft, line 13: >c++14 [](http://example.com/codeflow?start=14&length=5)

Other places are using c++11. Is it a concern?

common/notificationproducer.cpp
```diff
@@ -1,7 +1,14 @@
 #include "notificationproducer.h"
 
+#define NON_BUFFERED_COMMAND_BUFFER_SIZE 1
+
 swss::NotificationProducer::NotificationProducer(swss::DBConnector *db, const std::string &channel):
-    m_db(db), m_channel(channel)
+    m_ownedpipe(std::make_unique<swss::RedisPipeline>(db, NON_BUFFERED_COMMAND_BUFFER_SIZE)), m_pipe(m_ownedpipe.get()), m_channel(channel), m_buffered(false)
```
qiluo-msft, line 6: >m_ownedpipe [](http://example.com/codeflow?start=101&length=11)

You are assume m_pipe initialized after m_ownedpipe initialized. The sequence is not well known knowledge. Suggest you move `m_pipe` to function body for simplicity purpose.

common/notificationproducer.cpp
```diff
@@ -19,5 +26,19 @@ int64_t swss::NotificationProducer::send(const std::string &op, const std::strin
 
     SWSS_LOG_DEBUG("channel %s, publish: %s", m_channel.c_str(), msg.c_str());
 
-    return m_db->publish(m_channel, msg);
+    RedisCommand command;
+    command.format("PUBLISH %s %s", m_channel.c_str(), msg.c_str());
+
+    if (m_buffered)
+    {
+        m_pipe->push(command, REDIS_REPLY_INTEGER);
```
qiluo-msft, line 34: >m_pipe [](http://example.com/codeflow?start=8&length=6)

You may use `m_ownedpipe` directly. Are you defining two member variables for better performance? If yes, I prefer one variable for simplicity.

common/notificationproducer.cpp
```diff
@@ -19,5 +26,19 @@ int64_t swss::NotificationProducer::send(const std::string &op, const std::strin
...
+
+        // if operating in buffered mode return -1 as we can't know the number
+        // of client's that have read the message immediately
+        return -1;
```
qiluo-msft, line 38: >-1 [](http://example.com/codeflow?start=15&length=2)

Is it better to use `0`?

qiluo-msft:
@bocon13 Could you also help review?

BUILD
stepanblyschak: @qiluo-msft I don't see a problem. Everything builds successfully. I don't think other projects need to stay with c++11. c++14 has been around for a long time and adopted by compiler. As long as the headers are c++11 compatible and ABI does not change - everything works.

common/notificationproducer.cpp
stepanblyschak: The order of member initialization is defined by the standard and is well known. Member variables are initialized in the order they are defined in the class definition. If by any chance members are reordered in the definition any modern compiler will raise a warning (-Wreorder in gcc) and many of them should be treated as errors. Since we use -Wall + -Werror it is going to be catched.

common/notificationproducer.cpp
stepanblyschak: @qiluo-msft NotificationProducer has two invariants 1) Created from DBConnector and having it's own RedisPipeline in ```m_ownedpipe``` smart pointer and ```m_pipe``` pointing to it 2) NotificationProducer created from RedisPipeline which is not owned - ```m_ownedpipe``` is ```nullptr```, ```m_pipe``` is a non-owning raw pointer to RedisPipeline object.
So, ```m_pipe``` is always pointing to RedisPipeline that needs to be used regardless of how NotificationProducer has been created.

I see the same pattern beeing used for Table class (https://github.com/sonic-net/sonic-swss-common/blob/master/common/table.cpp#L38) and it was a motivation to follow the same desing.

However, the table class uses raw pointer and manual memory allocation/deallocation + a flag to indicate wether we own the RedisPipeline object. So I came up with a bit simpler approach.

common/notificationproducer.cpp
stepanblyschak: @qiluo-msft As stated by the comment on send() method the return code does not indicate the success or failure but the number of clients that read the message. In buffered mode we can't imidatelly tell how many clients read the message, so returning 1 might fool the user of this API that his consumer has already read message.

BUILD
qiluo-msft: Is it a blocking issue?
If the change is really needed, I prefer changing the c++NN option in a consistent way, and change dependent libraries/applications first (swss->sairedis->swss-common). And change pyext/py[23]/Makefiles.am at the same time.

stepanblyschak: sairedis is c++14 - https://github.com/sonic-net/sonic-sairedis/blob/master/configure.ac#L121
swss is c++14 - https://github.com/sonic-net/sonic-swss/blob/master/configure.ac#L57
I believe not every app using swss-common is now c++14, not to mention app extensions.
My point is that swss-common may use c++-14 in .cpp but must have c++-11 compatible headers for those apps that might not be able to migrate yet. Python swss-common bindings compile a wrapper cpp based of the headers, that's why py[23]/Makefile.am aren't changed and stay c++-11.

qiluo-msft (approved):

PR #714, opened 2022-11-22
Use SWIGPYTHON to improve SWIG for GO wrapper.

ganglyu:
#### Why I did it
Use SWIG to generate GO wrapper for sonic-swss-common, and SWIG report syntax error for "%pythoncode%".
Related lines are only used for python wrapper, so we should use "#ifdef SWIGPYTHON".

#### How I did it
Use SWIGPYTHON to improve SWIG for python wrapper.

#### How to verify it
Build succeed and pass all UT.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Use SWIGPYTHON to improve SWIG, and then SWIG can generate GO wrapper.


liuh-80 (approved):

common/configdb.h
```diff
@@ -37,6 +37,7 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
     std::string m_db_name;
 };
 
+#ifdef SWIG
```
liuh-80, line 40: suggest use: #if defined(SWIG) && defined(SWIGPYTHON)

liuh-80 (dismissed):

qiluo-msft:
> class SonicDBConfig

suggest add a minimal go unit test in swss-common repo. The useful test case may be sonicdbconfig init

---
Refers to: common/dbconnector.h:38 in 974b61c. [](commit_id = 974b61c37ea735370d7c328ca43d6af6d786a8f5, deletion_comment = False)

goext/swsscommon.i
```diff
@@ -0,0 +1,13 @@
+%module swsscommon
```
qiluo-msft, line 1: Is it possible to reuse pyext/swsscommon.i?

ganglyu: pyext/swsscommon.i is designed for python, and it has embedded python code:
https://github.com/sonic-net/sonic-swss-common/blob/master/pyext/swsscommon.i#L108
I think use another swig file will be better.

.azure-pipelines/build-template.yml
```diff
@@ -138,6 +138,9 @@ jobs:
         pytest --cov=. --cov-report=xml
         mv coverage.xml tests/coverage.xml
         gcovr -r ./ -e ".*/swsscommon_wrap.cpp" --exclude-unreachable-branches --exclude-throw-branches -x --xml-pretty  -o coverage.xml
+        redis-cli FLUSHALL
```
qiluo-msft, line 141: >redis-cli FLUSHALL [](http://example.com/codeflow?start=8&length=18)

I guess you flush redis for test purpose, so it should be just above `make -C goext check` line.

goext/Makefile
```diff
@@ -0,0 +1,25 @@
...
+
+SWIG_FLAG = -go -cgo -c++ -intgosize 64
+ifeq ($(CONFIGURED_ARCH),arm64)
+SWIG_FLAG += -DSWIGWORDSIZE64
+endif
+
+.PHONY: all check clean
+
+all:
+	-$(LN) -s ../pyext/swsscommon.i swsscommon.i
```
qiluo-msft, line 17: >- [](http://example.com/codeflow?start=1&length=1)

why `-`?

.azure-pipelines/build-template.yml
```diff
@@ -131,13 +131,17 @@ jobs:
...
         sudo dpkg -i libswsscommon_*.deb
+        sudo dpkg -i libswsscommon-dev_*.deb
         sudo dpkg -i python-swsscommon_*.deb
 
         ./tests/tests
         redis-cli FLUSHALL
         pytest --cov=. --cov-report=xml
         mv coverage.xml tests/coverage.xml
         gcovr -r ./ -e ".*/swsscommon_wrap.cpp" --exclude-unreachable-branches --exclude-throw-branches -x --xml-pretty  -o coverage.xml
+        make -C goext
```
liuh-80, line 142: As my undertsand, following code are not necessary:

        sudo dpkg -i libswsscommon-dev_*.deb
        make -C goext

Because before we run the test, swsscommon go lib already build to some deb package, so we only need build and run test here?

If the go lib not include in any deb package, then you can't use it in any other project, and also sonic-buildimage may break because that repo using a package cache, so in most case the swsscommon will not build, just install deb from cache.

pyext/swsscommon.i
```diff
@@ -49,7 +49,9 @@
 %include <std_vector.i>
 %include <std_pair.i>
 %include <std_map.i>
+#ifdef SWIGPYTHON
```
liuh-80, line 52: The std_shared_ptr.i is not only for python.
swsscommon return some shared_ptr<string> in public API.
Not sure if not include this in go, there will memory leak happen.

pyext/swsscommon.i
```diff
@@ -64,6 +66,7 @@
 %template(GetConfigResult) std::map<std::string, std::map<std::string, std::map<std::string, std::string>>>;
 %template(GetInstanceListResult) std::map<std::string, swss::RedisInstInfo>;
 
+#ifdef SWIGPYTHON
 %exception {
     try
```
liuh-80, line 71: All these codes are for python so maybe you need create UT for counterpart of go, and make sure the go API does not have memory leak or return value issue.

Suggest use blame to check the history of these code and create go version UT to cover the scenario, for example for the _FieldValueMap__get, you can find this change:
https://github.com/sonic-net/sonic-swss-common/commit/40b255b7dae51a0ee99bb3bbdb663aa78c9120ed#diff-e95e67eabf12f4e97181558babbde4d4a1d320173950633a6534fd0a7ca72999

then maybe add a UT to call an API which return FieldValuePair, then validate get field from FieldValuePair can work.

.azure-pipelines/build-template.yml
ganglyu: We find that origin swsscommon library can't generate GO wrapper swig, and this unit test will prevent future commit from blocking GO wrapper.
For sonic-gnmi, we will install libswsscommon-dev, and generate GO wrapper with swig, so we don't need go lib.

pyext/swsscommon.i
ganglyu: SWIG go does not support shared_ptr, and for now we don't need public API using shared_ptr for go.

pyext/swsscommon.i
ganglyu: I have adde UT

liuh-80 (approved):

qiluo-msft:
Looks good to me.

PR #716, opened 2022-11-24
Update sonic-swss-common pipeline build for Bullseye

saiarcot895:
With docker-sonic-vs now upgraded to Bullseye (in sonic-net/sonic-buildimage#13294), make the necessary changes in the sonic-swss-common pipeline build to build for and use Bullseye.


liushilongbuaa:
@saiarcot895 , https://github.com/sonic-net/sonic-swss-common/blob/e3d9024602847d5bb41e14f153099431ef35f291/debian/control#L5
Why this line doesn't need change?

.azure-pipelines/docker-sonic-vs/Dockerfile
```diff
@@ -4,16 +4,4 @@ ARG docker_container_name
...
-RUN dpkg -i /debs/python3-swsscommon_1.0.0_amd64.deb
-RUN dpkg -i /debs/sonic-db-cli_1.0.0_amd64.deb
-
-RUN dpkg -i /debs/libsaimetadata_1.0.0_amd64.deb
-RUN dpkg -i /debs/libsairedis_1.0.0_amd64.deb
-RUN dpkg -i /debs/libsaivs_1.0.0_amd64.deb
-RUN dpkg -i /debs/syncd-vs_1.0.0_amd64.deb
-
-RUN dpkg -i /debs/swss_1.0.0_amd64.deb
+RUN dpkg -i /debs/libswsscommon_1.0.0_amd64.deb /debs/python3-swsscommon_1.0.0_amd64.deb /debs/sonic-db-cli_1.0.0_amd64.deb /debs/libsaimetadata_1.0.0_amd64.deb /debs/libsairedis_1.0.0_amd64.deb /debs/libsaivs_1.0.0_amd64.deb /debs/syncd-vs_1.0.0_amd64.deb /debs/swss_1.0.0_amd64.deb
```
liushilongbuaa, line 7: libswsscommon depends on libboost1.71-dev. (in debian/control Line5)
But docker-sonic-vs uninstalled libboost1.71. This line should fail. @saiarcot895 can you rerun the pipeline? Maybe the former PR check used old docker-sonic-vs image.

liushilongbuaa:
I didn't see this PR. So I have a fix for platform-daemons. https://github.com/sonic-net/sonic-platform-daemons/pull/349/files
It is running in sonic-slave-bullseye, so it succeeded.
If it run in docker-sonic-vs, it fails. https://github.com/sonic-net/sonic-sairedis/pull/1227/files

.azure-pipelines/docker-sonic-vs/Dockerfile
saiarcot895: The dependency string is specified as `libboost-dev | libboost1.71-dev`, which means either `libboost-dev` or `libboost1.71-dev` needs to be installed. Since `libboost-dev` is installed in Bullseye, the dependency is met.

In Buster, Boost 1.71 is installed through the `libboost1.71-dev` (and associated) packages. The Buster default is Boost 1.67, and I guess something explicitly wanted to use Boost 1.71, so all packages are built with that. Because libswsscommon might still get built for Buster (it's needed for the nat and sflow containers), that part of the dependency string cannot be removed just yet.

liushilongbuaa: /mnt$ docker run --rm -ti --entrypoint bash docker-sonic-vs:latest 
root@9f79d2f3aae7:/# apt list| grep libboost

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

libboost-serialization1.74.0/now 1.74.0-9 amd64 [installed,local]

liushilongbuaa: docker-sonic-vs didn't install libboost-dev or libboost1.71-dev

saiarcot895: The docker container shouldn't need the dev package. The dev package is only for building libswsscommon, and should only be present in the slave container.

saiarcot895: Line 5 of `debian/control` lists the build-time dependencies. This does not necessarily correspond to runtime dependencies. The runtime dependency list is automatically generated, and gets filled into the `shlibs:Depends` variable for each package. For the `libswsscommon` package, this dependency list is libboost-serialization1.74.0 (>= 1.74.0), libc6 (>= 2.14), libgcc-s1 (>= 3.0), libhiredis0.14 (>= 0.14.0), libnl-3-200 (>= 3.2.21), libnl-nf-3-200 (>= 3.2.7), libnl-route-3-200 (>= 3.2.7), libstdc++6 (>= 9), libuuid1 (>= 2.16), libyang, libzmq5 (>= 4.0.1+dfsg) (for a package built on Bullseye).

liushilongbuaa (approved):
pipeline succeeded.

saiarcot895:
@liushilongbuaa can you force merge this PR? Looks like the status check on Github is looking at a merge commit instead.

liushilongbuaa:
> @liushilongbuaa can you force merge this PR? Looks like the status check on Github is looking at a merge commit instead.

@xumia, please help

xumia:
> > @liushilongbuaa can you force merge this PR? Looks like the status check on Github is looking at a merge commit instead.
> 
> @xumia, please help

It is weird, the results in the GitHub check page are succeeded.
image

xumia:
@saiarcot895 , could you please try to run /easycla ? I can merge it if the easycla passed.
Or could you please send another PR?

saiarcot895:
I can run EasyCLA, but the merge section of the PR is looking at some other commit. If this fails, I'll push an empty commit.

PR #720, opened 2022-12-03
Revamp module build script to make it work for 5.15 on Ubuntu 20.04

saiarcot895:
The current version of the script does not work at all when the host kernel is Azure's 5.15 kernel on Ubuntu 20.04. It does seem to work with 5.11 and older.

Update the script to add support for building and loading the kernel modules for the 5.15 kernel, while keeping support for the non-HWE kernel.

In addition, use a separate `updates/sonic` directory for the kernel modules we compile, to make it clear they were compiled separately and don't come from the official kernel build packages.

Finally, add a warning if we're compiling a kernel module for a different version of the kernel than what is currently running.


saiarcot895:
@xumia @qiluo-msft could you review this when you get a chance?

.azure-pipelines/build_and_install_module.sh
```diff
@@ -26,64 +26,62 @@ function build_and_install_kmodule()
...
+    fi
     make  allmodconfig
     mv .config .config.bk
     cp /boot/config-$(uname -r) .config
     grep NET_TEAM .config.bk >> .config
-    echo CONFIG_NET_VRF=m >> .config
-    echo CONFIG_MACSEC=m >> .config
-    echo CONFIG_NET_VENDOR_MICROSOFT=y >> .config
-    echo CONFIG_MICROSOFT_MANA=m >> .config
-    echo CONFIG_SYSTEM_REVOCATION_LIST=n >> .config
```
qiluo-msft, line 55: why removing?

saiarcot895: Ideally, when building kernel modules, the config used for building the kernel module must match the system's kernel config. This is especially true for config options that aren't for just building a module (i.e. options that may affect multiple parts of the kernel). If there's a difference, there a risk of structure sizes not matching, or function parameters not matching.

These kernel options, as far as I can tell, and based on the CI test run, don't have any functional impact.

saiarcot895: Ideally, when building kernel modules, the config used for building the kernel module must match the system's kernel config. This is especially true for config options that aren't for just building a module (i.e. options that may affect multiple parts of the kernel). If there's a difference, there a risk of structure sizes not matching, or function parameters not matching.

saiarcot895:
@xumia @liushilongbuaa Could you review this when you get a chance?

xumia (approved):

xumia:
> @xumia @liushilongbuaa Could you review this when you get a chance?

LGTM, thanks for the fix.

PR #723, opened 2022-12-07
keep del when del-set-sequence exist

inspurSDN:
what I did:  
    when del-set-sequence exist in consumer table, I keep the del event been execute even a set of the same key is already seen after the del.
why I did:
    in the old code, the del is setting an key without fv, so if a set with fv after, it will only execute a set, so the multimap m_toSync can not  seen the del event, sometimes the lost del may result errors if something is depending on the del event.


inspurSDN:
failed on sairedis? seems not what I modified.
retest please

inspurSDN:
retest please

PR #733, opened 2023-01-04
[hash]: Add GH DB schema

nazariig:
Signed-off-by: Nazarii Hnydyn <nazariig@nvidia.com>

**HLD:** https://github.com/sonic-net/SONiC/pull/1101


common/schema.h
```diff
@@ -421,7 +422,7 @@ namespace swss {
 
 /***** STATE DATABASE *****/
 
-#define STATE_SWITCH_CAPABILITY_TABLE_NAME          "SWITCH_CAPABILITY_TABLE"
+#define STATE_SWITCH_CAPABILITY_TABLE_NAME          "SWITCH_CAPABILITY"
```
prsunny, line 425: Please revert this change and follow the existing convention of _TABLE

nazariig: @prsunny let me clarify the need of this change:

Today in SONiC define `STATE_SWITCH_CAPABILITY_TABLE_NAME` (which supposed to be a reference point to the switch capabilities table) is currently not used. Instead, we have a lot of places with hardcoded  string `SWITCH_CAPABILITY`. The motivation of this change is to start using it for switch hash OA with the relevant table name.

Once this merged, we can raise a separate PR and replace all hardcoded values (python, C++, etc.) with the relevant named definition. And only then change the table name to be following the convention you mentioned.
Since this is a separate activity, i would prefer to avoid doing it as part of current PR.

Please let me know what do you think.

prsunny: ok

prsunny (approved):

StormLiangMS:
@nazariig  looks like new feature, pls wait for next branch cut.

@prsunny  for vis.

PR #736, opened 2023-01-09
RedisPipeline ignore flush when call dtor from another thread.

liuh-80:
#### Why I did it
Because RedisPipeline is not thread safe, in c++ if code call exit() from another thread, the dtor of RedisPipeline will  be called from another thread and trigger a race condition issue.

#### How I did it
RedisPipeline ignore flush when call dtor from another thread.

#### How to verify it

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

RedisPipeline ignore flush when call dtor from another thread.

### DIFF
--- a/common/redispipeline.h
+++ b/common/redispipeline.h
@@ -6,6 +6,7 @@
 #include "redisreply.h"
 #include "rediscommand.h"
 #include "dbconnector.h"
+#include "logger.h"
 
 #include "unistd.h"
 #include "sys/syscall.h"
@@ -31,6 +32,10 @@ public:
         {
             // call flush from different thread will trigger race condition issue.
             flush();
+        }
+        else
+        {
+            SWSS_LOG_NOTICE("RedisPipeline dtor is called from another thread, possibly due to exit(), Database: %s", getDbName().c_str());
         }
 
         delete m_db;
--- a/common/redispipeline.h
+++ b/common/redispipeline.h
@@ -7,6 +7,10 @@
 #include "rediscommand.h"
 #include "dbconnector.h"
 
+#include "unistd.h"
+#include "sys/syscall.h"
+#define gettid() syscall(SYS_gettid)
+
 namespace swss {
 
 class RedisPipeline {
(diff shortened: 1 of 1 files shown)


common/redispipeline.h
```diff
@@ -7,6 +7,10 @@
 #include "rediscommand.h"
 #include "dbconnector.h"
 
+#include "unistd.h"
+#include "sys/syscall.h"
+#define gettid() syscall(SYS_gettid)
```
liuh-80, line 12: there is no wrapper for gettid() in glibc <= 2.30, so we need use syscall(SYS_gettid)

common/redispipeline.h
```diff
@@ -19,10 +23,16 @@ class RedisPipeline {
...
+        initializeOwnerTid();
     }
 
     ~RedisPipeline() {
-        flush();
+        if (m_ownerTid == gettid())
+        {
+            // call flush from different thread will trigger race condition issue.
+            flush();
+        }
```
qiluo-msft, line 34: >} [](http://example.com/codeflow?start=8&length=1)

Add a notice logging in `else` branch.

common/redispipeline.h
```diff
@@ -19,10 +23,16 @@ class RedisPipeline {
         , m_remaining(0)
     {
         m_db = db->newConnector(NEWCONNECTOR_TIMEOUT);
+        initializeOwnerTid();
     }
 
     ~RedisPipeline() {
```
qiluo-msft, line 29: >RedisPipeline [](http://example.com/codeflow?start=5&length=13)

Thanks for fixing the bug. It is from the design that DBConnector could be used only in one thread. Do you find similar issues in other classes in swss-common?

liuh-80: No, I check all dtor in swss-common only find RedisPipeline class run redis command in dtor.

common/redispipeline.h
liuh-80: Fixed

qiluo-msft (approved):

qiluo-msft:
Do we need backport, for example to 202012 branch?

PR #725, opened 2022-12-19
Monitor tables for Sirius

siqbal1986:
Added 2 VNET_MONITOR tables. one in APP DB and the other in State DB for Vxlan tunnel endpoint custom monitor.

PR #738, opened 2023-01-13
[202012] Monitor tables for VNET custom monitoring

siqbal1986:
Case of Orignal PR https://github.com/sonic-net/sonic-swss-common/pull/725


prsunny (approved):

prsunny:
Cherring-picking #725

PR #739, opened 2023-01-19
Increase the netlink buffer size from 3MB to 16MB.

kishorekunal01:
As error is seen in fdbsyncd with netlink reports "out of memory on reading a netlink socket" It is seen when kernel is sending 10k remote mac to fdbsyncd.

- What I did
    Increase the buffer size of the netlink buffer from 3MB to 16MB

- How I did it
    Increase the buffer size in the API nl_socket_set_buffer_size

- How to verify it
    Verified with 10k remote mac, and restarting the fdbsyncd process. So that kernel send the bridge fdb dump to the fdbsyncd.
    Verified that the netlink buffer error is not reported in the sys log.

Signed-off-by: kishore.kunal@broadcom.com

### DIFF
--- a/common/netlink.cpp
+++ b/common/netlink.cpp
@@ -35,8 +35,8 @@ NetLink::NetLink(int pri) :
     }
 
     nl_socket_set_nonblocking(m_socket);
-    /* Set socket buffer size to 3MB */
-    nl_socket_set_buffer_size(m_socket, 3145728, 0);
+    /* Set socket buffer size to 16MB */
+    nl_socket_set_buffer_size(m_socket, 16777216, 0);
 }
 
 NetLink::~NetLink()


dgsudharsan (approved):

snider-nokia:
I suspect there is a problem here in terms of completeness.

In my testing related to https://github.com/sonic-net/sonic-buildimage/issues/12587 I verified that the netlink socket rcv buffer size is inevitably constrained by the host net.core.rmem_max setting.  So, no matter what size is requested in netlink.cpp the associated socket rcv buffer size will never exceed the host net.core.rmem.max setting.

The preferable method, it seems, for manipulating the aforementioned host setting is by changing it in **.../files/image_config/sysctl/sysctl-net.conf** (as snapshot below) to the appropriate value (in this case 16MB).  Once the host value is appropriately changed (when database is instantiated) then the subsequent netlink socket rcv buffer size increase will be honored.

Suggest also that care should be taken to run a full OC test suite regression after this (host net.core.rmem_max) change, since a significant amount of additional memory is going to be allocated for socket rcv buffers that was not previously.  It is possible that other (non-netlink) sockets may also be allocating more socket rcv buffer memory after the host change, as well, since any existing code that is currently attempting to allocate a socket rcv buffer larger than host net.core.rmem_max setting (currently set to 3MB) is being constrained to that (3MB) value.  After the host net.core.rmem_max change now any/all legacy code that requests more than 3MB is going to actually be granted what it wants (up to the new ceiling of 16MB) and it is possible that this could subsequently result in unanticipated memory pressure.

image

PR #740, opened 2023-01-19
Fix minor issue found during static code scan

DavidZagury:
Fix issue of using uninitialized value on unit tests.


prsunny (dismissed):

keboliu:
@prsunny @qiluo-msft  would you be able to merge this PR?

common/ipprefix.h
```diff
@@ -76,7 +76,8 @@ class IpPrefix
                 {
                     assert(mid >= 0 && mid < 16);
                     ipa.ip_addr.ipv6_addr[mid] = (uint8_t)(0xFF << (8 - bits));
-                    memset(ipa.ip_addr.ipv6_addr + mid + 1, 0, 16 - mid - 1);
+                    if (mid < 15)
```
qiluo-msft, line 79: >if (mid < 15) [](http://example.com/codeflow?start=20&length=13)

Wondering who is complaining? the compiler?

liat-grozovik:
@qiluo-msft any further comments or can we merge this one?

common/ipprefix.h
qiluo-msft: This error does not make sense, it indicates the limitation of scanner/linter you are using.
I suggest not to sacrifice code concision to workaround scanner’s issue. If you like, you can add a code comment to suppress scanner negative alarm.

qiluo-msft (approved):

PR #544, opened 2021-10-22
[CI]: Swap the python code coverage report with the cpp report

xumia:
Show CPP report by default.

PR #746, opened 2023-02-06
[CI][202012]: Swap the python code coverage report with the cpp report (#544)

xumia:
Change to show CPP code full coverage, make the coverage rate the same as it displayed in the coverage page.


liuh-80:
This PR is for manually cherry-pick https://github.com/sonic-net/sonic-swss-common/pull/544

liuh-80 (commented):
gcovr: error: unrecognized arguments: --exclude-throw-branches

PR #766, opened 2023-03-13
[logger] Add public `restartLogger` to restart logger thread

lolyu:
This is to support adding custom observers to the swss logger.
With `restartLogger`, now applications could start logger with the following lines:
```
linkToDb(demoDb, myCustomPrioNotify, defaultPrio);
restartLogger()
```

Signed-off-by: Longxiang Lyu <lolv@microsoft.com>

### DIFF
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -166,6 +166,11 @@ void Logger::linkToDbNative(const std::string& dbName, const char * defPrio)
     getInstance().restartSettingThread();
 }
 
+void Logger::restartLogger()
+{
+    getInstance().restartSettingThread();
+}
+
 Logger& Logger::getInstance()
 {
     static Logger m_logger;
--- a/common/logger.h
+++ b/common/logger.h
@@ -91,6 +91,7 @@ public:
     static void linkToDb(const std::string& dbName, const PriorityChangeNotify& notify, const std::string& defPrio);
     // Must be called after all linkToDb to start select from DB
     static void linkToDbNative(const std::string& dbName, const char * defPrio="NOTICE");
+    static void restartLogger();
     void write(Priority prio, const char *fmt, ...)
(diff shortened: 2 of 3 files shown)


common/logger.cpp
```diff
@@ -166,6 +166,11 @@ void Logger::linkToDbNative(const std::string& dbName, const char * defPrio)
     getInstance().restartSettingThread();
 }
 
+void Logger::restartLogger()
```
yxieca, line 169: Who will call this function other than unit tests?

lolyu: This PR depends on the change: https://github.com/sonic-net/sonic-linkmgrd/pull/182
The reason is if we want to support swss logger for `linkmgrd`, the log message path will be:
```
mux log messages -> log macros[0] -> mux logger[1] -> swsslogger[2]
                                             |----------> extra log file[3]
```
Mux log messages are collected by mux logger[1] first, then directed to swsslogger[2]. The reason for keeping the mux logger is that it has early-stage filtering inside those log macros[0] and it is needed by the `linkmgrd` ut to save the log to an extra log file.
Currently, `linkmgrd` could only use the default swss log level observer function, and it only changes the log level of swsslogger[2], so all mux log messages will be forwarded through the log message path [0] -> [1] -> [2], this will double the CPU usage of linkmgrd from 10% average to 20%.
So this PR is to support custom swss log level observer function, `linkmgrd` could use its own log level observer function so after user modifies log level via `swssloglevel`, both mux logger[1] and swsslogger[2] log level will be updated, the early stage filtering inside log macros[0] could drop most of the logs so the mux logger[1] and swsslogger[2] will not process them anymore.

lolyu: @yxieca, could you please take a review? Thanks!

yxieca (approved):

PR #791, opened 2023-06-05
Add update_only in SubscriberStateTable.

mint570:
If the flag is set, the API will return empty attribute (instead of full list of attributes) when there is no change for SET requests. This will be useful when CONFIG DB is updated with no actual change. After setting the flag, the user of the SubscriberStateTable will do no-op when empty attribute is return.


mint570:
@qiluo-msft do you mind reviewing this PR please? Thanks.

qiluo-msft (approved):

qiluo-msft:
> @qiluo-msft I prepared a swss PR: [sonic-net/sonic-swss#3741](https://github.com/sonic-net/sonic-swss/pull/3741)
> 
> The two PRs will need to be merged together.

It will be tricky to merge 2 PR together since PR checker failure. Could you figure out a path to merge one PR at one time?

PR #792, opened 2023-06-08
Add SEND_TO_INGRESS port table in schema.h

baxia-lan:
HLD: https://github.com/sonic-net/SONiC/blob/master/doc/pins/send_to_ingress_hld.md


baxia-lan:
@prsunny PTAL

baxia-lan:
@mint570

PTAL

common/schema.h
```diff
@@ -271,6 +272,9 @@ namespace swss {
 #define CFG_PORT_TABLE_NAME           "PORT"
 #define CFG_PORT_CABLE_LEN_TABLE_NAME "CABLE_LENGTH"
 
+#define CFG_SEND_TO_INGRESS_PORT_TABLE_NAME  "SEND_TO_INGRESS_PORT"
+#define SEND_TO_INGRESS_PORT_NAME            "SEND_TO_INGRESS"
```
prsunny, line 276: Please prefix this with CFG_. Also is this a new table in Config_DB? if so, what is the above one for?

baxia-lan: Removed this port name.
It was used as hardcoded send_to_ingress port name.

prsunny (approved):

wen587:
Hi @baxia-lan , could you add YANG model for SEND_TO_INGRESS_PORT?

PR #797, opened 2023-06-16
Force use unix domain socket when namespace not empty

liuh-80:
Force use unix domain socket when namespace not empty

#### Work item tracking
Microsoft ADO (number only): 21776191

#### Why I did it
Change cli behavior same with python version.

#### How I did it
Force use unix domain socket when namespace not empty

#### How to verify it
Add UT.
Pass all UT.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Force use unix domain socket when namespace not empty


liuh-80:
Use this PR to test the cli chaneg can pass all E2E test:
https://github.com/sonic-net/sonic-buildimage/pull/15509

sonic-db-cli/sonic-db-cli.cpp
```diff
@@ -284,14 +284,11 @@ int sonic_db_cli(
         if (!netns.empty())
         {
             initializeGlobalConfig();
+
+            // Use the unix domain connectivity if namespace not empty.
+            isTcpConn = false;
```
qiluo-msft, line 289: >false [](http://example.com/codeflow?start=24&length=5)

Will you add unit test?

liuh-80: These code already cover by UT:
https://github.com/sonic-net/sonic-swss-common/blob/master/tests/cli_ut.cpp

UT failed after code change, working on fix it.

liuh-80:
PR validation failed:
https://github.com/sonic-net/sonic-buildimage/pull/15509

This is because swss-common code change break sonic cli:
https://github.com/sonic-net/sonic-swss-common/pull/789

Will wait for that issue fix first.

liuh-80:
Will run validation again after https://github.com/sonic-net/sonic-swss-common/pull/804 merged

tests/redis_multi_db_ut_config/database_config0.json
```diff
@@ -3,7 +3,7 @@
         "redis":{
             "hostname" : "127.0.0.1",
             "port": 6379,
-            "unix_socket_path": "/var/run/redis0/redis.sock"
+            "unix_socket_path": "/var/run/redis/redis.sock"
```
qiluo-msft, line 6: >redis [](http://example.com/codeflow?start=42&length=5)

why changing test data?

qiluo-msft:
Please add PR description.

tests/redis_multi_db_ut_config/database_config0.json
liuh-80: This value is not correct.
Our UT never using unix_socket_path before.
But when change sonic-db-cli to user unix domian socket. some UT failed, so need this this value.

liuh-80:
> Please add PR description.

Fixed.

tests/redis_multi_db_ut_config/database_config0.json
qiluo-msft: I agree that we need to fix test code/data if it just start under testing.
This is used in multi-db feature, and you can see a pattern that database_configN.json will have a path /var/run/redisN/redis.sock. So if it is not working due to missing sock file, we need to start multiple redis-server, each with proper config on sock files.

liuh-80: Fixed, keeps the original config no change and add new config file for asic2 and asic3.

qiluo-msft (approved):

rlhui:
@qiluo-msft , @liuh-80 , chassis community thinks that this change may bring bring back the multi-asic show commands issues as in https://github.com/sonic-net/sonic-buildimage/issues/6982, please share your thoughts and/or whether these show commands tests are passing (these should be  in multi-asic taccas sonic-mgmt tests)

PR #801, opened 2023-06-26
Producer/Consumer table binary support

Pterosaur:
We want to make the Producer/Consumer table can support binary messages, The native C string (char *) will be replaced to pointers and its lengths in all paths.
Meanwhile, the Python interfaces of SWIG can only handle the type, str, with UTF-8. So, we need to specialize the SWIG interfaces from bytes of Python to string of C++.


common/luatable.cpp
```diff
@@ -57,10 +57,12 @@ bool LuaTable::get(const vector<string> &luaKeys, vector<FieldValueTuple> &value
     // Transform data structure
     vector<const char *> args1;
     transform(args.begin(), args.end(), back_inserter(args1), [](const string &s) { return s.c_str(); } );
+    vector<size_t> lengths;
```
liuh-80, line 60: Why size_t, as my undertsand the the binary is a byte array, so it should be 'char'

Pterosaur: This vector keeps the lengths of above `args1`.

common/rediscommand.h
```diff
@@ -70,6 +70,7 @@ class RedisCommand {
 
 private:
     char *temp;
+    int len;
```
qiluo-msft, line 73: I see a new class `RedisString`.

common/redisapi.h
```diff
@@ -90,9 +90,15 @@ static inline std::set<std::string> runRedisScript(RedisContext &ctx, const std:
             args.end(),
             std::back_inserter(c_args),
             [](const std::string& s) { return s.c_str(); } );
+    std::vector<size_t> lengths;
+    transform(
```
qiluo-msft, line 94: >transform [](http://example.com/codeflow?start=4&length=9)

Instead of two transform, better to use one for-loop.

common/luatable.cpp
```diff
@@ -57,10 +57,12 @@ bool LuaTable::get(const vector<string> &luaKeys, vector<FieldValueTuple> &value
...
+    transform(args.begin(), args.end(), back_inserter(lengths), [](const string &s) { return s.length(); } );
```
qiluo-msft, line 61: >transform [](http://example.com/codeflow?start=4&length=9)

Instead of two transform, better to use one for-loop.

common/redisapi.h
Pterosaur: Any difference? Better readability?

common/rediscommand.h
Pterosaur: Done

common/redisstring.h
```diff
@@ -0,0 +1,23 @@
...
+class RedisString {
+public:
+
+    RedisString();
+
+    ~RedisString();
+
+    void reset(char *str = nullptr, int len = 0);
+
+    const char *getStr() const;
```
qiluo-msft, line 14: >getStr [](http://example.com/codeflow?start=16&length=6)

If you align the function names with RedisCommand, then RedisCommand could be a subclass of RedisString

common/rediscommand.cpp
```diff
@@ -124,12 +113,12 @@ void RedisCommand::formatDEL(const std::string& key)
 
 const char *RedisCommand::c_str() const
```
qiluo-msft, line 114: >c_str [](http://example.com/codeflow?start=26&length=5)

In your new code, m_str may contains `\0`. Then it is super dangerous to expose `c_str` to outside. Do you want to deprecate this function?

Pterosaur: I don't think so.
1. Because we have already provided the function `len` to get the length, so `\0` will be handled.
2. As the STL string's convention, `const char *` doesn't matter to the `\0'.
3. If we don't expose the `const char *` instead of expose a `std::string`, which is safer but MAY expose one extra deep copy which is unnecessary.

qiluo-msft (approved):

common/stringutility.h
```diff
@@ -166,4 +166,61 @@ static inline std::string binary_to_hex(const void *buffer, size_t length)
     return s;
 }
 
+static inline std::string binary_to_printable(const void *buffer, size_t length)
```
qiluo-msft, line 169: >binary_to_printable [](http://example.com/codeflow?start=26&length=19)

Will you implement a reverse function?

common/rediscommand.h
```diff
@@ -64,12 +66,17 @@ class RedisCommand {
     /* Format DEL key command */
     void formatDEL(const std::string& key);
 
+    int appendTo(redisContext *ctx) const;
+
+    std::string printable_string() const;
```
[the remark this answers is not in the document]
qiluo-msft: Or if you think the semantics are different, you may follow the function naming convention `toPrintableString`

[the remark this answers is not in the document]
Pterosaur: Done

common/rediscommand.h
```diff
@@ -64,12 +66,17 @@ class RedisCommand {
...
+
+private:
```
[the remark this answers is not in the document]
Pterosaur: It's not available, because functions, c_str and len, have been used in other functions of this class.

qiluo-msft (dismissed):

qiluo-msft (approved):

PR #802, opened 2023-07-06
[Ci] Fix the sairedis build target not found issue

xumia:
Fix the sairedis build target not found issue


saiarcot895 (approved):
I also have #703 open to update the build command here. I'm fine with either being merged.

xumia:
@saiarcot895 , #703 is pending Prince's review, I merged it to unblock the build.

PR #803, opened 2023-07-07
Add batch support in ZmqProducerStateTable.

mint570:
The existing batch implementation in ZmqProducerStateTable only calls the individual API repeatedly. This change brings in the real batch implementation:
1. Add a new method in ZmqProducerStateTable, send(), which can batch both SET and DEL operations in a single call.
2. Update the BinarySerializer to encode multiple SET & DEL requests in a single message. The new encoding format is the following:

* db name
* table name
* key1
* number of attrs in key1
* field
* value
* ...
* key2
* number of attrs in key2
* field
* value
* ...

3. Increase ZMQ buffer size from 4MB to 16MB.

### DIFF
--- a/common/binaryserializer.h
+++ b/common/binaryserializer.h
@@ -3,6 +3,8 @@
 
 #include "common/armhelper.h"
 
+#include <string>
+
 using namespace std;
 
 namespace swss {
@@ -12,27 +14,35 @@ public:
     static size_t serializeBuffer(
         const char* buffer,
         const size_t size,
-        const std::string& key,
-        const std::vector<swss::FieldValueTuple>& values,
-        const std::string& command,
         const std::string& dbName,
-        const std::string& tableName)
+        const std::string& tableName,
+        const std::vector<KeyOpFieldsValuesTuple>& kcos)
     {
         auto tmpSerializer = BinarySerializer(buffer, size);
 
+        // Set the first pair as DB name and table name.
         tmpSerializer.setKeyAndValue(
                                     dbName.c_str(), dbName.length(),
                                     tableName.c_str(), tableName.length());
-        tmpSerializer.setKeyAndValue(
-                                    key.c_str(), key.length(),
-                                    command.c_str(), command.length());
-        for (auto& kvp : values)
+        for (auto& kco : kcos)
         {
-            auto& field = fvField(kvp);
-            auto& value = fvValue(kvp);
+            auto& key = kfvKey(kco);
+            auto& fvs = kfvFieldsValues(kco);
+            std::string fvs_len = std::to_string(fvs.size());
+            // For each request, the first pair is the key and the number of attributes,
+            // followed by the attribute pairs.
+            // The operation is not set, when there is no attribute, it is a DEL request.
             tmpSerializer.setKeyAndValue(
-                                        field.c_str(), field.length(),
-                                        value.c_str(), value.length());
+                                        key.c_str(), key.length(),
+                                        fvs_len.c_str(), fvs_len.length());
+            for (auto& fv : fvs)
+            {
+                auto& field = fvField(fv);
+                auto& value = fvValue(fv);
+                tmpSerializer.setKeyAndValue(
+                                            field.c_str(), field.length(),
+                                            value.c_str(), value.length());
+            }
         }
 
         return tmpSerializer.finalize();
@@ -85,6 +95,56 @@ public:
             tmp_buffer += *pvallen;
 
             values.push_back(std::make_pair(pkey, pval));
+        }
+    }
+
+    static void deserializeBuffer(
+        const char* buffer,
+        const size_t size,
+        std::string& dbName,
+        std::string& tableName,
+        std::vector<std::shared_ptr<KeyOpFieldsValuesTuple>>& kcos)
+    {
+        std::vector<FieldValueTuple> values;
+        deserializeBuffer(buffer, size, values);
+        int fvs_size = -1;
+        KeyOpFieldsValuesTuple kco;
+        auto& key = kfvKey(kco);
+        auto& op = kfvOp(kco);
+        auto& fvs = kfvFieldsValues(kco);
+        for (auto& fv : values)
+        {
+            auto& field = fvField(fv);
+            auto& value = fvValue(fv);
+            // The first pair is the DB name and the table name.
+            if (fvs_size < 0)
+            {
+                dbName = field;
+                tableName = value;
+                fvs_size = 0;
+                continue;
+            }
+            // This is the beginning of a request.
+            // The first pair is the key and the number of attributes.
+            // If the attribute count is zero, it is a DEL request.
+            if (fvs_size == 0)
+            {
+                key = field;
+                fvs_size = std::stoi(value);
+                op = (fvs_size == 0) ? DEL_COMMAND : SET_COMMAND;
+                fvs.clear();
+            }
+            // This is an attribut pair.
+            else
+            {
+                fvs.push_back(fv);
+                --fvs_size;
+            }
+            // We got the last attribut pair. This is the end of a request.
+            if (fvs_size == 0)
+            {
+                kcos.push_back(std::make_shared<KeyOpFieldsValuesTuple>(kco));
+            }
         }
     }
 
--- a/common/zmqconsumerstatetable.cpp
+++ b/common/zmqconsumerstatetable.cpp
@@ -39,26 +39,28 @@ ZmqConsumerStateTable::ZmqConsumerStateTable(DBConnector *db, const std::string 
     SWSS_LOG_DEBUG("ZmqConsumerStateTable ctor tableName: %s", tableName.c_str());
 }
 
-void ZmqConsumerStateTable::handleReceivedData(std::shared_ptr<KeyOpFieldsValuesTuple> pkco)
+void ZmqConsumerStateTable::handleReceivedData(const std::vector<std::shared_ptr<KeyOpFieldsValuesTuple>> &kcos)
 {
-    std::shared_ptr<KeyOpFieldsValuesTuple> clone = nullptr;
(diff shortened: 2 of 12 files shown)


mint570:
@qiluo-msft can you take a look as well? Thanks.

common/zmqserver.cpp
```diff
@@ -63,16 +64,10 @@ ZmqMessageHandler* ZmqServer::findMessageHandler(
...
-    BinarySerializer::deserializeBuffer(buffer, size, values);
-
-    // get table name
-    swss::FieldValueTuple fvt = values.at(0);
-    string dbName = fvField(fvt);
-    string tableName = fvValue(fvt);
-    values.erase(values.begin());
+    std::string dbName;
+    std::string tableName;
+    std::vector<std::shared_ptr<KeyOpFieldsValuesTuple>> kcos;
```
liuh-80, line 69: handleReceivedData will add kcos to a queue, so kcos can't create on stack.

common/zmqserver.cpp
```diff
@@ -14,6 +14,7 @@ namespace swss {
 ZmqServer::ZmqServer(const std::string& endpoint)
     : m_endpoint(endpoint)
 {
+    m_buffer.resize(MQ_RESPONSE_MAX_COUNT);
```
liuh-80, line 17: The MQ_RESPONSE_MAX_COUNT is 4MB, so if we enable batch mode, there is a risk that Producer send too much data then Consumer failed to receive.

And it's difficult for client side decide how many entries should send within a batch, because the data are serialized when send. to support batch actually consumer code will complicate than send entry one by one.

I test the ZMQ producer/consumer on device, currently the bottom neck is network badwidth, ZMQ can utilize all bandwidth.

So, the batch support seems not necessary because client side can send request one by one without performance impact.

May I know what's the scenario we need batch send?

common/binaryserializer.h
```diff
@@ -86,6 +92,54 @@ class BinarySerializer {
...
+        std::vector<std::shared_ptr<KeyOpFieldsValuesTuple>>& kcos)
+    {
+        std::vector<FieldValueTuple> values;
+        deserializeBuffer(buffer, size, values);
+        int fvs_size = -1;
+        KeyOpFieldsValuesTuple kco;
+        auto& key = kfvKey(kco);
+        auto& op = kfvOp(kco);
+        auto& fvs = kfvFieldsValues(kco);
+        for (auto& fv : values)
```
liuh-80, line 109: This loop is a state machine, suggest define an enum to control the state change and make the code easy to undertsand.

common/zmqconsumerstatetable.cpp
```diff
@@ -52,29 +52,31 @@ ZmqConsumerStateTable::~ZmqConsumerStateTable()
...
+        }
+        {
+            std::lock_guard<std::mutex> lock(m_receivedQueueMutex);
+            m_receivedOperationQueue.push(kco);
+        }
+        if (m_dbUpdateThread != nullptr)
+        {
+            {
+                std::lock_guard<std::mutex> lock(m_dbUpdateDataQueueMutex);
+                m_dbUpdateDataQueue.push(clone);
```
liuh-80, line 73: After push data to m_dbUpdateDataNotifyCv.notify_all();

Notify client via m_dbUpdateDataNotifyCv.notify_all();

Then the client side will start handle received data immediately.

common/binaryserializer.h
mint570: This is not a state machine. (There is a fvs_size == 0 check at the end of close a request.) The logic just builds on top of the original deserialize code.
I have put the encoding format in the PR description.
Basically the first pair is db name and table name. Then for each request, the first pair is key and attr count, follow by field-value pairs.
Maybe I will put some comments in the codes to explain the encoding format and what the code does?

common/zmqconsumerstatetable.cpp
mint570: We have m_dbUpdateDataNotifyCv.notify_all() at the end.
Are you suggesting to do it in each iteration?

liuh-80: Yes, the reason we switch to ZMQ is we want to improve GNMI to Orch performance.
If we send notification after all data push to queue, the Orch will waiting there during the data push to queue and performance will lower.

common/zmqserver.cpp
liuh-80: Here is an example:
1. If client side want send 100K entry, how many entries client side can put into a batch API call? it's difficult for client side to calculate, because the batch depends on entry size and ZMQ buffer size. and client side don't know ZMQ buffer size, a potential issue is ZMQ buffer size change will break client.

2. If client side send a big request that buffer can't handle, how the ZmqServer side handle it? Client side will never know the request failed because current API not reply status, currently 4MB is safe for single entry. but for batch entries if ZmqServer receive failed how to handle it?

3. another performance impact is: if we send 100K entries in batch, Orch need wait for all data received and deserialized. but if we send one by one, Orch can start process when received first entry, and all other entries will continue send/receive in background thread.

Could you please share the performance improve result, and message side different between send entry one by one and send in batch mode? as my understand the message size actually only have minor different: the different is DB and table are removed from every entry.

common/binaryserializer.h
liuh-80: Please add more comments to make other understand the logic.

Also line 107 get op from KCO, Why line 123 and 136 set op based on fvs_size? 
This have potential risk. in the original code we always use op send from client side.

common/zmqserver.cpp
mint570: Yes, client will need to limit the number of requests if this is a concern. There is no guarantee way of ensuring that as the size of a request can vary (a single request can overflow the buffer if big enough). That's why we increase the buffer to make this unlikely to happen. In our cases, the typical batch size is ~30. We are doing P4, which is RPC based. Each batch corresponds to a RPC call, and 30 is the typical size.

In our case, we have the response path feature, where the application receives notification on the status of their requests (sync mode must be enabled b/w orchagent and syncd), so that the application can reply to the RPC with the correct status reflecting the real hardware state. When buffer overflows, or message lost, or any kind of failure happens, the application will be notified (or timeout), and the gRPC status in the reply will reflect the failure. In this particular case, when buffer overflow happens, the code might throw exception and crash the program. (We have critical state feature to report such fatal failure so that external can take repair action).

About performance, in our case, CPU is not the bottleneck. The bottleneck is the IPC (internal connections) and Redis operations. We saw big improvements on reducing the IPC (using bulk and batch) and reducing Redis operations (DB write in a thread, remove unnecessary DB access etc). We are not comparing this ZMQ work with the original; we are comparing it with our approach. We have the very similar approach, which put the DB write in a thread and get rid of the ProducerStateTable/ConsumerStateTable. The only difference is that, instead of ZMQ, we use the Redis channel as the IPC. We encode the whole batch request into a single Redis channel message. Our experiment found out that: for small batch size (single request), Redis channel is faster (about 2%), and for large batch size (~30), ZMQ is faster (about 5%). This is overall end-to-end comparison in the RPC client end point. The difference will be larger if we just look at the layer b/w application and orchagent. So ZMQ only shines when sending large amount of data. And doing request one-by-one cannot compare to batch (~30 size) at all in our case. It took 6X more time overall to program 1k routes in the single request compare to batch 30. There are just too many IPCs in the single request case. (Many of the orchagents may not support bulk SAI call. We implement bulk support in P4Orch for multiple objects for performance improvement.)

common/zmqconsumerstatetable.cpp
mint570: I moved m_dbUpdateDataNotifyCv.notify_all() back into the loop.

Notice that this only wake up the DB write thread, which we do not care much. The one that wakes up orchagent is m_selectableEvent.notify(). That one is still at the end.

As mentioned in the other comment, bulking makes a big difference in our case. We cannot wake up orchagent prematurely. Orchagent needs to get the full batch and bulk together to syncd.

This may not be the dash use case or other use case. This PR only adds the new batch API. Applications can still use the single request API if that fits more in their scenarios.

common/binaryserializer.h
mint570: Added more comments.

We don't encode the op in the new format. If the attribute size is zero, it is assumed to be a DEL request. This logic is also used in the ConsumerStateTable lua. I moved the size check at the beginning of a request.

common/zmqserver.cpp
liuh-80: For the buffer issue, increase side does not fix the issue, my concern is if any client side send more than buffer side, the server side will break:
1. current received data can't be deserialized because it's incomplete. 
2. Next time received data also can't be deserialized, because the data left in last time will break deserialized.

When server came into this status, I guess the only solution is notify all client stop send data then server side read and drop all data, then notify client can continue sending.

My concern is that if this API merged, and other applications may also use this API, but left this issue here will make other user very difficult to debug and found the root cause.

Please add a receive buffer overflow UT to check if the above issue happens or not.

mint570: There is no return status in the ZmqProducerStateTable API. The client won't directly know the server status. What we have now is the server to throw exception when buffer overthrown happens:

https://github.com/sonic-net/sonic-swss-common/blob/91677ebc86f5ae00ac3c6c3da0e6685cbbd71380/common/zmqserver.cpp#L151

I added the same check in the client side now in this PR. The producer API will throw exception if the caller is sending too much. The caller can catch the exception and reduce the message size.

liuh-80 (approved):

liuh-80:
@mint570 , there still 7 UT failed, it's blocking this PR, please fix and ping me if you need approve again.

mint570:
> @mint570 , there still 7 UT failed, it's blocking this PR, please fix and ping me if you need approve again.

The failed tests are vs pytest. I am not sure if this PR is causing the failure since the vs pytest doesn't use the new ZMQ API.
I re-triggered the test to run and will see.

PR #809, opened 2023-08-01
[redisCommand]: Not store the error return code of redisFormat

Pterosaur:
In the previous implementation, the private member variable, len, will store the error code of redisFormat. This PR is for fixing that if the redisFormat raise an unexpected error code, the `len` will be assigned to zero.

### DIFF
--- a/common/rediscommand.cpp
+++ b/common/rediscommand.cpp
@@ -24,17 +24,19 @@ void RedisCommand::format(const char *fmt, ...)
     {
         redisFreeCommand(temp);
         temp = nullptr;
+        len = 0;
     }
 
     va_list ap;
     va_start(ap, fmt);
-    len = redisvFormatCommand(&temp, fmt, ap);
+    int ret = redisvFormatCommand(&temp, fmt, ap);
     va_end(ap);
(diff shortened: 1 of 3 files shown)


common/rediscommand.cpp
```diff
@@ -24,17 +24,19 @@ void RedisCommand::format(const char *fmt, ...)
     {
         redisFreeCommand(temp);
         temp = nullptr;
+        len = 0;
```
qiluo-msft, line 27: >len [](http://example.com/codeflow?start=8&length=3)

Better move this line out of if-block.

common/rediscommand.cpp
```diff
@@ -43,12 +45,14 @@ void RedisCommand::formatArgv(int argc, const char **argv, const size_t *argvlen
...
```
qiluo-msft, line 48: >len [](http://example.com/codeflow?start=8&length=3)

The same.

qiluo-msft (approved):

Pterosaur:
@qiluo-msft I don't have permission to merge it. Please feel free to merge this PR if it looks good to you.

PR #811, opened 2023-08-08
Fix the issue of ignoring callback calls for removed keys.

oleksandrivantsiv:
**What I did**
Fix the issue of ignoring callback calls for removed keys.

**Why I did it**
ConfigDBConnector.listen method has a caching mechanism (added in https://github.com/sonic-net/sonic-swss-common/pull/587 PR) that preloads the DB state before starting. When the notification about the changed key is received the listen method gets key data from the DB (in all cases when the key was added, updated, or removed) and compares the data with the cache. It fires the callback only if data differ from the cache. Otherwise, the callback is ignored. 

If the `client.hgetall(key)` is called for the removed key it returns an empty dictionary (`{}`). This can be confused with the data of the key with no attributes. For example: `{"TABLE": {"KEY": {}}}`. 

So if preloaded data contains a key with no attributes and the listener method receives a notification about the removal of such key the callback is not called. The listener will simply remove the key from the cache without calling the callback. This leads to the situation when the removal of the key is not handled.

The solution is to get data for the added or updated keys, and for the removed keys use `None` instead. This will ensure that the comparison with the cache will work as expected.

**How I verified it**
Compile the package and run the unit test. Unit tests are extended to cover the expected behavior.

**Details if related**


oleksandrivantsiv:
This PR is a follow-up of https://github.com/sonic-net/sonic-swss-common/pull/789

Yakiv-Huryk (approved):

common/configdb.h
```diff
@@ -105,8 +105,11 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
                     try:
                         (table, row) = key.split(self.TABLE_NAME_SEPARATOR, 1)
                         if table in self.handlers:
-                            client = self.get_redis_client(self.db_name)
-                            data = self.raw_to_typed(client.hgetall(key))
+                            if item['data'] == 'del':
```
qiluo-msft, line 108: >del [](http://example.com/codeflow?start=48&length=3)

Could you add some test cases for hdel?
1. hdel one field, but there are still remaining fields
2. hdel the last field, so the whole key disappears.

dgsudharsan (approved):

common/configdb.h
oleksandrivantsiv: @qiluo-msft done. Please review

oleksandrivantsiv:
@qiluo-msft kindly reminder. Please review the changes.

qiluo-msft (approved):

PR #820, opened 2023-09-15
Export RedisError to SWIG wrapper.

liuh-80:
#### Why I did it
RedisError class and other class defined in redisreply.h can't be used in python.
This is bcause redisreply.h nor gnerate SWIG wrapper.

#### How I did it
Add redisreply.h to swsscommon.i for generate SWIG wrapper.

#### How to verify it
Manually check the classes in redisreply.h exist in SWIG wrapper.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Add redisreply.h to swsscommon.i for generate SWIG wrapper.


ganglyu (dismissed):

common/consumertable.cpp
```diff
@@ -90,7 +90,7 @@ void ConsumerTable::pops(deque<KeyOpFieldsValuesTuple> &vkco, const string &pref
             if (i+1 >= ctx->elements)
             {
                 SWSS_LOG_ERROR("invalid number of elements in returned table: %zu >= %zu", i+1, ctx->elements);
-                throw runtime_error("invalid number of elements in returned table");
+                throw RedisError("invalid number of elements in returned table");
```
qiluo-msft, line 93: >RedisError [](http://example.com/codeflow?start=22&length=10)

changing very old exception type will be dangerous, since the library is used by too many applications, and they may catch by some exception class in a chain.

liuh-80: Fixed, revert to first commit to only export RedisError class.

qiluo-msft (approved):

PR #821, opened 2023-09-27
Support async DB update for both ZMQ producer&consumer table.

liuh-80:
#### Why I did it
Some scenarios require sonic-gnmi side update data to APPL_DB.

#### How I did it
Move async DB update code to AsyncDBUpdater.
Reuse AsyncDBUpdater in both ZmqProducerStateTable and ZmqConsumerStateTable

#### How to verify it
Add new UT.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Support async DB update for both ZMQ producer&consumer table.

### DIFF
--- a/common/zmqconsumerstatetable.cpp
+++ b/common/zmqconsumerstatetable.cpp
@@ -90,4 +90,14 @@ void ZmqConsumerStateTable::pops(std::deque<KeyOpFieldsValuesTuple> &vkco, const
     }
 }
 
+size_t ZmqConsumerStateTable::dbUpdaterQueueSize()
+{
+    if (m_asyncDBUpdater == nullptr)
+    {
+        return 0;
+    }
+
+    return m_asyncDBUpdater->queueSize();
+}
+
 }
--- a/common/zmqconsumerstatetable.cpp
+++ b/common/zmqconsumerstatetable.cpp
@@ -26,13 +26,12 @@ ZmqConsumerStateTable::ZmqConsumerStateTable(DBConnector *db, const std::string 
     if (dbPersistence)
     {
         SWSS_LOG_DEBUG("Database persistence enabled, tableName: %s", tableName.c_str());
-        m_runThread = true;
-        m_dbUpdateThread = std::make_shared<std::thread>(&ZmqConsumerStateTable::dbUpdateThread, this);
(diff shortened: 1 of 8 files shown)


tests/zmq_state_ut.cpp
```diff
@@ -109,11 +109,21 @@ static void producerWorker(string tableName, string endpoint)
...
 
+    // wait all data been received by consumer
     while (!allDataReceived)
     {
         sleep(1);
     }
 
+    if (dbPersistence)
+    {
+        // wait all persist data write to redis
```
qiluo-msft, line 120: >all persist data write to redis [](http://example.com/codeflow?start=16&length=31)

Do you test that "persist data write to redis" is actually as expected? That will be a good test to make sure this PR does not regress.

common/zmqconsumerstatetable.cpp
```diff
@@ -174,4 +90,14 @@ void ZmqConsumerStateTable::pops(std::deque<KeyOpFieldsValuesTuple> &vkco, const
     }
 }
 
+size_t ZmqConsumerStateTable::dbUpdaterQueueSize()
+{
+    if (m_asyncDBUpdater == nullptr)
+    {
+        return 0;
```
qiluo-msft, line 97: >return [](http://example.com/codeflow?start=8&length=6)

Should we throw to be safe?

liuh-80: Fixed, throw operation_not_supported exception when DB persistence disabled.

common/zmqconsumerstatetable.h
```diff
@@ -3,10 +3,11 @@
...
-#include "dbconnector.h"
-#include "table.h"
+#include "asyncdbupdater.h"
 #include "consumertablebase.h"
+#include "dbconnector.h"
 #include "selectableevent.h"
+#include "table.h"
 #include "zmqserver.h"
 
 #define MQ_RESPONSE_MAX_COUNT (4*1024*1024)
```
Pterosaur, line 13: This MACRO has been defined in asyncdbupdater.h, why redefine it?

liuh-80: Fixed, remove these dupe macro

qiluo-msft (approved):

PR #825, opened 2023-10-12
Update sonic-swss-common for compilation on Bookworm

saiarcot895:
This PR includes the following two changes:

- Use `json.hpp` form the `nlohmann-json-dev` package instead of having a local copy. This is because this version as-is now has compilation errors on Bookworm, and this file is coming from an external source.
- Fix compilation errors with GCC 12, related to reference-vs-copy errors in for-loops and casting issues.


common/events.cpp
```diff
@@ -400,14 +400,14 @@ EventSubscriber::prune_track()
     map<time_t, vector<runtime_id_t> > lst;
 
     /* Sort entries by last touched time */
-    for(const auto e: m_track) {
+    for(const auto &e: m_track) {
```
qiluo-msft, line 403: >const [](http://example.com/codeflow?start=8&length=5)

@zbud-msft Could you help review events related code?

qiluo-msft (approved):

common/events.cpp
saiarcot895: @zbud-msft Reminder for this

zbud-msft: LGTM

zbud-msft (approved):

PR #834, opened 2023-11-16
Inspur keep del when del set sequence exist

inspurSDN:
Why I did it:
     when set after delete in a short time, consumer state may lost the del event.
     then what the multimap bring us for it may lost still.
 What I do:
     add the delete event even set after
     and modify the ut test for these changes.
     this can be verified by these testcase.


mint570:
This is the same as https://github.com/sonic-net/sonic-swss-common/pull/833

mint570:
Closing this PR now since it is the same as https://github.com/sonic-net/sonic-swss-common/pull/833.
Please review and comment on https://github.com/sonic-net/sonic-swss-common/pull/833.

PR #837, opened 2023-11-21
[sonic-db-cli]: SONiC DB CLI for DPU objects

Pterosaur:
Wait for PR: https://github.com/sonic-net/sonic-swss/pull/2966
Read DPU objects from Redis via SONiC DB CLI


tests/cli_ut.cpp
```diff
@@ -539,4 +541,34 @@ TEST(sonic_db_cli, test_cli_not_throw_exception)
...
+    args[0] = "sonic-db-cli";
+    args[1] = "TEST_DB";
+
+    // clear database
+    args[2] = "FLUSHDB";
+    auto output = runCli(3, args);
+    EXPECT_EQ("True\n", output);
+
+    // restore binary pb message to TEST_DB
+    EXPECT_EQ(system("cat ./tests/cli_test_data/pb_appliance.bin "
```
qiluo-msft, line 558: >pb_appliance.bin [](http://example.com/codeflow?start=48&length=16)

Do you want to check-in this binary file with this PR? Better to prevent binary file in codebase, instead you can generate it.

configure.ac
```diff
@@ -93,6 +93,15 @@ CFLAGS_COMMON+=" -fstack-protector-strong"
 
 AC_SUBST(CFLAGS_COMMON)
 
+AC_CHECK_LIB([dashapi], [PbBinaryToJsonString], [
+    AC_DEFINE(DASH_API_INSTALLED, 1, [Define if DASH API is installed])
+    LIBS="$LIBS -ldashapi"
```
qiluo-msft, line 98: >ldashapi [](http://example.com/codeflow?start=17&length=8)

swss-common is widely used, it is awkward to ask others to install dashapi.
I see below options:
1. remove the dependency on dashapi, and make this PR a generic feature
2. refactor sonic-db-cli to support plugin, and develop a new plugin with dashapi dependency 
3. develop another CLI like sonic-db-dash-cli (just random name).

Pterosaur:
Refactor PR: https://github.com/sonic-net/sonic-dash-api/pull/12

PR #838, opened 2023-11-30
Add support of 'with' statement to ConfigDBConnector

liuh-80:
#### Why I did it
Add support of 'with' statement to ConfigDBConnector

#### How I did it
Add __enter__ and __exit__ method to ConfigDBConnector

##### Work item tracking
- Microsoft ADO: 24222755

#### How to verify it
Add new UT to check ConfigDBConnector support 'with' statement

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Add support of 'with' statement to ConfigDBConnector


common/configdb.h
```diff
@@ -59,6 +59,12 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
             self.handlers = {}
             self.fire_init_data = {}
 
+        def __enter__(self):
+            return self
+
+        def __exit__(self, exc_type, exc_value, exc_tb):
```
qiluo-msft, line 65: >__exit__ [](http://example.com/codeflow?start=12&length=8)

We should add cleanup code here.

liuh-80: Fixed, add new method to close all connection in current DBConnector and call it in __exit__

common/dbinterface.cpp
```diff
@@ -32,6 +32,11 @@ void DBInterface::close(const std::string& dbName)
     m_redisClient.erase(dbName);
```
qiluo-msft, line 32: >erase [](http://example.com/codeflow?start=18&length=5)

This `erase` does not release the resource immediately. We need to recusively close these objects, which may need new implementation in other classes.

common/dbinterface.cpp
```diff
@@ -32,6 +32,11 @@ void DBInterface::close(const std::string& dbName)
...
 }
 
+void DBInterface::close_all()
```
qiluo-msft, line 35: >close_all [](http://example.com/codeflow?start=18&length=9)

Is it possible to overload `close` function? I am not sure how SWIG handle the overload or reusing the same function with default parameter None easily.

common/dbinterface.cpp
liuh-80: The 'with' state will not call this method, it will call the map::clear(), and because m_redisClient not store pointer or reference, the dtor of RedisContext will be invoke by clear() and release redis connection:

RedisContext::~RedisContext()
{
    if(m_conn)
        redisFree(m_conn);
}

common/dbinterface.cpp
liuh-80: Fixed

common/dbinterface.cpp
qiluo-msft: Agree. The original comment is invalid.

qiluo-msft (approved):

PR #843, opened 2023-12-14
Add SonicDBConfig::reset method

ganglyu:
#### Why I did it
sonic-gnmi will use swsscommon API to read database configuration, and sonic-gnmi unit test needs to reload database configuration in the same process.

#### How I did it
Add SonicDBConfig::reset method, and sonic-gnmi unit test will use this method to clear previous database configuration.

#### How to verify it

- [ ] 202012
- [ ] 202205
- [ ] 202305

Add SonicDBConfig::reset method

#### A picture of a cute animal (not mandatory but encouraged)

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -185,6 +185,18 @@ void SonicDBConfig::initialize(const string &file)
     m_init = true;
 }
 
+// This API is used to uninitialize the SonicDBConfig class.
+// And then user can call initialize with different config file.
+void SonicDBConfig::uninitialize()
+{
+    std::lock_guard<std::recursive_mutex> guard(m_db_info_mutex);
+    m_init = false;
+    m_global_init = false;
+    m_inst_info.clear();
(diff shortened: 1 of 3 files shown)


common/dbconnector.cpp
```diff
@@ -185,6 +185,18 @@ void SonicDBConfig::initialize(const string &file)
     m_init = true;
 }
 
+// This API is used to uninitialize the SonicDBConfig class.
+// And then user can call initialize with different config file.
+void SonicDBConfig::uninitialize()
```
qiluo-msft, line 190: >uninitialize [](http://example.com/codeflow?start=20&length=12)

Does it make more sense to improve existing `initialize`, so user could call it more than once?

ganglyu: Got it, will update

qiluo-msft: Offline discussed, the first iteration should be good.

qiluo-msft (approved):

PR #859, opened 2024-02-22
Fix swig template memory leak

praveenraja1:
Problem:
=======
Device gets stuck if left switched on for prolonged period of time for more than 30 hrs.

Observation:
=========

Pmon docker seems to be leaking memory over time, leading to system wide out-of-memory condition.
Python process in pmon such as psud, thermalctld were all leaking memory over time.
The issue was traced back to table.set() operations that were done in these python process.
We have swig template files which are used to generate the code mappings for python to cpp conversions.

In these constructor functions, we have allocated new references using PySequence_GetItem(), which these references were not marked with Py_DECREF, when done using them. This was causing stale objects to present which was leaking memory over time.
/usr/lib/python3/dist-packages/swsscommon/swsscommon.py:184: size=2851 B, count=49, average=58 B
/usr/lib/python3.9/tracemalloc.py:505: size=1456 B, count=24, average=61 B
/usr/lib/python3.9/tracemalloc.py:67: size=1344 B, count=21, average=64 B
/usr/lib/python3.9/tracemalloc.py:558: size=1224 B, count=24, average=51 B
/usr/lib/python3.9/tracemalloc.py:498: size=1104 B, count=23, average=48 B
/usr/lib/python3.9/tracemalloc.py:193: size=768 B, count=16, average=48 B
/usr/lib/python3.9/tracemalloc.py:533: size=576 B, count=1, average=576 B
/usr/lib/python3.9/threading.py:574: size=536 B, count=1, average=536 B

Fix:
===
Used Py_DECREF in constructor functions.

Unit test:
=======

Tested using a custom py script which does table.set in while loop (and checking tracemalloc) to ensure memory leak doesnt happen.

Allowed the system to be up for sometime to ensure memory usage of pmon docker doesnt increase like earlier.

Check show platform fan/psu to ensure that pmon functionality isnt affected due to this change.


pyext/swsscommon.i
```diff
@@ -164,6 +164,7 @@
         temp.push_back(std::pair< std::string,std::string >());
         PyObject *item = PySequence_GetItem($input, i);
         if (!PyTuple_Check(item) || PyTuple_Size(item) != 2) {
+            Py_DECREF(item);
```
qiluo-msft, line 167: >Py_DECREF [](http://example.com/codeflow?start=12&length=9)

Thanks for the bugfix! the code LGTM. Is it possible to add unit test?

Pterosaur: I feel we already have a case: https://github.com/sonic-net/sonic-swss-common/blob/00445404fff6d3673464e2a5defca1dc97ab3a43/tests/test_redis_ut.py#L165 to test this scenario, but I don't know why it was not exposed by this case. I'm investigating it.

Pterosaur:
Issue has been fixed, thanks

PR #862, opened 2024-03-12
Add TupleHelper and SelectHelper to improve SWIG Go wrapper.

liuh-80:
#### Why I did it
SWIG not support std::tuple, also not support C++ inheritance well, this make go wrapper method of select and pops method does not work.

#### How I did it
Add TupleHelper and SelectHelper to improve SWIG Go wrapper. 

##### Work item tracking
- Microsoft ADO: 26731929

#### How to verify it
Pass all test cases.
Add new test case.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Add TupleHelper and SelectHelper to improve SWIG Go wrapper.

### DIFF
--- a/common/configdb.h
+++ b/common/configdb.h
@@ -38,26 +38,10 @@ protected:
 };
 
 #if defined(SWIG) && defined(SWIGGO)
-%go_import("fmt")
 %insert(go_wrapper) %{
 
 type ConfigDBConnector struct {
     ConfigDBConnector_Native
-}
-
-func (configDbConnector ConfigDBConnector) Hget(table string, key string, field string) (string, error) {
-    var fieldValuePairs = configDbConnector.Get_entry(table, key)
-    if fieldValuePairs.Has_key(field) {
-        return fieldValuePairs.Get(field), nil
-    }
-
-    return "", fmt.Errorf("Can't read %s:%s from %s table", key, field, table)
-}
-
-func (configDbConnector ConfigDBConnector) Hset(table string, key string, field string, value string) {
-    var fieldValuePairs = configDbConnector.Get_entry(table, key)
-    fieldValuePairs.Set(field, value)
-    configDbConnector.Set_entry(table, key, fieldValuePairs)
 }
 
 func NewConfigDBConnector(a ...interface{}) *ConfigDBConnector {
--- a/common/configdb.h
+++ b/common/configdb.h
@@ -37,6 +37,38 @@ protected:
     std::string m_db_name;
 };
 
+#if defined(SWIG) && defined(SWIGGO)
+%go_import("fmt")
+%insert(go_wrapper) %{
+
+type ConfigDBConnector struct {
+    ConfigDBConnector_Native
+}
+
+func (configDbConnector ConfigDBConnector) Hget(table string, key string, field string) (string, error) {
+    var fieldValuePairs = configDbConnector.Get_entry(table, key)
+    if fieldValuePairs.Has_key(field) {
+        return fieldValuePairs.Get(field), nil
+    }
+
+    return "", fmt.Errorf("Can't read %s:%s from %s table", key, field, table)
+}
+
+func (configDbConnector ConfigDBConnector) Hset(table string, key string, field string, value string) {
+    var fieldValuePairs = configDbConnector.Get_entry(table, key)
+    fieldValuePairs.Set(field, value)
+    configDbConnector.Set_entry(table, key, fieldValuePairs)
+}
+
+func NewConfigDBConnector(a ...interface{}) *ConfigDBConnector {
+    return &ConfigDBConnector{
+        NewConfigDBConnector_Native(a...),
+    }
+}
+%}
+#endif
+
+
 #if defined(SWIG) && defined(SWIGPYTHON)
 %pythoncode %{
     ## Note: diamond inheritance, reusing functions in both classes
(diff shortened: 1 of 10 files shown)


common/configdb.h
```diff
@@ -37,6 +37,38 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
...
 
+#if defined(SWIG) && defined(SWIGGO)
+%go_import("fmt")
+%insert(go_wrapper) %{
+
+type ConfigDBConnector struct {
+    ConfigDBConnector_Native
+}
+
+func (configDbConnector ConfigDBConnector) Hget(table string, key string, field string) (string, error) {
```
qiluo-msft, line 48: >Hget [](http://example.com/codeflow?start=43&length=4)

We intend to aligned the library APIs in programming languages. We do not want to add extra functions in golang. If really needed, we need to add into all languages.

common/configdb.h
```diff
@@ -37,6 +37,38 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
...
+    var fieldValuePairs = configDbConnector.Get_entry(table, key)
+    if fieldValuePairs.Has_key(field) {
+        return fieldValuePairs.Get(field), nil
+    }
+
+    return "", fmt.Errorf("Can't read %s:%s from %s table", key, field, table)
+}
+
+func (configDbConnector ConfigDBConnector) Hset(table string, key string, field string, value string) {
```
qiluo-msft, line 57: >Hset [](http://example.com/codeflow?start=43&length=4)

The same. If you remove 2 functions, ConfigDBConnector is exact the same as ConfigDBConnector_Native?

common/configdb.h
liuh-80: Fixed, this method removed.

common/configdb.h
liuh-80: Removed.

common/tuplehelper.h
```diff
@@ -0,0 +1,28 @@
+ #ifndef __TUPLE_HELPER__
```
qiluo-msft, line 1: >  [](http://example.com/codeflow?start=0&length=1)

extra blank

common/tuplehelper.cpp
```diff
@@ -0,0 +1,32 @@
...
+{
+    return kfvKey(tuple);
+}
+
+std::string TupleHelper::getOp(const KeyOpFieldsValuesTuple& tuple)
+{
+    return kfvOp(tuple);
+}
+
+std::vector<FieldValueTuple> TupleHelper::getFieldsValues(const KeyOpFieldsValuesTuple& tuple)
```
qiluo-msft, line 27: >vector [](http://example.com/codeflow?start=5&length=6)

Constructing this return object is heavy.

common/tuplehelper.cpp
```diff
@@ -0,0 +1,32 @@
...
+{
+    return fvField(tuple);
+}
+
+std::string TupleHelper::getValue(const FieldValueTuple& tuple)
+{
+    return fvValue(tuple);
+}
+
+std::string TupleHelper::getKey(const KeyOpFieldsValuesTuple& tuple)
```
qiluo-msft, line 17: >getKey [](http://example.com/codeflow?start=25&length=6)

I guess the original kfvKey macro could not easily transformed by SWIG. If yes, recommend to change kfvKey to an inline function, instead of defining new class/functions.

common/select.h
```diff
@@ -73,6 +73,27 @@ class Select
     std::set<Selectable *, Select::cmp> m_ready;
 };
 
+class SelectHelper
```
qiluo-msft, line 76: >SelectHelper [](http://example.com/codeflow?start=6&length=12)

Why we need a new class? Can SWIG just transform old class?

common/configdb.h
```diff
@@ -37,6 +37,22 @@ class ConfigDBConnector_Native : public SonicV2Connector_Native
...
 };
 
+#if defined(SWIG) && defined(SWIGGO)
+%insert(go_wrapper) %{
+
+type ConfigDBConnector struct {
+    ConfigDBConnector_Native
+}
+
+func NewConfigDBConnector(a ...interface{}) *ConfigDBConnector {
```
qiluo-msft, line 47: >NewConfigDBConnector [](http://example.com/codeflow?start=5&length=20)

SWIG will automatically create New* functions for class for golang. Why you need explicitly define this function?

liuh-80: SWIG will only create New* functions for C++ class, but in this case ConfigDBConnector is a go struct, so we need implement NewConfigDBConnector method manually.

common/select.h
liuh-80: We need this new class because SWIG can't handle pointer to pointer parameter:

    int Select::select(**Selectable **c,** int timeout = -1, bool interrupt_on_signal = false);

We need this helper class handle the pointer to pointer parameter in C++ code.

common/tuplehelper.cpp
liuh-80: Will check and update later, currently those macros are used by many other projects, change to inline have potential risk to break their code.

common/tuplehelper.h
liuh-80: Fixed

common/tuplehelper.cpp
liuh-80: Fixed, return reference

liuh-80:
Close because Go wrapper still need improve. the change for GNMI support will create new PR.

PR #869, opened 2024-04-01
Remove the error log form SonicDBConfig::initializeGlobalConfig

wumiaont:
**Issue**: If run "show interfaces status" we often see the error in syslog "SonicDBConfig Global config is already initialized". 

**Analysis**: During investigation it is found the swsscommon.load_sonic_global_db_config() has been triggered multiple times. Espically in a multi Asic scenario.

CLI triggers "intfutil -c status" with a separate process. Within the process it will trigger IntfStatus __init__ --> MultiAsic() __init__-->load_db_config(). This time initializeGlobalConfig() will run to the end as m_global_init flag is false.

CLI will also run get_port_config() --> db_connect_configdb() --> load_db_config() (for different namepaces. In our case is asic0 and asic1).  This time m_global_init flag is true which causes error log and return.  

**Conclusion**: After researching we found load_db_config() could be called by different Classes. The code has already added protection to avoid to initialize global config multiple time. The error log is not necessary to be there. 

**Solution**: Removed the error log from initializeGlobalConfig() when found global config is already initialized.

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -84,7 +84,6 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
 
     if (m_global_init)
     {
-        SWSS_LOG_ERROR("SonicDBConfig Global config is already initialized");
         return;
     }
 


wumiaont:
@deepak-singhal0408 @mlok-nokia Please help to review.

deepak-singhal0408 (approved):
LGTM..

deepak-singhal0408:
MSFT ADO: 25516271

deepak-singhal0408:
Hi @qiluo-msft, could you please help review this PR? Thanks!

common/dbconnector.cpp
```diff
@@ -84,7 +84,6 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
 
     if (m_global_init)
     {
-        SWSS_LOG_ERROR("SonicDBConfig Global config is already initialized");
```
qiluo-msft, line 87: >SWSS_LOG_ERROR [](http://example.com/codeflow?start=8&length=14)

The error message is designed intentionally for application to realize there is a problem that calling initialization more than once. we need to fix application.

qiluo-msft (changes requested):
Block this PR, adding comment.

deepak-singhal0408:
Hi @wumiaont , please have a look at this PR.. https://github.com/sonic-net/sonic-buildimage/pull/10960..
Please take this change and give it a try?

PR #870, opened 2024-04-02
Add Bookworm build to PR checks

saiarcot895:
Add a step to verify the build for Bookworm for amd64, armhf, and arm64.


saiarcot895:
Depends on sonic-net/sonic-buildimage#18544.

PR #873, opened 2024-05-01
sonic-db-cli: Don't use unix socket for redis_chassis.server

arista-nwolfe:
Described in https://github.com/sonic-net/sonic-buildimage/issues/18733 we see that swss.sh will try to access `CHASSIS_APP_DB` via `sonic-db-cli`.
If we're on a multi-asic system it will include the `-n <asic>` argument.
Having the `-n <asic>` argument causes `sonic-db-cli` to use unix sockets due to https://github.com/sonic-net/sonic-swss-common/pull/797
If we use a unix socket to try to reach the supervisor from the LC the command will fail with:
```
admin@cmp217-5:/var/log$ sudo sonic-db-cli -n asic0 CHASSIS_APP_DB keys "*"
Invalid database name input : 'CHASSIS_APP_DB'
Unable to connect to redis (unix-socket): Cannot assign requested address
```

https://github.com/sonic-net/sonic-swss-common/pull/866 fixed a similar issue with database.sh by forcing a tcp connection when accessing `CHASSIS_APP_DB` (`redis_chassis.server`) specifically for the `PING` `SAVE` and `FLUSHALL` operations.
This change provides that same tcp connection override for the other operations.

### DIFF
--- a/sonic-db-cli/sonic-db-cli.cpp
+++ b/sonic-db-cli/sonic-db-cli.cpp
@@ -139,14 +139,14 @@ int executeCommands(
     try
     {
         int db_id =  SonicDBConfig::getDbId(db_name, netns);
-        if (useUnixSocket)
+        auto host = SonicDBConfig::getDbHostname(db_name, netns);
+        if (useUnixSocket && host != "redis_chassis.server")
         {
             auto db_socket = SonicDBConfig::getDbSock(db_name, netns);
             client = make_shared<DBConnector>(db_id, db_socket, 0);
         }
         else
         {
-            auto host = SonicDBConfig::getDbHostname(db_name, netns);
             auto port = SonicDBConfig::getDbPort(db_name, netns);
             client = make_shared<DBConnector>(db_id, host, port, 0);
         }


arista-nwolfe:
Tagging @mlok-nokia to confirm if this is a viable solution to the `CHASSIS_APP_DB` unix socket issue on multi-asic LCs we discussed in the meeting today.

mlok-nokia:
> Tagging @mlok-nokia to confirm if this is a viable solution to the `CHASSIS_APP_DB` unix socket issue on multi-asic LCs we discussed in the meeting today.

This fix looks good to me.  It uses the same mechanism as how it has been handled in the local host. I think we can use this fix to make sonic-db-cli command behaves the same as 202205 branch.

mlok-nokia (approved):
It looks good to me.

rlhui:
@qiluo-msft  ping again, thx

arlakshm (approved):

qiluo-msft:
sonic-db-cli is behaving similar to redis-cli. If user provide `options.m_unixsocket`, we should stick with it. Even in existing code, I do not understand why options.m_unixsocket could be changed.

arlakshm:
> sonic-db-cli is behaving similar to redis-cli. If user provide `options.m_unixsocket`, we should stick with it. Even in existing code, I do not understand why options.m_unixsocket could be changed.

@qiluo-msft in the PR, https://github.com/sonic-net/sonic-swss-common/pull/797, change was made to always use unix socket when passing the `namespace` option to the `sonic-db-cli` command.
The user option is overwritten to use unixsocket in this case

This PR is fixing communication to chassis redis instance from namespace, because we cannot use unixsocket in this case.

lguohan:
@qiluo-msft , can you check this one?

judyjoseph:
@arista-nwolfe could you rebase this branch

liuh-80 (approved):

qiluo-msft:
@arista-nwolfe Could you check why https://github.com/sonic-net/sonic-swss-common/pull/883 is failing

arista-nwolfe:
> @arista-nwolfe Could you check why #883 is failing

I was told that there is a dependency issue in `sonic-swss-common` on 202405 causing all builds to fail.
Tagging @arlakshm here.

I'm not sure why the 202405 cast was closed.

PR #893, opened 2024-07-16
[common] add PerformanceTimer

a114j0y:
What I did

Extend the `PerformanceIntervalTimer` utility in `sonic-sairedis` repo  to measure API performance.

To optimize performance of a periodically invoked function,  we not only pay attention to how long it takes to finish execution (busy time), but also care about how long it waits to get executed (idle time). The total time determines the actual throughput in the real world system.

Hence we want to see data in 3 ways:
1. `Idle` time between two calls (a gap between this start and the last end, during which the thread just waits)
2. `Busy` time of each API call (interval between this start and this end, it's the true execution time)
3. `Total` time elapsed (idle + busy)

` | <- Initial Gap -> | <- - 1st **Execution** - -> | <- 2nd Gap -> | <- - 2nd **Execution** - -> |`

`  | <---------------- 1st **Total** --------------> |  <------------ 2nd **Total**--------------> | `

Other features
- calculate the number of tasks finished per call (Routes Per Second in the context of routes loading)
- batch the logging output
- adjust the `SWSSLOGLEVEL` using a file indicator, default level is "INFO", hence we could enable perf logs in a dynamic way. 

Why I did it
- original utility has limited functionality
    - we need gap data to obtain more insight on the mutual influence of upstream and downstream modules
- original utility is in libsairedis, to use it in sonic-swss, there would be much Makefile.am change, not necessary. 

- a utility tool for swss should be included in swss-common

- when enabled on-demand, it could help measure the API performance under scaled traffic 
    - one use case: help measuring perf in pr [optimized bgp loading speed](https://github.com/sonic-net/SONiC/pull/1521)

### DIFF
--- a/common/performancetimer.cpp
+++ b/common/performancetimer.cpp
@@ -1,0 +1,128 @@
+#include "performancetimer.h"
+
+#include "logger.h"
+#include <nlohmann/json.hpp>
+#include <fstream>
+
+using namespace swss;
+
+bool PerformanceTimer::m_enable = true;
+#define LIMIT 5
+#define INDICATOR "/var/log/syslog_notice_flag"
+
+PerformanceTimer::PerformanceTimer(
+        _In_ std::string funcName,
+        _In_ uint64_t threshold,
+        _In_ bool verbose):
+        m_name(funcName),
+        m_threshold(threshold),
+        m_verbose(verbose)
+{
+    reset();
+    m_stop = std::chrono::steady_clock::now();
+}
+
+void PerformanceTimer::reset()
+{
+    SWSS_LOG_ENTER();
+
+    m_tasks = 0;
+    m_calls = 0;
+    m_busy = 0;
+    m_idle = 0;
+
+    m_intervals.clear();
+    m_gaps.clear();
+    m_incs.clear();
+}
+
+void PerformanceTimer::start()
+{
+    SWSS_LOG_ENTER();
+
+    m_start = std::chrono::steady_clock::now();
+    // measures the gap between this start() and the last stop()
+    m_gaps.push_back(std::chrono::duration_cast<std::chrono::milliseconds>(m_start-m_stop).count());
+}
+
+void PerformanceTimer::stop()
+{
+    SWSS_LOG_ENTER();
+    m_stop = std::chrono::steady_clock::now();
+}
+
+void PerformanceTimer::inc(uint64_t count)
+{
+    SWSS_LOG_ENTER();
+
+    m_calls += 1;
+
+    m_tasks += count;
+
+    m_idle += m_gaps.back();
+
+    uint64_t interval = std::chrono::duration_cast<std::chrono::nanoseconds>(m_stop - m_start).count();
+
+    m_busy += interval;
+
+    if (count == 0) {
+        m_gaps.pop_back();
+        m_calls -= 1;
+        return;
+    }
+
+    if (m_incs.size() <= LIMIT) {
+        m_incs.push_back(count);
+        m_intervals.push_back(interval/1000000);
+    } else {
+        m_gaps.pop_back();
+    }
+
+    if (m_tasks >= m_threshold)
+    {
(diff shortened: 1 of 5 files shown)


common/performancetimer.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+#include "logger.h"
+#include <nlohmann/json.hpp>
+#include <fstream>
+
+using namespace swss;
+
+bool PerformanceTimer::m_enable = true;
+#define LIMIT 5
+PerformanceTimer::PerformanceTimer(
```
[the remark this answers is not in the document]
a114j0y: It's a utility class and there is already a similar tool named `PerformanceIntervalTimer` in sonic-sairedis, which is active in use , however it's in libsairedis and has limited functionality.
This feature `PerformanceTimer` is an extension based on that. We include it in `swsscommon` library so that it can be utilized in `sonic-swss`.
This feature could be on-demand enabled. It would not hurt performance, since we have achieved a nice number with this feature turned on.

common/performancetimer.cpp
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    if (m_tasks >= m_threshold)
+    {
+        uint64_t mseconds = m_busy/1000000;
+
+        if (m_enable && mseconds > 0)
+        {
+            output = getTimerState();
+            std::ifstream indicator("/var/log/syslog_notice_flag");
```
qiluo-msft, line 88: >indicator [](http://example.com/codeflow?start=26&length=9)

I do not understand why it is related to syslog, why using this file? It is not covered in HLD.

a114j0y: The timer uses `SWSS_LOG_INFO` to print logs, but the default `loglevel` of `orchagent` is `NOTICE`, so the logs would not appear in syslog unless we manually upgrade the `loglevel` of `orchagent` to be `INFO`. 

In spytest, we need those logs to appear in syslog for us to analyze module performance, but it's not convenient to modify orchagent loglevel during the test, so we utilize this file indicator. 

In a spytest, we could create such a file `/var/log/syslog_notice_flag`, when timer detects the existence of this file, it prints in `NOTICE` level. We delete this file at the end of a test, to make the timer print in `INFO` level again.

Instead of modifying orchagent loglevel, we use the file to adjust whether the timer prints logs in syslog.

eddieruan-alibaba:
@qiluo-msft @liuh-80 @lguohan  can you help to review and approve it?

liuh-80 (approved):

liuh-80:
Hi @a114j0y , the performance timer test is unstable, almost every PR need trigger validation 5-10 times to pass this test. can you improve it?

Expected equality of these values:
  output
    Which is: "{\"API\":\"basic_set_name\",\"RPS[k]\":5.0,\"Tasks\":3000,\"Total[ms]\":603,\"busy[ms]\":400,\"idle[ms]\":203,\"m_gaps\":[3,200],\"m_incs\":[1000,2000],\"m_intervals\":[100,300]}"
  expected
    Which is: "{\"API\":\"basic_set_name\",\"RPS[k]\":5.0,\"Tasks\":3000,\"Total[ms]\":600,\"busy[ms]\":400,\"idle[ms]\":200,\"m_gaps\":[0,200],\"m_incs\":[1000,2000],\"m_intervals\":[100,300]}"
[  FAILED  ] PerformancetimerTest.basic (704 ms)

arista-hpandya:
I was looking at the recent PR failures, and it seems like the test is running into issues because it's checking for exact millisecond timings.

This is a common issue, things like system load and OS scheduler delays mean that sleep_for is never perfectly exact, so the test can become a bit flaky. It doesn't look like there's any actual bug in your timer logic, it's just the test being a little too strict.

I'm working on a small patch to address this. The idea is to have the test check for a range of acceptable values instead of one exact number. This should make it much more stable and prevent false negatives on the CI. I'll have that ready for you to review soon!

Let me know what you think @a114j0y 

cc: @vmittal-msft

PR #895, opened 2024-07-22
[common] enable redispipeline to only publish after flush

a114j0y:
**What I did**
- optimize redispipeline flush performance by remove unnecessary publish commands
- add a new parameter`bool flushPub` in `producerstatetable` constructor function
     - to enable/disable batch publish feature
     - default value of `m_flushPub` is `false`, so no impact on existing codes
     - optimization is effective only explicitly set this option
- remove individual publish command from the `producerstatetable` APIs' lua scripts
- add a publish command when the pipeline flushes [if `m_flushPub` is `true`]

**Why I did it**

- save TCP traffic and increase fpmsyncd efficiency

It's a feature included in  [BGP Loading Optimization HLD #1521](https://github.com/sonic-net/SONiC/pull/1521)  GitHub issue/pull request detail

### DIFF
--- a/common/producerstatetable.cpp
+++ b/common/producerstatetable.cpp
@@ -31,7 +31,7 @@ ProducerStateTable::ProducerStateTable(RedisPipeline *pipeline, const string &ta
     , m_tempViewActive(false)
     , m_pipe(pipeline)
 {
-    resetRedisScript();
+    reloadRedisScript();
 
     string luaClear =
         "redis.call('DEL', KEYS[1])\n"
@@ -54,10 +54,18 @@ ProducerStateTable::~ProducerStateTable()
     }
 }
 
-void ProducerStateTable::resetRedisScript()
+void ProducerStateTable::reloadRedisScript()
 {
-    if (m_flushPub && m_buffered)
+    // Set m_flushPub to remove publish from a single lua string and let pipeline do publish once per flush
+
+    // However, if m_buffered is false, follow the original one publish per lua design
+    // Hence we need to check both m_buffered and m_flushPub, and reload the redis script once setBuffered() changes m_buffered
+
+    /* 1. Inform the pipeline of what channel to publish, when flushPub feature is enabled */
+    if (m_buffered && m_flushPub)
         m_pipe->addChannel(getChannelName(m_pipe->getDbId()));
+
+    /* 2. Setup lua strings: determine whether to attach luaPub after each lua string */
 
     // num in luaSet and luaDel means number of elements that were added to the key set,
     // not including all the elements already present into the set.
@@ -105,6 +113,7 @@ void ProducerStateTable::resetRedisScript()
         luaBatchedDel += luaPub;
     }
 
+    /* 3. load redis script based on the lua string */
     m_shaSet = m_pipe->loadRedisScript(luaSet);
     m_shaDel = m_pipe->loadRedisScript(luaDel);
     m_shaBatchedSet = m_pipe->loadRedisScript(luaBatchedSet);
@@ -114,7 +123,7 @@ void ProducerStateTable::resetRedisScript()
 void ProducerStateTable::setBuffered(bool buffered)
 {
     m_buffered = buffered;
-    resetRedisScript();
+    reloadRedisScript();
 }
 
 void ProducerStateTable::set(const string &key, const vector<FieldValueTuple> &values,
--- a/common/producerstatetable.cpp
+++ b/common/producerstatetable.cpp
@@ -14,68 +14,24 @@ using namespace std;
 namespace swss {
 
 ProducerStateTable::ProducerStateTable(DBConnector *db, const string &tableName)
-    : ProducerStateTable(new RedisPipeline(db, 1), tableName, false)
+    : ProducerStateTable(new RedisPipeline(db, 1), tableName, false, false)
 {
     m_pipeowned = true;
 }
 
 ProducerStateTable::ProducerStateTable(RedisPipeline *pipeline, const string &tableName, bool buffered)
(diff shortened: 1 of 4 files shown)


common/Makefile.am
```diff
@@ -64,7 +64,7 @@ common_libswsscommon_la_SOURCES = \
     common/countertable.cpp          \
     common/redisutility.cpp          \
     common/restart_waiter.cpp        \
-    common/profileprovider.cpp       \
+    common/profileprovider.cpp        \
```
liuh-80, line 67: This change seems not necessary, can this be revert?

common/producerstatetable.cpp
```diff
@@ -13,40 +14,37 @@ using namespace std;
...
+    : ProducerStateTable(new RedisPipeline(db, 1), tableName, false, flushPub)
 {
     m_pipeowned = true;
 }
 
-ProducerStateTable::ProducerStateTable(RedisPipeline *pipeline, const string &tableName, bool buffered)
+ProducerStateTable::ProducerStateTable(RedisPipeline *pipeline, const string &tableName, bool buffered, bool flushPub)
     : TableBase(tableName, SonicDBConfig::getSeparator(pipeline->getDBConnector()))
     , TableName_KeySet(tableName)
+    , m_flushPub(flushPub)
```
liuh-80, line 26: m_flushPub and some other fields missing in header file, there are build errors.

common/redispipeline.h
```diff
@@ -113,11 +117,19 @@ class RedisPipeline {
 
     void flush()
     {
+        lastHeartBeat = std::chrono::steady_clock::now();
```
liuh-80, line 120: What's these timing related code for?
If for performance measurement, please use DEBUG macro to only build in debug build.

common/redispipeline.h
```diff
@@ -113,11 +117,19 @@ class RedisPipeline {
...
+            return;
+        }
+
         while(m_remaining)
         {
             // Construct an object to use its dtor, so that resource is released
             RedisReply r(pop());
         }
+
+        publish();
```
liuh-80, line 132: Where is the definition of publish()?
And publish after flush is a behavior change, will this break existing scenario and impact performance? Please add test case.

common/redispipeline.h
```diff
@@ -145,12 +157,31 @@ class RedisPipeline {
         m_ownerTid = gettid();
     }
 
+    void addChannel(std::string channel)
+    {
+        m_luaPub +=
+        "redis.call('PUBLISH', '" + channel + "', 'G');";
+
+        m_shaPub = loadRedisScript(m_luaPub);
```
liuh-80, line 165: What will happen if cann this method multiple times? for example when 2 ProducerStateTable share same channel?

common/redispipeline.h
a114j0y: it's not a performance measurement, it's used in fpmsyncd to get idle time from last pipeline flush

common/redispipeline.h
```diff
@@ -145,12 +158,43 @@ class RedisPipeline {
...
+    }
+
+    int getIdleTime(std::chrono::time_point<std::chrono::steady_clock> tcurrent=std::chrono::steady_clock::now())
+    {
+        return static_cast<int>(std::chrono::duration_cast<std::chrono::milliseconds>(tcurrent - lastHeartBeat).count());
+    }
+
+    void publish() {
+        if (m_shaPub == "") {
```
liuh-80, line 177: suggest use string::empty()

liuh-80 (dismissed):

a114j0y:
I submit a commit per your advice, but it then requires another approval
now 🥲

And do you know why it fails a test case？

common/producerstatetable.cpp
```diff
@@ -59,48 +83,38 @@ ProducerStateTable::ProducerStateTable(RedisPipeline *pipeline, const string &ta
...
-    m_shaApplyView = m_pipe->loadRedisScript(luaApplyView);
-}
-
-ProducerStateTable::~ProducerStateTable()
-{
-    if (m_pipeowned)
+    if (!m_flushPub || !m_buffered)
     {
-        delete m_pipe;
+        string luaPub =
```
liuh-80, line 98: When m_flushPub is true, and m_buffered is false. after every redis operation, the lua code still will publish.

But the pipeline flush() method also will publish.

Why flush twice in this case？

a114j0y: if the flushPub feature is off, m_shaPub.empty() is true, then publish() API would do nothing
(only if m_flushPub is true and m_buffered is true, would flushpub feature be turned on)

liuh-80: Suggest improving code by:
1. Add comments to improve readability and understanding.
2. Add a test case to cover this scenario: ProducerStateTable p(&pipeline, tableName, true, false);

a114j0y: ProducerStateTable p(&pipeline, tableName, true, false) is the default case, since flushPub mode is off by default, hence it's already tested by the existing test cases;

so i just add `ProducerStateTable p(&pipeline, tableName, false, true)`
and p.setBuffered()

to test whether flushPub mode is successfully turned on and works as expected.

flushPub mode needs not only m_flushPub to be true, but also m_buffered to be true. So that we maintain the existing optimization which collapse commands with the same key when m_buffered is false.

eddieruan-alibaba:
@liuh-80 do you have further comments?

liuh-80 (approved):

a114j0y:
rebased

a114j0y:
just rebased on upstream/master

liuh-80 (approved):

PR #906, opened 2024-08-27
Account for control character as part of zmq_read_part

zbud-msft:
ADO: 28728116

Capturing control character in do_capture as part of eventd is difficult as control character can be delayed such that we it is not sometimes caught when initially starting capture service. (SUB -> PUB (part of zmq proxy)).

Instead of dropping the control character as part of eventd do_capture, I will drop it as part of zmq_message_read which eventd calls via events_service.

1) Check rc value of zmq_msg_recv
2) if rc == 1 (meaning 1 byte long)
3) verify it is control character, log that we received subscription messages
4) drop it such that empty event is sent (empty events are dropped by eventd)

LIBZMQ: Control character logic: https://github.com/zeromq/libzmq/blob/cbb9925a109152426e4a8a3f8f8eaeddc0516977/src/xsub.cpp#L102

Added UT to confirm, as well as should see success rate of sonic-mgmt testcases go up for telemetry/test_events

### DIFF
--- a/common/events_common.h
+++ b/common/events_common.h
@@ -312,8 +312,10 @@ struct serialization
         int rc = zmq_msg_recv(&msg, sock, flag);
         if (rc == 1) {
             char control_character = *(char*)zmq_msg_data(&msg);
-            if (control_character == 0x01) {
-                SWSS_LOG_INFO("Received subscription message when XSUB connect to XPUB");
+            if (control_character == 0x01 || control_character == 0x00) {
+                SWSS_LOG_INFO("Received subscription/unsubscription message when XSUB connect to XPUB: %c", control_character);
+            } else {
+                SWSS_LOG_DEBUG("Received non subscription based control character: %c", control_character);
             }
             rc = 0;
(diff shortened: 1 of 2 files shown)


common/events_common.h
```diff
@@ -310,16 +310,21 @@ struct serialization
         more = 0;
         zmq_msg_init(&msg);
         int rc = zmq_msg_recv(&msg, sock, flag);
-        if (rc != -1) {
+        if (rc == 1) {
+            char control_character = *(char*)zmq_msg_data(&msg);
+            if (control_character == 0x01) {
```
liuh-80, line 315: If the control_character not 1, do we need log or return error?

zbud-msft: Fixed. We won't return error, as we should drop any control characters or 1 byte messages. I do think we should log control characters that do come that are not subscription based "0x01 or 0x00".

liuh-80 (approved):

zbud-msft:
@StormLiangMS Can you please review for 202311?

PR #915, opened 2024-09-12
Initial implementation of C api

erer1243:
Implement a C interface to some of libswsscommon in support of sonic-dash-ha.

Related:
https://github.com/sonic-net/sonic-dash-ha/pull/6
https://github.com/sonic-net/sonic-swss-common/pull/921

Incoming follow up PR:
https://github.com/erer1243/sonic-swss-common/pull/1


common/c-api/util.h
```diff
@@ -0,0 +1,107 @@
...
+#include <iostream>
+#include <utility>
+
+#include "../logger.h"
+#include "../rediscommand.h"
+
+// In the catch block, we must abort because passing an exception across an ffi boundary is
+// undefined behavior. It was also decided that no exceptions in swss-common are recoverable, so
+// there is no reason to convert exceptions into a returnable type.
+#define SWSSTry(...)                                                                               \
```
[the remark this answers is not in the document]
erer1243: We will do a test to see if abort preserves the stack in a core dump

common/c-api/util.h
```diff
@@ -0,0 +1,107 @@
...
+        SWSS_LOG_ERROR("Aborting due to exception: %s", e.what());                                 \
+        std::abort();                                                                              \
+    }
+
+// T is anything that has a .size() method and which can be iterated over for pair<string, string>
+// eg unordered_map<string, string> or vector<pair<string, string>>
+template <class T> static inline SWSSFieldValueArray makeFieldValueArray(const T &in) {
+    SWSSFieldValueArray out;
+    out.len = (uint64_t)in.size();
+    out.data = (SWSSFieldValuePair *)malloc(out.len * sizeof(SWSSFieldValuePair));
```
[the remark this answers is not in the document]
erer1243: freeing is implemented on the other side of the ffi boundary

common/Makefile.am
```diff
@@ -1,13 +1,13 @@
 
 lib_LTLIBRARIES += common/libswsscommon.la
 
-dist_swss_DATA = \
-    common/consumer_state_table_pops.lua \
-    common/consumer_table_pops.lua \
+dist_swss_DATA =                               \
```
liuh-80, line 4: Can we revert all unnecessary change?
One reason is this may break cherry-pick to old branch.

common/zmqconsumerstatetable.cpp
```diff
@@ -66,29 +66,16 @@ void ZmqConsumerStateTable::handleReceivedData(const std::vector<std::shared_ptr
...
+    vkco.clear();
 
-        // For new data append to m_dataQueue during pops, will not be include in result.
-        count = m_receivedOperationQueue.size();
-        if (!count)
-        {
-            return;
-        }
-    }
+    std::lock_guard<std::mutex> lock(m_receivedQueueMutex);
```
liuh-80, line 71: I'm concerned about the performance impact of this change. The original code only locks when operating on m_receivedOperationQueue. This is because vkco.push_back(std::move(kco)) will copy a large amount of data and take a long time.

With this change, it may affect the ZMQ channel throughput.

common/c-api/consumerstatetable.cpp
```diff
@@ -0,0 +1,34 @@
+#include <cstdlib>
```
liuh-80, line 1: Why create this c wrapper? Is there some issue to use swss-common lib with https://github.com/rust-lang/rust-bindgen?

erer1243: bindgen does not help you call C++. it only works for C. It doesn't support types of undefined sizes (bool) or any classes (vector, string).

common/Makefile.am
erer1243: Sure

common/zmqconsumerstatetable.cpp
erer1243: There are two race conditions in the code as-is if 2+ threads are running pops(): `count` can be invalidated, leading to `auto& kco = m_receivedOperationQueue.front()` giving a null reference; `kco` can also be popped and freed while still being pushed onto `vkco`, leading to use-after-free. 

If we are really afraid of allocating while holding the lock, we should 1) lock & get count, 2) drop lock and allocate space in vkco, 3) re-lock and pop all with one lock. I really suspect that this won't matter though.

r12f: to avoid any potential perf issue. let's try this -
1. define a new temporary queue, which will be empty.
2. Lock the member queue and swap with the empty queue.
3. pop all items from the temporary queue and push them all into the output queue.

the item 3, there is no lock in place, which should minimize the overall block/lock time.

r12f: hope this solves the concern. @erer1243

r12f (dismissed):

common/zmqconsumerstatetable.cpp
erer1243: Riff's suggestion is a good idea. I think that will fix both sides of the issue.

common/table.h
```diff
@@ -229,16 +229,16 @@ class Table : public TableBase, public TableEntryEnumerable {
...
-    virtual bool hget(const std::string &key, const std::string &field,  std::string &value);
+    virtual bool hget(const std::string &key, const std::string &field,  std::string &value) override;
     virtual void hset(const std::string &key,
                           const std::string &field,
                           const std::string &value,
                           const std::string &op = "",
                           const std::string &prefix = EMPTY_PREFIX);
 
-    void getKeys(std::vector<std::string> &keys);
+    virtual void getKeys(std::vector<std::string> &keys) override;
```
r12f, line 241: the parent is virtual, hence warning here. this is a good catch.

common/zmqconsumerstatetable.cpp
liuh-80: This change is not necessary. By design, the pops operation in sonic-swss-common is not thread-safe. Similar to all other table objects, the ConsumerStateTable object should not be shared by multiple threads.

erer1243: @liuh-80 Why, then, is there a mutex at all?

erer1243: Also, I think the new implementation is more performant anyway

r12f: yea, Hua. I am confused as well now. I do agree with you - In my understanding, the ConsumerStateTable is not going to be shared to multiple threads as well. But I am seeing you added this function with lock added. In that case, I think the same as Oliver - it will be better to remove the lock for good.

If we would like to keep the lock, then I feel the new implementation is more performant because it is going to block less indeed.

r12f: Please let me know which way you would like to go @liuh-80 ! :D

erer1243: On further thought, I understand that the lock is there to synchronize with the ZmqServer thread, and that since the queue's count is monotonically increasing in that thread, the race conditions I mentioned can never happen. I do still think that the newest implementation is faster, though.

r12f: ah, the background thread! that makes sense!

common/table.h
qiluo-msft: Agree, let's separate the bugfix into a standalone PR for easy review/revert/cherrypick.

common/zmqconsumerstatetable.cpp
```diff
@@ -66,29 +66,16 @@ void ZmqConsumerStateTable::handleReceivedData(const std::vector<std::shared_ptr
...
-        std::lock_guard<std::mutex> lock(m_receivedQueueMutex);
-
-        // For new data append to m_dataQueue during pops, will not be include in result.
-        count = m_receivedOperationQueue.size();
-        if (!count)
-        {
-            return;
-        }
+        lock_guard<mutex> l(m_receivedQueueMutex);
+        swap(m_receivedOperationQueue, q);
```
qiluo-msft, line 72: >swap [](http://example.com/codeflow?start=8&length=4)

If this is a bugfix for ZMQ, let's use a standalone PR.

common/c-api/util.h
```diff
@@ -0,0 +1,181 @@
...
+#define SWSSTry(...)                                                                               \
+    if (cApiTestingDisableAbort) {                                                                 \
+        { __VA_ARGS__; }                                                                           \
+    } else {                                                                                       \
+        try {                                                                                      \
+            { __VA_ARGS__; }                                                                       \
+        } catch (std::exception & e) {                                                             \
+            std::cerr << "Aborting due to exception: " << e.what() << std::endl;                   \
+            SWSS_LOG_ERROR("Aborting due to exception: %s", e.what());                             \
+            std::abort();                                                                          \
```
qiluo-msft, line 73: >abort [](http://example.com/codeflow?start=17&length=5)

If I understand correctly, all exception in this repo will abort an rust application. This is behavior change. Could you keep original behavior?

tests/c_api_ut.cpp
```diff
@@ -0,0 +1,325 @@
+#include <cstdlib>
```
qiluo-msft, line 1: >cstdlib [](http://example.com/codeflow?start=10&length=7)

the only usage of c_api is rust. Why not just test in rust language?

erer1243: It is tested in rust, but without these tests this repo fails code coverage checks 😛.

common/zmqconsumerstatetable.cpp
erer1243: Sure

common/table.h
erer1243: Ok sure thing

common/c-api/util.h
erer1243: Unfortunately this is not possible. It is UB to allow a c++ exception to unwind into rust, so there are two options: 1. catch exceptions and convert into normal data types which can be returned to rust to recover from, or 2. abort on exception. Option 1 was decided against because all thrown exceptions are basically fatal (out of memory, invalid config file, etc). Option 2 also gives us a core dump to debug against. If we were to ignore the exceptions, rust would abort automatically anyway but with a bad core dump.

common/zmqconsumerstatetable.cpp
erer1243: Removed these changes from this PR. If you think these are valid changes @liuh-80, I will put them into a new PR. Otherwise we can forget about it.

liuh-80 (approved):

r12f (approved):

common/c-api/util.h
qiluo-msft: Thanks for the explanation!

 all thrown exceptions are basically fatal -> I do not agree. Some exceptions are related to runtime env for example a redis server could not respond right now but may respond after retry. ref: [https://github.com/sonic-net/sonic-swss-common/blob/00445404fff6d3673464e2a5defca1dc97ab3a43/common/dbconnector.cpp#L564](https://github.com/sonic-net/sonic-swss-common/blob/00445404fff6d3673464e2a5defca1dc97ab3a43/common/dbconnector.cpp#L564)

qiluo-msft:
@erer1243 In future could you reduce the force merge during PR iteration? We are using tool to compare between commits to understand what is newly added, and it improve code review efficiency.

common/c-api/util.h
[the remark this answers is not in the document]
erer1243: Did NOT use cpptrace - we made exceptions abort to preserve a stack trace in the core dump.

common/c-api/util.h
erer1243: Hm, I understand and that is a good point, but this leaves us with a couple of non-ideal options. 

## Option 1: 
Enumerate all non-fatal exceptions and code those into the signature of the C apis.
Eg: we would change SWSSDBConnector_new_tcp to something like this
```c
// If this function returns null, db did not respond, but it might if you retry
SWSSDBConnector SWSSDBConnector_new_tcp(int32_t dbId, const char *hostname, uint16_t port, uint32_t timeout) {
    SWSSTry({
        try {
            return (SWSSDBConnector) new DBConnector(dbId, string(hostname), port, timeout);
        } catch (system_error &e) {
            // we know this error is nonfatal so we hard coded it in here
            return nullptr;
        }

        // any other error will still abort() because any other error must be fatal
    });
}
```
This is not ideal because now the C api becomes logically coupled to the implementation of the underlying function. If another nonfatal exception is added to DBConnector, they also have to change this C function to properly return null. We will have to do this for every possible nonfatal exception that one of the C functions might cause. 

This may be OK if there are very few nonfatal exceptions, we expect to practically never add more, and if the C api stays very small. Otherwise, anyone who works on swsscommon now also has to understand these implications on the C api.

## Option 2 simple: 
Introduce a generic error interface to the C api. This is a very bad way of doing it, but I mention it because this is how I previously implemented the C api. We store errors as a global error string, and SWSSTry + code using it looks something like this:
```c

#include <cstring> // for strdup
#include <cstdlib> // for free

const char *globalErrorString = nullptr;

extern "C" { const char *SWSSGetError(void) { return globalErrorString; } }

#define SWSSTry(failure_value, ...) \
    try { \
        __VA_ARGS__ \
        ; \
    } catch (std::exception &e) { \
        free(globalErrorString); \
        globalErrorString = strdup(e.what().c_str()); \
        return failure_value; \
    }

SWSSDBConnector SWSSDBConnector_new_tcp(int32_t dbId, const char *hostname, uint16_t port, uint32_t timeout) {
    SWSSTry(nullptr, {
            // upon any exception, returns nullptr and globalErrorString is set to e.what()
            return (SWSSDBConnector) new DBConnector(dbId, string(hostname), port, timeout)
    });
}
```

On the rust (or any other language) side:
```rs
fn make_a_dbconnector() {
    loop {
        let dbconnector = unsafe { SWSSDBConnector_new_tcp(...) };
        if dbconnector.is_null() {
            // If any error occurred fatal or nonfatal, all we get is a descriptive string
            let error_str =  unsafe { str::from_utf8_unchecked(SWSSGetError()) };
            
            // Magic strings!! yuck!!
            if error_str.contains("address not available") {
                // nonfatal error - retry later
               sleep_10_seconds();
            } else {
                panic!("fatal error: {error_str}");
            }
        } else {
            return dbconnector;
        }
    }
}
```
I originally wrote the api this way, but it was nixed by Riff (https://github.com/sonic-net/sonic-swss-common/pull/915#discussion_r1757420168) because we are losing the core dump/stack trace for fatal errors. Another problem is that all we get is a string - if we want to catch specific exceptions, we have to know what `e.what()` will return. This is really bad because foreign code is logically coupled to the exact exception messages in swss-common. If somebody fixes a typo in an exception message, that the magic string in foreign code is now wrong!

## Option 2 advanced version:
Here's another approach to a generic error interface that may be less bad. Introduce a generic try/catch to the C interface, like so:
```c
bool globalDisableAborts = false;
const char *globalErrorString = nullptr;

extern "C" { 
    void SWSSTry_(void) { free(globalErrorString); globalErrorString = nullptr; globalDisableAborts = true; }
    const char *SWSSCatch(void) { globalDisableAborts = false; return globalErrorString; } 
}

#define SWSSTry(failure_value, ...) \
    try { \
        __VA_ARGS__ \
        ; \
    } catch (std::exception &e) { \
        if (globalDisableAborts) {
            free(globalErrorString); \
            globalErrorString = strdup(e.what().c_str()); \
            return failure_value; \
        } else { \
            SWSS_LOG_ERROR(... e.what() ...); \
            abort(); \
        } \
    }

```rs
fn make_a_dbconnector_retrying() {
    // We will only try 10 times before performing an aborting call
    for tries in 0..10 {
        // begin a pseudo try block
        unsafe { SWSSTry_() }; 
        let dbconnector = unsafe { SWSSDBConnector_new_tcp(...) };
        // end the pseudo try block, get the error if one happened.
        let err = unsafe { SWSSCatch() }; 
        if err.is_null() {
            // happy path: error is null so there was no error - dbconnector is valid
            return dbconnector;
        } else {
            // This might've been a fatal or nonfatal error but we aren't checking with a magic string. 
            // We don't need to check because we are limiting our number of tries.
            let error_str =  unsafe { str::from_utf8_unchecked(SWSSGetError()) };
            println!("error: {error_str}, retrying");
            sleep_10_seconds();
        }
    }

    // We have tried 10 times - it's probably a fatal error, so let's bite the bullet and do an aborting call
    // Since we did not use SWSSTry_, any exception will still abort. Just like current behavior.
    // We will get a nice core dump because abort() was called;
    let dbconnector = unsafe { SWSSDBConnector_new_tcp(...) };
    return dbconnector;
}
```

This is the best idea IMO because we can choose where to abort and where not to. Using `SWSSTry_` and `SWSSCatch` we can retry something a few times before electing to lay down our cards and get a proper core dump. In simple cases, where any error is fatal, we can simply ignore `SWSSTry_` and `SWSSCatch`, and use aborting behavior on the first try.

## Final thoughts
If there are very very few nonfatal errors in code that the C api uses, and we can find all of them and hard code their nonfatality into the C api functions, then we should do option 1. 

If the C api is going to grow a lot later, or swss-common might change what errors are nonfatal, or we really want the C api to be very simple, or we want to allow people working on swss-common to ignore the C api, or we want to give generic error handling decisions to downstream foreign code, I think we should do option 2 advanced.

r12f: Even with this it still creates challenge in the Humberger stack, where c++ -> rust -> c++, because with return code approach, every function in rust need to understand the error code and passing it to the upper layer. This limits the error handling in rust to only Anyhow-ish crates with downcast...

This is a problem that me and Oliver struggled for days on finding the solutions, but so far there is no obvious/quick answer to it. So I have discussed this issue with Qi offline. We will need to fix this issue, but since it is only impacting the services that uses rust in the exception handling path, we are going to create a backlog to get this fixed, and get the current PR merged to unblock our immediate project progress.

r12f: this issue is fundamentally caused by rust not having try-catch support, and we are trying to recreate the try-catch support for the language, which will take time to get it right.

r12f: Work Item created here: MSADO:29998324 and assigned to Oliver at this moment.

r12f (dismissed):
With the backlog created and sync'ed with Qi, approving this PR, so we can get it merged.

erer1243:
Sorry, forgot to include the c api stuff in the deb packages :)

common/c-api/util.h
erer1243: > this issue is fundamentally caused by rust not having try-catch support, and we are trying to recreate the try-catch support for the language, which will take time to get it right.

Rust panics use the same mechanism as c++ exceptions - it really does have try/catch, but the issue is that no C ABIs permit unwinding. Technically it would even be incorrect if we called these C api functions from within another C++ program, if we crossed a dynamic-linking boundary (i.e. the "hamburger stack" would look like C++ -> C -> C++). This is because internal ABI details are negotiated per linkage unit (at least without other flags/magic applied like rust's C-unwind). If we had a C++ function that called a C function, it's valid to optimize out any unwind-catching code because, by spec, no C function can ever throw. Rust is also making this assumption - that no C function can ever unwind.

erer1243: Personally I think we should either implement the "option 2 advanced" thing I wrote, or just ignore this for now until catching specific errors becomes necessary in hamgrd or other services.

r12f (approved):

r12f:
sync'ed with Qi. all comments are resolved now, hence merging it.

PR #918, opened 2024-09-17
Add missing flex counter operations to consumer_table_pops.lua used in async mode

aviramd:
when configure  "synchronous_mode": "disable" the syncd and swss containers crashes
because of missing flex-counter operations in consumer_table_pops.lua

Fixed Problems/New Changes:

https://github.com/sonic-net/sonic-swss-common/issues/917

Should be Tested:


aviramd:
@prsunny can you review this PR

PR #921, opened 2024-09-24
Add BinarySerializer function that uses a thread-local shared buffer, and fix errors

erer1243:
Fixed dubious undefined behavior stemming from unaligned writes to the serialization buffer. Add `serializeSharedBuffer` which uses a shared 16MB buffer instead of requiring the caller to have one available. This would make c-api code which uses these functions significantly more efficient. If we like these changes, we can also change ZmqClient to use the shared buffer functions. Every ZmqClient currently allocates its own 16MB serialization buffer, so doing this would save a nonnegligible amount of memory.

Also fixed a couple misc errors - uninitialized array in tests/stringutility_ut.cpp made tests fail to compile; ZMQ_NOBLOCK was deprecated 6 years ago and is now called ZMQ_DONTWAIT. I just noticed these while making changes to & testing BinarySerializer.

Related: https://github.com/sonic-net/sonic-swss-common/pull/915

### DIFF
--- a/common/zmqclient.cpp
+++ b/common/zmqclient.cpp
@@ -119,7 +119,7 @@ void ZmqClient::sendMsg(
         const std::vector<KeyOpFieldsValuesTuple>& kcos,
         std::vector<char>& sendbuffer)
 {
-    size_t serializedlen = serializeBuffer(sendbuffer.data(), sendbuffer.size(), dbName, tableName, kcos);
+    size_t serializedlen = BinarySerializer::serializeBuffer(sendbuffer.data(), sendbuffer.size(), dbName, tableName, kcos);
 
     if (serializedlen >= MQ_RESPONSE_MAX_COUNT)
     {
--- a/common/zmqclient.cpp
+++ b/common/zmqclient.cpp
@@ -119,37 +119,32 @@ void ZmqClient::sendMsg(
         const std::vector<KeyOpFieldsValuesTuple>& kcos,
         std::vector<char>& sendbuffer)
 {
-    int serializedlen = (int)BinarySerializer::serializeBuffer(
-                                                        sendbuffer.data(),
-                                                        sendbuffer.size(),
-                                                        dbName,
-                                                        tableName,
-                                                        kcos);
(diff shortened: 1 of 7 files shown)


saiarcot895:
Can you elaborate on the undefined behavior that you saw?

erer1243:
> Can you elaborate on the undefined behavior that you saw from the unaligned writes?

Lines like these are UB, as they are writing a size_t to char*. These break strict aliasing/provenance and alignment rules. The WARNINGS_NO_CAST_ALIGN macro is telling gcc to not print the warning, but the problem still exists. The solution to both of these problems is memcpy, so I replaced those pointer casts with read/write_unaligned.
https://github.com/sonic-net/sonic-swss-common/blob/24979b05ff9c2daa18432abcd0dd8518bd5c7e60/common/binaryserializer.h#L56-L58

saiarcot895:
SONiC currently only supported amd64, armhf, and arm64, all of which support unaligned memory access (for armhf, ARMv7 and newer support it, which is what I think compile for). Nevertheless, it would be nice to allow SIMD operations by making sure the memory is aligned.

saiarcot895 (approved):

common/zmqclient.cpp
```diff
@@ -119,37 +119,32 @@ void ZmqClient::sendMsg(
...
-    for (int i = 0; i <=  MQ_MAX_RETRY; ++i)
+    for (int i = 0; i <= MQ_MAX_RETRY; ++i)
     {
         {
             // ZMQ socket is not thread safe: http://api.zeromq.org/2-1:zmq
             std::lock_guard<std::mutex> lock(m_socketMutex);
 
-            // Use none block mode to use all bandwidth: http://api.zeromq.org/2-1%3Azmq-send
-            rc = zmq_send(m_socket, sendbuffer.data(), serializedlen, ZMQ_NOBLOCK);
+            // Use nonblocking mode to use all bandwidth: http://api.zeromq.org/2-1%3Azmq-send
```
qiluo-msft, line 141: >nonblocking [](http://example.com/codeflow?start=19&length=11)

This fix is not relevant to your PR title. Suggest use another smaller PR.

PR #922, opened 2024-09-25
add support for debug generate dump tables

aviramd:
This update adds support for the sai_dbg_gen_dump API, enabling the generation of a debug dump file by the SAI when there is no SAI failure. For scenarios involving SAI failures, a separate flow called dump_on_sai_failure is available.

An HLD will be publish soon with all relevant PRs


aviramd:
Hi @qiluo-msft, could you please review and, if everything looks good, integrate this PR? It is part of the HLD: [sonic-net/SONiC#1846](https://github.com/sonic-net/SONiC/pull/1846). As part of adding support for the debug generate dump file, SWSS has dependencies on new APPL DB tables defined in this PR. Due to these dependencies, the SWSS PR [sonic-net/sonic-swss#3301](https://github.com/sonic-net/sonic-swss/pull/3301) is currently failing its pipeline.

PR #931, opened 2024-10-21
Add helper function to validate interface name length

stepanblyschak:
This PR is accompanied by https://github.com/sonic-net/sonic-buildimage/pull/20108 and https://github.com/sonic-net/sonic-utilities/pull/3580

Add helper function to validate that interface name length does not exceed allowed length configured as IFNAMSIZ. This will be used both by sonic-utilities CLI (config) and by cfgmgr.


common/interface.h
```diff
@@ -0,0 +1,25 @@
...
+const size_t IFACE_NAME_MAX_LEN = IFNAMSIZ - 1;
+
+bool isInterfaceNameValid(const std::string &ifaceName)
+{
+    return !ifaceName.empty() && (ifaceName.length() < IFNAMSIZ);
+}
+
+#if defined(SWIG) && defined(SWIGPYTHON)
+%pythoncode %{
+    def is_interface_name_valid(iface_name):
```
qiluo-msft, line 18: If python could call `isInterfaceNameValid`, no need to define another one just different with function name. We previoslyu defined some python function due to SWIG could not automatically translate function from one language to another, like C++ missing dynamic features.

stepanblyschak: I don't have a strong argument besides that I wanted to keep original author's intent to have a function named in snake_case style

stepanblyschak: @qiluo-msft Done

PR #935, opened 2024-10-30
C API: improve performance by not memcpy-ing string DB values across ffi boundary 

erer1243:
This introduces a couple of new FFI types, representing owned C++ strings and arrays. We use those owned strings for database values, so they do not need to be memcpy'd across ffi boundaries. This is in response to Riff's concerns that database values which hamgrd will be reading may be large and expensive to memcpy.

Partner pr to https://github.com/sonic-net/sonic-dash-ha/pull/11

### DIFF
--- a/common/c-api/dbconnector.cpp
+++ b/common/c-api/dbconnector.cpp
@@ -1,5 +1,6 @@
 #include <cstring>
 #include <string>
+#include <utility>
 
 #include "../dbconnector.h"
 #include "dbconnector.h"
@@ -37,14 +38,14 @@ int8_t SWSSDBConnector_del(SWSSDBConnector db, const char *key) {
     SWSSTry(return ((DBConnector *)db)->del(string(key)) ? 1 : 0);
 }
 
-void SWSSDBConnector_set(SWSSDBConnector db, const char *key, const char *value) {
-    SWSSTry(((DBConnector *)db)->set(string(key), string(value)));
+void SWSSDBConnector_set(SWSSDBConnector db, const char *key, SWSSStrRef value) {
+    SWSSTry(((DBConnector *)db)->set(string(key), takeStrRef(value)));
 }
 
-char *SWSSDBConnector_get(SWSSDBConnector db, const char *key) {
+SWSSString SWSSDBConnector_get(SWSSDBConnector db, const char *key) {
     SWSSTry({
         shared_ptr<string> s = ((DBConnector *)db)->get(string(key));
-        return s ? strdup(s->c_str()) : nullptr;
+        return s ? makeString(move(*s)) : nullptr;
     });
 }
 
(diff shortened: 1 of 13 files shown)


common/c-api/dbconnector.cpp
```diff
@@ -37,14 +38,14 @@ int8_t SWSSDBConnector_del(SWSSDBConnector db, const char *key) {
     SWSSTry(return ((DBConnector *)db)->del(string(key)) ? 1 : 0);
 }
 
-void SWSSDBConnector_set(SWSSDBConnector db, const char *key, const char *value) {
-    SWSSTry(((DBConnector *)db)->set(string(key), string(value)));
+void SWSSDBConnector_set(SWSSDBConnector db, const char *key, SWSSStrRef value) {
```
liuh-80, line 41: I can't find the definition of SWSSStrRef, where is the source code?

common/c-api/util.h
```diff
@@ -93,22 +143,21 @@ static inline SWSSSelectResult selectOne(swss::Selectable *s, uint32_t timeout_m
...
+    SWSSFieldValueTuple *data = new SWSSFieldValueTuple[in.size()];
 
     size_t i = 0;
-    for (const auto &pair : in) {
-        SWSSFieldValuePair entry;
+    for (auto &pair : in) {
+        SWSSFieldValueTuple entry;
         entry.field = strdup(pair.first.c_str());
-        entry.value = strdup(pair.second.c_str());
+        entry.value = makeString(std::move(pair.second));
```
liuh-80, line 160: Already move inside makeString(), maybe don't need move here?

common/c-api/dbconnector.cpp
erer1243: https://github.com/sonic-net/sonic-swss-common/blob/3017a8f8fb4b237070c148b6395ef529451cadd2/common/c-api/util.h#L15-L19

common/c-api/util.h
erer1243: You need the move call there to be able to call the function. C++ won't automatically make an rvalue ref

```
common/c-api/util.h:160:39: error: cannot bind rvalue reference of type 'std::string&&' {aka 'std::__cxx11::basic_string<char>&&'} to lvalue of type 'std::__cxx11::basic_string<char>'
  160 |         entry.value = makeString(pair.second);
      |                                  ~~~~~^~~~~~
common/c-api/util.h:146:51: note:   initializing argument 1 of 'SWSSStringOpaque* swss::makeString(std::string&&)'
  146 | static inline SWSSString makeString(std::string &&s) {
      |                                     ~~~~~~~~~~~~~~^
```

lguohan:
@erer1243 , can you help to resolve the conflict?

lguohan (dismissed):

erer1243:
Closed it from git cli by accident :stuck_out_tongue:

erer1243:
@lguohan should be good now

lguohan (approved):

PR #952, opened 2024-11-14
[EVPN-MH] Added new APP_L2_NEXTHOP_GROUP_TABLE_NAME table in APPL_DB

kishorekunal01:
Why I did it
Added the new APP_L2_NEXTHOP_GROUP_TABLE_NAME table in APPL_DB, along with support for registering and deregistering raw Netlink messages.

Direct handling of raw Netlink messages is required because libnl3 does not support Next Hop Group (NHG) message processing. Therefore, fdbsyncd needs to decode Kernel NHG messages to update the L2_NEXTHOP_GROUP_TABLE.

Check HLD: https://github.com/sonic-net/SONiC/pull/1702

Signed-off-by: Kishore Kunal <kishore.kunal@broadcom.com>


adyeung:
@pbrisset please have someone from CSCO to help review

liuh-80:
@kishorekunal01 , can you improve code coverage by add new test case?

common/netdispatcher.cpp
```diff
@@ -35,12 +35,45 @@ void NetDispatcher::unregisterMessageHandler(int nlmsg_type)
...
+
 void NetDispatcher::nlCallback(struct nl_object *obj, void *context)
 {
     NetMsg *callback = (NetMsg *)context;
     callback->onMsg(nl_object_get_msgtype(obj), obj);
 }
 
+void NetDispatcher::onNetlinkMessageRaw(struct nl_msg *msg)
+{
+    struct nlmsghdr *nlmsghdr = nlmsg_hdr(msg);
```
liuh-80, line 65: I notice there is a MUTEX in NetDispatcher::getCallback(int nlmsg_type)
Is this method also need a MUTEX?

common/netdispatcher.h
```diff
@@ -23,9 +23,24 @@ namespace swss
              */
             void registerMessageHandler(int nlmsg_type, NetMsg *callback);
 
+            /*
+             * Register callback class according to message-type.
+             *
+             * Throw exception if
```
liuh-80, line 29: Please improve this code comments.

common/netdispatcher.cpp
```diff
@@ -35,12 +35,45 @@ void NetDispatcher::unregisterMessageHandler(int nlmsg_type)
     m_handlers.erase(it);
 }
 
+void NetDispatcher::registerRawMessageHandler(int nlmsg_type, NetMsg *callback)
+{
+    if (m_rawhandlers.find(nlmsg_type) != m_rawhandlers.end())
+        throw "Trying to registered on already registerd netlink message";
```
FengPan-Frank, line 41: better to throw specific exception type like runtime_error instead of string literals.

common/netdispatcher.cpp
```diff
@@ -35,12 +35,45 @@ void NetDispatcher::unregisterMessageHandler(int nlmsg_type)
     m_handlers.erase(it);
 }
 
+void NetDispatcher::registerRawMessageHandler(int nlmsg_type, NetMsg *callback)
```
FengPan-Frank, line 38: is m_rawhandlers protected by m_mutex? as well as unregisterRawMessageHandler() function.

common/netdispatcher.cpp
kishorekunal01: This is similar to the method registerMessageHandler. Hence similar error handling is done.

common/netdispatcher.cpp
kishorekunal01: Good catch, Sure I will put this in mutex as well.

common/netdispatcher.cpp
kishorekunal01: This method doesn't need mutex as this method is only called from getCallback

common/netdispatcher.h
kishorekunal01: Sure, I will update this.

liuh-80:
@kishorekunal01 , you need add new test case to cover code change:

Total: 20 lines
Missing: 20 lines
Coverage: 0%
Threshold: 80%

kishorekunal01:
> @kishorekunal01 , you need add new test case to cover code change:
> 
> Total: 20 lines Missing: 20 lines Coverage: 0% Threshold: 80%

Sure.

common/netdispatcher.cpp
selva-nexthop: onNetlinkMessageRaw() is called from onNetlinkMessage() (line 101-105), which does NOT hold the mutex.

The new raw message handler registration functions have MUTEX protection, but onNetlinkMessageRaw() (line 69) does NOT have mutex protection when accessing m_rawhandlers.

Risk: Race condition between:
- Thread A: registerRawMessageHandler() modifies m_rawhandlers (under mutex)
- Thread B: onNetlinkMessageRaw() reads m_rawhandlers (NO mutex)
- Could lead to iterator invalidation or undefined behavior

common/netdispatcher.cpp
```diff
@@ -35,12 +35,49 @@ void NetDispatcher::unregisterMessageHandler(int nlmsg_type)
...
+    struct nlmsghdr *nlmsghdr = nlmsg_hdr(msg);
+    NetMsg *reg_callback = NULL;
+    auto callback = m_rawhandlers.find(nlmsghdr->nlmsg_type);
+
+    /* Drop not registered messages */
+    if (callback == m_rawhandlers.end())
+        return;
+
+    reg_callback = (NetMsg *)callback->second;
+    reg_callback->onMsgRaw(nlmsghdr);
```
selva-nexthop: If m_rawhandlers[nlmsg_type] somehow contains nullptr, this will segfault.

selva-nexthop (changes requested):

PR #954, opened 2024-11-21
use "\n" instead of semicolon to separate redis commands in a c++ string

a114j0y:
replace ";" with "\n" when concatenating multiple redis cmds

### DIFF
--- a/common/redispipeline.h
+++ b/common/redispipeline.h
@@ -164,7 +164,7 @@ public:
             return;
 
         m_channels.insert(channel);
-        m_luaPub += "redis.call('PUBLISH', '" + channel + "', 'G');";
+        m_luaPub += "redis.call('PUBLISH', '" + channel + "', 'G')\n";
         m_shaPub = loadRedisScript(m_luaPub);
     }
 


liuh-80 (approved):

FengPan-Frank (approved):

qiluo-msft (approved):

lguohan:
@qiluo-msft , @liuh-80 , @yuezhoujk , please explain motivation of such change.

a114j0y:
> @qiluo-msft , @liuh-80 , @yuezhoujk , please explain motivation of such change.

We see such an exception when running the testcase in the azure pipeline:

C++ exception with description "RedisReply catches system_error: command: *3\r\n$6\r\nSCRIPT\r\n$4\r\nLOAD\r\n$52\r\nredis.call('PUBLISH', 'ROUTE_TABLE_CHANNEL@0', 'G');\r\n, reason: Expected to get redis type 1 got type 3, err: NON-STRING-REPLY: Input/output error: Input/output error" thrown in SetUp().

which indicates that the redis server fails to do SCRIPT LOAD, although it works on our end.

We checked the c++ string `m_luaPub` that is to be loaded into a redis script:
m_luaPub += "redis.call('PUBLISH', '" + channel + "', 'G');";

it could be successfully loaded into script on our end, but throws exception in the azure pipeline.

We think it's due to the redis server of some specific version fails to parse ";" as a part of the redis script.

PR #960, opened 2024-12-11
Return allocated memory with malloc_trim when queue is empty

theasianpianist:
When device is encountering rapid MAC moves, malloc_trim gets called anywhere from 50-300 times per second depending on MAC move rate. 

When device is stable, malloc_trim gets called 5-10 times per hour in short bursts.


common/notificationconsumer.cpp
```diff
@@ -154,6 +155,27 @@ void swss::NotificationConsumer::pop(std::string &op, std::string &data, std::ve
...
+    if (m_queue.empty())
+    {
+        /***
+         * If there is a burst of notifications that causes the queue to grow in size,
+         * memory allocated by the queue will not be released even after the all items 
+         * have been popped.
+         * 
+         * Force the memory to be released by destroying existing queue and creating a new one.
+         */
+        SWSS_LOG_DEBUG("%s queue is empty, calling malloc_trim()", m_channel.c_str());
```
prsunny, line 167: Can we move this to `INFO `level?

prsunny (commented):
Could you update description with stats of this api call in a stable system vs continuous notification scenario?

theasianpianist:
Closing in favor of https://github.com/sonic-net/sonic-swss/pull/3473

PR #963, opened 2024-12-30
Fix warnings in BinarySerializer

erer1243:
This PR adjusts BinarySerializer in these ways: 
- make m_buffer not const (we write to it, why was it ever const)
- unaligned reads/writes are done correctly, using std::memcpy
- signedness cast warnings are fixed

A short discussion on this was held here previously: https://github.com/sonic-net/sonic-swss-common/pull/921#issuecomment-2372670379

On x86/ARM, unaligned reads/writes are acceptable with general purpose registers, but they may crash on x86 if the optimizer decides to use a SIMD register.


common/binaryserializer.h
```diff
@@ -34,7 +43,7 @@ class BinarySerializer {
     }
 
     static size_t serializeBuffer(
-        const char* buffer,
+        char* buffer,
```
hdwhdw, line 46: Just curious why remove const here? Are we modifying the buffer?

erer1243: Yes, the function serializes kcos into buffer.

liuh-80 (approved):

common/binaryserializer.h
```diff
@@ -11,6 +10,16 @@ using namespace std;
 
 namespace swss {
 
+template <class T> static inline T read_unaligned(const char *buffer) {
```
FengPan-Frank, line 13: do we need to check the case that T R/W has enough space for buffer memcpy?

erer1243: By the time these functions are called, the de/serializer has already checked the buffer's capacity.

erer1243:
Closed as unlikely to be merged. Feel free to copy the code and reopen.

PR #964, opened 2025-01-08
Add swss::Table to c api

erer1243:


r12f (approved):
looks straight forward change to me.

common/c-api/util.cpp
```diff
@@ -31,3 +31,7 @@ void SWSSFieldValueArray_free(SWSSFieldValueArray arr) {
 void SWSSKeyOpFieldValuesArray_free(SWSSKeyOpFieldValuesArray kfvs) {
     SWSSTry(delete[] kfvs.data);
 }
+
+void SWSSStringArray_free(SWSSStringArray arr) {
+    SWSSTry(delete[] arr.data);
```
liuh-80, line 36: This is a "const char **data;", do you need delete all string pointer in this array first?

erer1243: The c api free functions don't do that because the data is sometimes user-owned.

liuh-80: Please check the makeStringArray and SWSSTable_getKeys.
If this method don't free string, there will be memory leak.

erer1243: The string array can contain user-owned data (strings from the other side of the ffi). Thus, the freeing must happen on the foreign side. This pattern is the same as SWSSFieldValueArray_free, and SWSSKeyOpFieldValuesArray_free, and those do not have memory leak issues.

liuh-80 (approved):

PR #965, opened 2025-01-09
[TEST PR] Extend timeout to build swss

stephenxs:


stephenxs:
it's a statistical issue on the server or env which has been resolved. no longer required.

PR #969, opened 2025-01-24
Add swss::Logger to c-api

yue-fred-gao:
### Why
We need to hook up to swss::Logger from swbusd/hamgrd in rust, which requires us to add the c-api for logger.

### What this PR does
Add c-api for swss::Logger and corresponding unit test


erer1243:
Looks good to me. CI errors seem unrelated.

common/c-api/logger.cpp
```diff
@@ -0,0 +1,29 @@
+#include <cstring>
```
FengPan-Frank, line 1: is <cstring> really needed here?

yue-fred-gao: Indeed. I removed it. Thanks

liuh-80 (approved):

common/c-api/logger.cpp
FengPan-Frank: Thanks for the fix, could you sync branch to complete verification as well?

r12f:
merged since Hua signed off and CI passed.

PR #978, opened 2025-02-24
Test with new agent os version

xwjiang-ms:
Test with new agent os version


Pterosaur:
Please wait for the PR: https://github.com/sonic-net/sonic-swss/pull/3549 to fix it.

PR #986, opened 2025-03-14
Enhance swss-common logging by writing a warning log when a command fails due to Redis loading the dataset.

liuh-80:
Enhance swss-common logging by writing a warning log when a command fails due to Redis loading the dataset.

#### Why I did it
Fix issue https://github.com/sonic-net/sonic-buildimage/issues/17472
sonic-db-cli write error log, this is because sonic-db-cli execute PING command when Redis is loading dataset.

#### How I did it
When command failed because Redis is loading dataset, write warning log instead of error log.

##### Work item tracking
- Microsoft ADO: 31817720

#### How to verify it
Pass all test cases.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Enhance swss-common logging by writing a warning log when a command fails due to Redis loading the dataset.


common/redisreply.h
```diff
@@ -114,5 +116,30 @@ class RedisReply
...
+    try
+    {
+        func();
+    }
+    catch (const std::system_error& ex)
+    {
+        // Combine more error message and throw again
+        std::string reason = ex.what();
+        std::string errmsg = "RedisReply catches system_error: command: " + std::string(command) + ", reason: " + reason;
+        if (reason.find("LOADING Redis is loading the dataset in memory") != std::string::npos)
```
qiluo-msft, line 131: >LOADING [](http://example.com/codeflow?start=25&length=7)

ex.what() is for human reading, and may change in future. So find logic is not robust.

Could we just check if command is "PING"?

liuh-80: There may be other exception throws when PING command.
We only want to warning any exception caused by loading dataset. for other exception, it's better to still write error, which will help developer debug issue.

qiluo-msft: Since ex.what() may change in future, could you write a testcase to catch future regression?

liuh-80: Will create sonic-mgmt test case

tests/test_redis_loading.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # wait for redis load dataset finish
+    wait_redis_ready()
+
+    # cleanup
+    os.system("redis-cli FLUSHALL")
+
+
+def test_redis_loading_exception(loading_dataset, capfd):
+    # write swss log to stderr, so capfd can capture it
```
liuh-80, line 64: Add comments here and pass when faliled

liuh-80: Fixed

common/redisreply.h
liuh-80: Fixed, add a test case to reproduce the issue.
Also, according to the discussion, add comments in the test case to ask developer manually to run the test case.

common/redisreply.h
```diff
@@ -114,5 +116,30 @@ class RedisReply
...
+    try
+    {
+        func();
+    }
+    catch (const std::system_error& ex)
+    {
+        // Combine more error message and throw again
+        std::string reason = ex.what();
+        std::string errmsg = "RedisReply catches system_error: command: " + std::string(command) + ", reason: " + reason;
+        if (reason.rfind("LOADING ", 0) == 0)
```
qiluo-msft, line 131: >0 [](http://example.com/codeflow?start=43&length=1)

Do you want to check npos instead? ref: [https://cplusplus.com/reference/string/string/rfind/](https://cplusplus.com/reference/string/string/rfind/)

liuh-80: No, the code here is check start with 'LOADING'. c++ does not have 'startwith' method.
If compare result with npos, then if a command text contains 'LOADING' the check here also pass.

liuh-80: This code is correct, add comments to code

PR #987, opened 2025-03-14
[strict yang] Add YANG enforcemnet for CONFIG_DB tables set ops

wen587:
Depends on https://github.com/sonic-net/sonic-buildimage/pull/21977 and https://github.com/sonic-net/sonic-swss-common/pull/984
#### Why I did it
Add YANG enforcement when update CONFIG_DB tables. 
#### How I did it
Add YANG tables check when set| mod ops in CONFIG_DB
##### Work item tracking
- Microsoft ADO: 

#### How to verify it
python unit test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/tests/test_redis_ut.py
+++ b/tests/test_redis_ut.py
@@ -375,16 +375,16 @@ def test_ConfigDBConnector():
     assert config_db.db_name == "CONFIG_DB"
     assert config_db.TABLE_NAME_SEPARATOR == "|"
     config_db.get_redis_client(config_db.CONFIG_DB).flushdb()
-    config_db.set_entry("TEST_PORT", "Ethernet111", {"alias": "etp1x"})
+    config_db.set_entry("PORT", "Ethernet111", {"alias": "etp1x"})
     allconfig = config_db.get_config()
-    assert allconfig["TEST_PORT"]["Ethernet111"]["alias"] == "etp1x"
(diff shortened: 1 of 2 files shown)


tests/test_redis_ut.py
```diff
@@ -375,16 +375,16 @@ def test_ConfigDBConnector():
     assert config_db.db_name == "CONFIG_DB"
     assert config_db.TABLE_NAME_SEPARATOR == "|"
     config_db.get_redis_client(config_db.CONFIG_DB).flushdb()
-    config_db.set_entry("TEST_PORT", "Ethernet111", {"alias": "etp1x"})
```
qiluo-msft, line 378: >TEST_PORT [](http://example.com/codeflow?start=25&length=9)

You only has happy testcase, could you also test original TEST_PORT in this testcase, and the expectation is blocked?

wen587: https://github.com/sonic-net/sonic-swss-common/pull/987/files#diff-e95e67eabf12f4e97181558babbde4d4a1d320173950633a6534fd0a7ca72999R554-R560

Already has negative case.

qiluo-msft (approved):

PR #989, opened 2025-03-17
[swss-common] Enhanced NotificationConsumer queue to avoid memory lea…

yongxin-acctoncn:
…ks in some cases

Why I did it

In the current notification mechanism, the NotificationConsumer stores notification messages in bulk to an internal queue. However, when the various Orch consumers receive notifications, they only retrieve one notification from the queue for processing at a time. This approach has a potential risk of memory leaks. If the device is hit by a large number of FDB move or link up/down flapping events, a significant increase in memory consumption by the orchagent can be observed, eventually leading to a system crash or hang.
(Testing has been done on SONiC 202111, where the orchagent's memory consumption reached up to 10GB and continued to increase)

How I did it

Modified the internal queue of the NotificationConsumer class to replace std::queue. Now, when processing existing duplicate messages, the NotificationConsumer will no longer enqueue them and will instead move the existing message to the end of the queue. This significantly reduces the number of duplicate notify messages, and the time complexity for both lookup and insertion in the new queue is O(1).

How I verify it

1. Performed 10,000,000 link flapping tests on SONiC 202111, and the virtual memory usage of orchagent was about 700M~900M.

2. Performed 7000 fdb move per second on SONiC 202111, and the virtual memory usage of orchagent was about 397M.

swss also has an enhancement for this PR:
https://github.com/sonic-net/sonic-swss/pull/3560


common/notificationconsumer.h
```diff
@@ -18,6 +20,54 @@ namespace swss {
...
+private:
+    std::list<T> dq;
+    std::unordered_map<T, typename std::list<T>::iterator> map;
+
+public:
+    void push(const T& value)
+    {
+        auto it = map.find(value);
+        if (it != map.end())
+            dq.erase(it->second);
```
anders-nexthop, line 35: so if the notification is a duplicate, instead of adding a second entry, we move it to the end? wouldn't it make more sense to leave the original at its place in the queue, and not enqueue the second one? is there a chance that the entry stalls indefinitely if duplicate notifications keep showing up and moving back to the end of the queue each time?

common/notificationconsumer.h
```diff
@@ -18,6 +20,54 @@ namespace swss {
...
+        if (dq.empty())
+            return;
+        map.erase(dq.front());
+        dq.pop_front();
+    }
+
+    T front() const
+    {
+        if (dq.empty())
+            throw std::runtime_error("Queue is empty");
```
anders-nexthop, line 51: can't we let dq.front() propogate its error? I don't think we need to change the existing behavior.

common/notificationconsumer.h
```diff
@@ -55,7 +105,7 @@ class NotificationConsumer : public Selectable
     swss::DBConnector *m_db;
     swss::DBConnector *m_subscribe;
     std::string m_channel;
-    std::queue<std::string> m_queue;
+    Queue<std::string> m_queue;
```
anders-nexthop, line 108: This is a widely used class. How can we be certain that some users aren't relying on getting duplicate notifications? maybe this feature needs to be opt-in.

common/notificationconsumer.h
yongxin-acctoncn: 1. so if the notification is a duplicate, instead of adding a second entry, we move it to the end?
Yes.

2. wouldn't it make more sense to leave the original at its place in the queue, and not enqueue the second one?
Consider the scenario of link flapping. Suppose the received port status notification sequence is: (1) up → (2) down → (3) up. If we retain the original notification and discard new duplicate messages, the notification sequence delivered to orchagent will become (1) up → (2) down, which will trigger incorrect behaviors in upper-layer modules.

3. is there a chance that the entry stalls indefinitely if duplicate notifications keep showing up and moving back to the end of the queue each time?
The purpose of this modification is to deduplicate messages by filtering out duplicates during the message enqueue process. Since redundant messages are eliminated, the total number of messages in the queue is theoretically capped. As long as the consumer orchagent processes messages normally, all queued messages will eventually be handled completely. The core difference between the original design and this modification lies in that the deduplication mechanism turns infinite duplicate messages into a limited set of unique messages.

common/notificationconsumer.h
yongxin-acctoncn: If upper services rely on duplicate message reception like counting logic, issues will occur, which is a scenario limitation. This NotificationConsumer modification aims to resolve orchagent memory issues. Maybe further consideration is needed to define proper conditions for opt-in enabling of this feature.

common/notificationconsumer.h
yongxin-acctoncn: It seems that the behavior of front() is undefined when std::list is empty... Perhaps it's safer to write it this way.

common/notificationconsumer.h
anders-nexthop: > 1. so if the notification is a duplicate, instead of adding a second entry, we move it to the end?
>    Yes.
> 2. wouldn't it make more sense to leave the original at its place in the queue, and not enqueue the second one?
>    Consider the scenario of link flapping. Suppose the received port status notification sequence is: (1) up → (2) down → (3) up. If we retain the original notification and discard new duplicate messages, the notification sequence delivered to orchagent will become (1) up → (2) down, which will trigger incorrect behaviors in upper-layer modules.

Ok, so since we want the most recent update to take precedence we move it to the end, makes sense. Is there any issue with the consumer seeing idempotent notifications (for instance, link is "up" and we get a notification that link is "up")? I would expect this case to be handled, but probably worth checking.

> 3. is there a chance that the entry stalls indefinitely if duplicate notifications keep showing up and moving back to the end of the queue each time?
>    The purpose of this modification is to deduplicate messages by filtering out duplicates during the message enqueue process. Since redundant messages are eliminated, the total number of messages in the queue is theoretically capped. As long as the consumer orchagent processes messages normally, all queued messages will eventually be handled completely. The core difference between the original design and this modification lies in that the deduplication mechanism turns infinite duplicate messages into a limited set of unique messages.

The degenerate case would be:

- consumer can handle N messages in a set amount of time T
- there is one message M in the queue (at the front)
- N new messages come in, plus one more message that is a duplicate M'. Message M is dropped and message M' is pushed to the back of the queue at index N+1
- repeat

If this exact sequence continues, the update M will never be seen. So I think there's a possibility of starvation here, but if we assume that orchagent will eventually drain the queue and quiesce then this is less of a concern. I guess the place this could come up is during bursts, notifications about a flapping port could be delayed. Not sure how critical that would be though.

common/notificationconsumer.h
anders-nexthop: We could add an optional constructor arg for this, and change the caller to pass it where relevant. Or we could look through the existing users and see if there are any that are relying on receiving duplicates. There might be a lot of callers though.

common/notificationconsumer.h
anders-nexthop: ah. this is ok then. It's not really necessary, since this is an internal class and presumably it already respected the invariant, but that nobody will ever see this unexpectedly unless someone makes a breaking change to this code later on.

common/notificationconsumer.h
yongxin-acctoncn: Hmm... This logic lies in the middle of the notification message chain. Theoretically, the ASIC, as the message publisher, should not deliver duplicate idempotent notifications, while the orchs need to properly handle such cases. From another angle, a burst of repeated notifications can be considered a type of DoS attack. Prioritizing system stability to prevent out-of-memory crashes matters most. I think the current design makes it difficult to balance notification processing accuracy and system stability under some extreme scenarios ~

anders-nexthop: I'm trying to think if there is a realistic scenario where we could hit a case like this. I agree that system stability is crucial, but it helps if we can identify the tradeoffs and either document them or find ways to work around them.

Lets focus on link flaps. Say we generate a storm of events, more than the consumer can keep up with, and the notification queue grows. Among the other events, one thing happening is a link flapping continuously. Even if we assume that the system eventually settles and the queue gets emptied, there are still issues we could hit. If the queue is large enough, we may coalesce the link up/down events many times. One issue is that since the event keeps getting moved to the end of the queue, we might not actually read it until towards the end of the storm, so the notification will occur later than the event occurred, and later than we would have detected it under the existing implementation. Another issue is that we lose the information about the link flaps that got coalesced. If the link flapped 10 times in quick succession, we may only detect that as a single event. This is compounded by the fact that Redis does not store any history; there is no way to "go back" and read through the event chain if you want the full story. So adding this change means the NotificationQueue is not usable if you are trying to detect fine-grained information about the event stream, such as link flaps.

anders-nexthop:
Is the queue size bounded anywhere? It seems like the issue is that the queue can grow infinitely large. We can't really bound the queue without dropping notifications (we could implement a pushback mechanism that would push this onto the producer, but that's out of scope for the change you're considering here - that might be the best actual fix though). We can't drop notifications because we could end up with invalid state (if we see a link down but drop the corresponding link up, link will be marked down even though it's actually fine).

We could incorporate bounds and coalescing together though. Start with normal behavior (no coalescing) and then once the queue size hits a certain threshold we could trigger coalescing? There would be some corner cases to think through, but we can emit a log message when we start/stop, which is at least _some_ signal that can be detected to indicate that we have started coalescing. The problem with this approach is that the scenario in which you want to detect individual events (link flap storm, for instance) is the same scenario in which you want coalescing. So maybe this wouldn't gain us much.

Maybe we could add some Notification metadata that would track when an entry was coalesced? Like have a count and maybe an initial timestamp that gets passed along with the Notification. So when a consumer receives a coalesced Notification they can check the timestamps and infer that the whatever state the Notification contains was triggered X times over a given time period. That's not perfect, but it would probably be enough information to cover most cases. For instance, if after an event storm we receive two coalesced link up/down notifications, we can check the timestamps and the counts and see if they match up, and assume that they are paired events, so we could at least say "we say roughly N link flaps over this specific time period". Any consumer who doesn't care about coalescing would just ignore those fields.

yongxin-acctoncn:
> Is the queue size bounded anywhere? It seems like the issue is that the queue can grow infinitely large. We can't really bound the queue without dropping notifications (we could implement a pushback mechanism that would push this onto the producer, but that's out of scope for the change you're considering here - that might be the best actual fix though). We can't drop notifications because we could end up with invalid state (if we see a link down but drop the corresponding link up, link will be marked down even though it's actually fine).
> 
> We could incorporate bounds and coalescing together though. Start with normal behavior (no coalescing) and then once the queue size hits a certain threshold we could trigger coalescing? There would be some corner cases to think through, but we can emit a log message when we start/stop, which is at least _some_ signal that can be detected to indicate that we have started coalescing. The problem with this approach is that the scenario in which you want to detect individual events (link flap storm, for instance) is the same scenario in which you want coalescing. So maybe this wouldn't gain us much.
> 
> Maybe we could add some Notification metadata that would track when an entry was coalesced? Like have a count and maybe an initial timestamp that gets passed along with the Notification. So when a consumer receives a coalesced Notification they can check the timestamps and infer that the whatever state the Notification contains was triggered X times over a given time period. That's not perfect, but it would probably be enough information to cover most cases. For instance, if after an event storm we receive two coalesced link up/down notifications, we can check the timestamps and the counts and see if they match up, and assume that they are paired events, so we could at least say "we say roughly N link flaps over this specific time period". Any consumer who doesn't care about coalescing would just ignore those fields.

No specific queue size bound exists in the NotificationConsumer implementation. It only defines a POP_BATCH_SIZE macro to throttle notifications sent to the upper layer, limiting the upper layer to processing POP_BATCH_SIZE notifications per-select-run.

Maybe we could try incorporating bounds and coalescing together?
We can start with normal behavior (no coalescing), and once the queue size hits a certain threshold (POP_BATCH_SIZE), switch to coalescing mode.
We can log the mode switching event to syslog and start attaching some metadata to the coalesced notifications. I'm not sure whether it's possible to do this without affecting the existing notification parsing interface in the upper layers.

PR #996, opened 2025-03-29
Fix dependency issues coming up with docker-sonic-vs build

saiarcot895:
Because of sonic-net/sonic-buildimage#20786, the framework package is installed into docker-sonic-vs. This package depends on libswsscommon, which means that the initial removal of libswsscommon (which is done for Docker purposes) will cause a dependency issue.

Tell dpkg to ignore that dependency issue and proceed.


zjswhhh (approved):
~~We can force merge this PR. There is a keyword to include this pending commit in pipeline runs, but I don't recall lol.~~ 

Synced offline. Might just revert https://github.com/sonic-net/sonic-buildimage/pull/20713 as a fix.

PR #1009, opened 2025-04-21
Adding a table to store peer info if peer creation is successful

sudarshankumar4893:
Adding a table to store peer info if peer creation is successful. This document explains the need for this table and the data that will go into it:
https://github.com/sonic-net/SONiC/pull/1963/files


anish-n (approved):

prsunny (approved):

prsunny:
@qiluo-msft , help with merge

PR #1018, opened 2025-05-19
[build] upgrade ubuntu 20.04 to 24.04

liushilongbuaa:


prabhataravind:
@liushilongbuaa test-docker-sonic-vs-template.yml  also needs to be updated.. Currently vstests are pulling 20.04 artifacts and it fails.

prabhataravind:
@prsunny @theasianpianist  for viz

azure-pipelines.yml
```diff
@@ -54,6 +54,7 @@ stages:
   jobs:
   - job:
     displayName: "amd64/ubuntu-20.04"
```
qiluo-msft, line 56: will you also change it to ubuntu-22.04 ?

liushilongbuaa: we already have ubuntu-22.04 at line 93.
In future this job will change to 24.04

qiluo-msft (approved):

prabhataravind:
@liushilongbuaa, it appears there are some dependency issues with the version of certain packages. @saiarcot895 for viz.

prsunny (approved):

PR #1039, opened 2025-07-11
ZMQ lib change to support one-to-one sync.

divyagayathri-hcl:
### What I did
Introduced one-to-one synchronization option (request/replay) in the ZMQ client and server implementation. This is a new addition to the existing push/pull model.

### Why I did it
To ensure reliable and faster message delivery and synchronization between ZMQ producer and consumer in scenarios where message acknowledgment or ordering is critical. This will be used only by P4RT and P4Orch for now.

### How I verified it
Ran unit tests in zmq_state_ut.cpp and confirmed a successful message exchange using the updated one-to-one sync logic.
Verified zmq_recv and zmq_send retry logic operates correctly using debug logs.

### Details if related
None.

### DIFF
--- a/common/zmqclient.h
+++ b/common/zmqclient.h
@@ -15,6 +15,9 @@ public:
 
     ZmqClient(const std::string& endpoint);
     ZmqClient(const std::string& endpoint, const std::string& vrf);
+    // If waitTimeMs is set to non-zero, it will enable one-to-one sync with the
+    // server. It will use ZMQ_REQ and ZMQ_REP socket type. There can only be
+    // one client and one server for a ZMQ socket.
     ZmqClient(const std::string& endpoint, uint32_t waitTimeMs);
     ~ZmqClient();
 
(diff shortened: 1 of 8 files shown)


mint570 (dismissed):

common/zmqproducerstatetable.h
```diff
@@ -16,33 +16,34 @@ namespace swss {
 class ZmqProducerStateTable : public ProducerStateTable
 {
 public:
-    ZmqProducerStateTable(DBConnector *db, const std::string &tableName, ZmqClient &zmqClient, bool dbPersistence = true);
```
mint570: Many of the lines are no update. Don't make changes if there are no difference.

divyagayathri-hcl: It was due to the clan-formatting. I have changed it now.

common/zmqclient.h
```diff
@@ -14,8 +14,11 @@ class ZmqClient
 public:
 
     ZmqClient(const std::string& endpoint);
-    ZmqClient(const std::string& endpoint, const std::string& vrf);
-    ZmqClient(const std::string& endpoint, uint32_t waitTimeMs);
+    ZmqClient(const std::string& endpoint, const std::string& vrf = "");
```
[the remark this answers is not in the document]
divyagayathri-hcl: Addressed the change.

common/zmqclient.h
```diff
@@ -43,8 +47,12 @@ class ZmqClient
 
     bool m_connected;
 
+    // This method should only be used in one-to-one sync mode with the server.
```
mint570: Delete this line

common/zmqclient.h
```diff
@@ -43,8 +47,12 @@ class ZmqClient
...
+    // If this is set to zero, one-to-one sync is disabled.
     uint32_t m_waitTimeMs;
 
+    bool m_oneToOneSync;
```
mint570: Can you default this to false here? So we don't need to set it in all the constructors.

mint570 (approved):

PR #1041, opened 2025-07-15
Ignore nonexistent include in database_global.json

yue-fred-gao:
### why
SonicDBConfig::initializeGlobalConfig aborts parsing global config if it encounters an include of a database config that doesn't exist. The global config file includes all possible includes in the switch. However, not all includes are relevant to a client. For example, in smartswitch, dash-ha container only cares NPU databases and databases of the DPU managed by the container. The current behavior requires a dash-ha container mounting all the DPU database instances.

### what this PR does
1. Add ignore_nonexistent flag, with default value false, to function parseDatabaseConfig. If it is set to true, it will not throw error if the included database config file doesn't exist.
2. initializeGlobalConfig sets ignore_nonexistent to true when calling parseDatabaseConfig. If inst_entry, db_entry and separator_entry are empty, it will ignore the key.


Pterosaur (approved):

zjswhhh (approved):

common/dbconnector.cpp
```diff
@@ -128,10 +134,16 @@ void SonicDBConfig::initializeGlobalConfig(const string &file)
                     continue;
                 }
 
-                parseDatabaseConfig(local_file, inst_entry, db_entry, separator_entry);
-                m_inst_info[key] = inst_entry;
-                m_db_info[key] = db_entry;
-                m_db_separator[key] = separator_entry;
+                parseDatabaseConfig(local_file, inst_entry, db_entry, separator_entry, true);
+                // all the entries are empty, then don't add them to the map. It may happen if the included
+                // config doesn't exist. For example, dash-ha container will only mount the redis instance
```
qiluo-msft, line 139: >dash-ha container will only mount the redis instance [](http://example.com/codeflow?start=54&length=52)

This should be considered as a config file issue instead of implementation issue. We'd better not to assume "The global config file includes all possible includes in the switch."

Thw swss-common deb package just provide a sample database_config.json which may be useful for most of the use cases, however the installer environment need to ensure the real database config files valid, and we should not relax the implementation on parsing errors.

Pterosaur (dismissed):

common/dbconnector.cpp
yue-fred-gao: This is not about database_config.json but database_global.json. The latter includes multiple  database_config.json and not all of them are accessible to the container. This does not assume global config includes all database_config. It works if global config only include a subset of all possible includes. 

The PR only ignores the error if the included file is not found for container case. It provides INFO log for troubleshooting, which should be sufficient  if it is a legitimate config issue.

Alternatively, we can have multiple database_global.json, one per container. And we need a naming scheme for containers to find their corresponding database_global.json. This may require a bit of redesign in database.

@Pterosaur , please share your thoughts.

qiluo-msft: got it, thanks!

crates/swss-common/src/lib.rs
```diff
@@ -13,7 +13,7 @@ pub fn sonic_db_config_initialize(path: &str) -> Result<(), Exception> {
 }
 
 /// Rust wrapper around `swss::SonicDBConfig::initializeGlobalConfig`.
-pub fn sonic_db_config_initialize_global(path: &str) -> Result<(), Exception> {
+pub fn sonic_db_config_initialize_global(path: &str, ignore_nonexistent: bool) -> Result<(), Exception> {
```
qiluo-msft, line 16: >sonic_db_config_initialize_global [](http://example.com/codeflow?start=7&length=33)

This will break existing Rust application using this function. Since Rust does not support default parameter value like Python/C++, does not support function overloading. There are 2 options:

1. keep original function unchanged, and add anotehr function like sonic_db_config_initialize_global_ignore_nonexistent usiing more parameters.
2. treat it as breaking change, and update all the Rust applications.

yue-fred-gao: hi @qiluo-msft, I plan to commit the change in dash-ha at the same time. When we update sonic-buildimage, it needs to go in at the same time as this. Are there other rust applications using this function? If there is and hard to coordinate, I can implement 1 to break the dependency.

qiluo-msft: As far as I know, only dash-has is using it.

Pterosaur: > As far as I know, only dash-ha is using it.

Hi @qiluo-msft , I'm using this rest API as well in the swss.
https://github.com/sonic-net/sonic-swss/pull/3796

yue-fred-gao: @Pterosaur , is it in the committed code that needs to be updated at the same time?

PR #1051, opened 2025-07-18
SHlorch changes for EVPN MH feature

jayakj:
What I did
Implementation of SHL orch from EVPN VXLAN Multihoming HLD.
https://github.com/sonic-net/SONiC/blob/master/doc/vxlan/EVPN/EVPN_VxLAN_Multihoming.md#3310-ShlOrch

Why I did it
This is part of EVPN Multi homing feature

How I verified it
By SWSS UT

Details if related
corresponding swss PR: https://github.com/sonic-net/sonic-swss/pull/3769


prsunny:
@qiluo-msft , please merge to unblock other dependent PRs

PR #1068, opened 2025-08-18
Add lazy binding support to ZmqServer.

liuh-80:
Add lazy binding support to ZmqServer.

#### Why I did it
When creating a ZmqServer followed by a ZmqProducerStateTable, there may be a time gap between the server starting to receive messages and the producer state table registering its handler. 

This gap can lead to dropped messages. To avoid this, use lazy binding and invoke bind() only after the handler is registered.

#### How I did it
Add lazy binding support to ZmqServer.

##### Work item tracking
- Microsoft ADO: 33995986

#### How to verify it
Pass all test cases.
Add new test case for lazy binding.

Pass all sonic-mgmt test with this PR:
https://github.com/sonic-net/sonic-buildimage/pull/23741

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Add lazy binding support to ZmqServer.

### DIFF
--- a/tests/zmq_state_ut.cpp
+++ b/tests/zmq_state_ut.cpp
@@ -582,13 +582,12 @@ TEST(ZmqServerLazzyBind, test)
     auto testKey = "testkey";
     kcos.push_back(KeyOpFieldsValuesTuple{testKey, SET_COMMAND, std::vector<FieldValueTuple>{}});
     std::vector<std::shared_ptr<KeyOpFieldsValuesTuple>> kcosPtr;
-    std::string dbName, tableName;
     p.send(kcos);
 
     // initialize ZMQ server with lazzy bind
     DBConnector server_db(TEST_DB, 0, true);
     ZmqServer server(pullEndpoint, "", true);
-    ZmqConsumerStateTable c(&db, tableName, server, 128, 0, false);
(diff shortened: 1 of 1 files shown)


common/zmqserver.cpp
```diff
@@ -13,34 +13,57 @@ using namespace std;
...
+    }
 }
 
-void ZmqServer::connect()
+void ZmqServer::bind()
 {
     SWSS_LOG_ENTER();
+    if (m_socket)
+    {
+        SWSS_LOG_THROW("ZmqServer has already been bound to the endpoint: %s", m_endpoint.c_str());
```
r12f, line 64: do we need to consider this bind as no-op instead of throw?

liuh-80: A ZMQ server should only be bound once.
To catch potential issues during development, I recommend throwing an exception if a second bind attempt is detected. This will help identify incorrect usage early and ensure proper lifecycle management of the server.

r12f: gotcha! so it is an explicit lazy binding instead of implicit one. makes sense!

r12f (approved):

PR #1069, opened 2025-08-19
Fix trixie compilation and add Trixie to PR pipeline

saiarcot895:
Fix a new compilation error on Trixie that popped up. Also, add a build for Trixie to the PR pipeline to prevent future errors.


.azure-pipelines/build-sairedis-template.yml
```diff
@@ -126,7 +126,7 @@ jobs:
       set -ex
       rm ../*.deb || true
       ./autogen.sh
-      DEB_BUILD_OPTIONS=nocheck DEB_BUILD_PROFILES=nopython2 fakeroot dpkg-buildpackage -b -us -uc -Psyncd,vs,nopython2 -j$(nproc)
```
qiluo-msft, line 129: >DEB_BUILD_PROFILES [](http://example.com/codeflow?start=32&length=18)

It is used in debian/rules

qiluo-msft (approved):

PR #1077, opened 2025-09-10
Relocate sonicdb and sonicdb-derive to swss-common

yue-fred-gao:
### why
currently, sonic-dash-ha adds sonic-dash-api as submodule to compile protos into rust structs. At the same time, both are submodules in sonic-buildimage. This creates unnecessary dependency and complexity. The solution is moving sonic-dash-api-proto crate into sonic-dash-api. sonic-dash-ha adds sonic-dash-api-proto as a dependent crate. At build time, it pulls the crate through git and the latter compiles protos on demand. This is similar to swss-common crate to sonic-dash-ha. However, sonic-dash-api-proto depends on sonicdb-derive and sonic-common crates in sonic-dash-ha. To break the inter-dependency, we need to move sonicdb-derive and the SonicDBTable in sonic-common, which is added to a new crate called sonicdb to another place. swss-common is the most natural location because it is a derive macro for SONiC db and it potentially can be used by other components in sonic.

### what this PR does
Move sonicdb-derive from sonic-dash-ha to swss-common. 
Move SonicDBTable that was previously in sonic-common crate in sonic-dash-ha to the sonicdb crate in swss-common.
NOTE: rust currently requires derive macro to be in its own crate so we have to use 2 crates.

### dependent PRs
https://github.com/sonic-net/sonic-dash-api/pull/45
https://github.com/sonic-net/sonic-dash-ha/pull/109

### DIFF
--- a/crates/sonicdb-derive/src/lib.rs
+++ b/crates/sonicdb-derive/src/lib.rs
@@ -118,7 +118,7 @@ pub fn serde_sonicdb_derive(input: TokenStream) -> TokenStream {
     let is_dpu_value = is_dpu;
 
     let expanded = quote! {
-        impl sonic_common::SonicDbTable for #struct_name {
+        impl sonicdb::SonicDbTable for #struct_name {
             fn key_separator() -> char {
                 #key_separator
             }


crates/sonicdb/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+use swss_common::KeyOpFieldValues;
+
+/// Trait for objects that can be stored in a Sonic DB table.
+pub trait SonicDbTable {
```
qiluo-msft, line 4: SonicDbTable seems a combination of SonicDBInfo and Table. Is this really necessary?

zjswhhh:
Hi @qiluo-msft - can you help review again please?

KrisNey-MSFT:
hi @yue-fred-gao - can you let us know where this needs to go from here?  TY!

PR #1081, opened 2025-09-24
Prevent adding new CFG table since it is auto generated

wen587:
Add a checker to prevent adding new CFG_ macros when update common/schema.h.
#### Why I did it
CFG_XXX_TABLE_NAME macros should be auto-generated from YANG models via cfg_schema.h.
#### How I did it
Introduced .github/workflows/prevent-cfg.yml which runs on PRs that modify common/schema.h.
It extracts all CFG_ macros defined in the file and compares them against an allowed legacy list.
##### Work item tracking
- Microsoft ADO: 35075048

#### How to verify it
Manual test CFG change and other change in common/schema.h

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/.github/workflows/prevent-cfg.yml
+++ b/.github/workflows/prevent-cfg.yml
@@ -100,8 +100,8 @@ jobs:
           allowed_dict["$macro"]=1
         done
 
-        # Find all CFG_ macros in the current schema.h
-        current_cfg_macros=$(grep -o 'CFG_[A-Za-z_]*' common/schema.h | sort | uniq)
+        # Find all CFG_ macros in the current schema.h (only from #define statements)
+        current_cfg_macros=$(grep '#define CFG_' common/schema.h | grep -o 'CFG_[A-Za-z0-9_]*' | sort | uniq)
 
         # Check if any new CFG_ macros have been added
         new_macros=""
--- a/.github/workflows/prevent-cfg.yml
+++ b/.github/workflows/prevent-cfg.yml
@@ -114,12 +114,10 @@ jobs:
         if [ ! -z "$new_macros" ]; then
           echo "ERROR: New CFG_ macros detected in common/schema.h:"
           echo -e "$new_macros"
-          echo "CFG_ macros should NOT be added directly to schema.h."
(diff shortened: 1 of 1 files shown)


.github/workflows/prevent-cfg.yml
```diff
@@ -0,0 +1,124 @@
...
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+
+    - name: Check for new CFG_ macros in schema.h
+      run: |
+        echo "Checking for new CFG_ macro additions in common/schema.h..."
+
+        # Allowed legacy CFG_ macros that should remain in schema.h
```
vaibhavhd, line 29: Aren't you also planning for the legacy macros to be slowly phased out from schema.h? This comment makes it look like the legacy ones are going to be forever live in schema.h

.github/workflows/prevent-cfg.yml
```diff
@@ -0,0 +1,124 @@
...
+        )
+
+        # Create associative array for fast lookup
+        declare -A allowed_dict
+        for macro in "${ALLOWED_CFG_MACROS[@]}"; do
+          allowed_dict["$macro"]=1
+        done
+
+        # Find all CFG_ macros in the current schema.h (only from #define statements)
+        current_cfg_macros=$(grep '#define CFG_' common/schema.h | grep -o 'CFG_[A-Za-z0-9_]*' | sort | uniq)
```
vaibhavhd, line 104: I have a question about this regex - do we enforce that all macros are defined by starting string `CFG_`?

vaibhavhd (approved):

.github/workflows/prevent-cfg.yml
wen587: Yes. CFG_XXX means it is Config_DB related which is covered by YANG

.github/workflows/prevent-cfg.yml
wen587: Yes. Legacy macros are kept by now. And will be removing it after yang model created or format being aligned in other repo.

PR #1084, opened 2025-09-30
Add pop batch size support for ZMQ Consumer

vivekrnv:
**What i did**

Add pop batch size support to ZmqConsumerState Table to optimize memory and increase the speed for updating CRM counters/DASH Feedback when applying dash configuration at scale

Related PR's:
https://github.com/sonic-net/sonic-swss/pull/3910
https://github.com/sonic-net/sonic-buildimage/pull/24168
https://github.com/sonic-net/sonic-sairedis/pull/1660

Example:

Let's say we have a GNMI server which pushed X entries to orchagent. Current logic of ZmqConsumerState table would move X entries to m_toSync map. 

Dashorch would create X entries in bulker. However, max_bulk size is often limited (currently 1000) And definitely much less than the size of m_toSync in this scale scenario. 

So, effective memory during this time is `2 * X (1 copy in m_toSync + 1 copy in bulker)* size per object `until all those entries are applied to ASIC.

1. With this change, only pop batch size entries are popped out to m_toSync and added to bulker. Thus peak memory utilization is cut in half in case of Dash Scale.

2. Another side effect of this change is the postprocessing for pop batch size  items is done immediately in orchagent and there is no delay on updating CRM or GNMI Feedback loop. If not, post processing starts only after all the entries in m_toSync are applied to syncd which is not capped for current design

**How i verified**

UT and applying DASH config and making sure everything works

Before the update:
```
[ RUN      ] ZmqConsumerStateTablePopSize.test
Consumer thread started
Entering select
Producer sent 150 elements
pops: 150
Consumer thread joined
tests/zmq_state_ut.cpp:636: Failure
Expected equality of these values:
 popCount
   Which is: 1
 4
popCount: 1, expected: 4
tests/zmq_state_ut.cpp:639: Failure
Expected equality of these values:
 recvdSizes[i]
   Which is: 150
 expectedSizes[i]
   Which is: 40
recvdSizes[0]: 150, expected: 40
[  FAILED ] ZmqConsumerStateTablePopSize.test (16017 ms)
```

After the update:
```
[----------] 1 test from ZmqConsumerStateTablePopSize
[ RUN      ] ZmqConsumerStateTablePopSize.test
Consumer thread started
Entering select
Producer sent 150 elements
pops: 40
Entering select
pops: 40
Entering select
pops: 40
Entering select
pops: 30
Consumer thread joined
[       OK ] ZmqConsumerStateTablePopSize.test (2008 ms)
[----------] 1 test from ZmqConsumerStateTablePopSize (2008 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (2012 ms total)
[  PASSED  ] 1 test.
```


oleksandrivantsiv (approved):

prsunny:
@qiluo-msft , please review/merge.

common/zmqconsumerstatetable.cpp
```diff
@@ -80,7 +81,7 @@ void ZmqConsumerStateTable::pops(std::deque<KeyOpFieldsValuesTuple> &vkco, const
     }
 
     vkco.clear();
-    for (size_t ie = 0; ie < count; ie++)
+    for (size_t ie = 0; ie < min(count, m_popBatchSize); ie++)
```
qiluo-msft, line 84: The min(count, m_popBatchSize) calculation happens on every pops() call. Could you move it out of the for loop?

qiluo-msft (approved):

PR #1093, opened 2025-10-23
Add L2MC Table

jasper-hou-micas:


common/debugframework.h
```diff
@@ -0,0 +1,118 @@
...
+    void setPostAction(const std::string &p);
+
+    std::vector<std::string> m_registeredComps;
+    std::map<std::string, std::fstream*> m_compFds;
+    std::unique_ptr<Table>  m_configParams;
+
+    TargetLocation     m_targetLocation;
+    PostAction         m_postAction;
+
+#ifndef NO_RET_TYPE
```
aidan-gallagher: **Medium:** `NO_RET_TYPE` is defined with `#define` inside the class body, but preprocessor macros don't respect C++ scope. This macro leaks into the global namespace for any file that includes this header.

pyext/swsscommon.i
```diff
@@ -361,3 +362,4 @@ T castSelectableObj(swss::Selectable *temp)
 %include "status_code_util.h"
 #include "redis_table_waiter.h"
 %include "restart_waiter.h"
+%include "debugframework.h"
```
aidan-gallagher: **Medium:** Exposing `DebugFramework` to Python via SWIG is problematic: `DumpInfoFunc` is a C function pointer typedef that can't be naturally constructed from Python, `dumpWrite` uses C varargs, and the framework spawns detached threads which are incompatible with Python's GIL.

common/debugframework.h
```diff
@@ -0,0 +1,118 @@
...
+#include <map>
+#include <thread>
+#include <assert.h>
+#include <execinfo.h>
+#include "dbconnector.h"
+#include "table.h"
+
+using namespace swss;
+
+#if 0
```
aidan-gallagher: **Medium:** This `#if 0` block (lines 13-23) and the one at lines 70-72, plus `debugframework.cpp:414-463`, contain dead code including a dangerous `#undef assert` macro and an `fopen` without null check. These should be removed rather than left as dead code.

PR #1118, opened 2025-12-01
Revert "Fix risky unwrap(), expect(), and casting"

liushilongbuaa:
Reverts sonic-net/sonic-swss-common#1113

https://github.com/sonic-net/sonic-swss/blob/eae91a23c01bcc0b1e915ab46c5228a854a6e42e/Cargo.toml#L67
sonic-swss refers to sonic-swss-common's latest commit.
So changes in swss-common breaks SONiC's build.


qiluo-msft:
I will force merge to unblock sonic-buildimage PR build and master branch build. There is no bug in the reverted commit, and we will bring it back after fixing other Rust repos.

PR #1155, opened 2026-02-25
[EVPN-MH] Add L2 nexthop group table and raw netlink message handling support

bdfriedman:
#### Why I did it

This PR adds infrastructure support in sonic-swss-common for EVPN Multi-Homing (EVPN-MH) feature. EVPN-MH requires:
1. A new L2 nexthop group table to manage L2 nexthop groups in APPL_DB
2. Raw netlink message handling capability to process EVPN-MH specific netlink attributes that cannot be fully parsed by libnl's generic object parsing

##### Work item tracking
- Microsoft ADO **(number only)**:

#### How I did it

1. **Added new APPL_DB table**: `APP_L2_NEXTHOP_GROUP_TABLE_NAME` for managing L2 nexthop groups
2. **Implemented raw netlink message handler infrastructure** in NetDispatcher:
   - Added `registerRawMessageHandler()` to register callbacks for raw netlink messages
   - Added `unregisterRawMessageHandler()` to unregister raw message handlers
   - Added `onNetlinkMessageRaw()` to handle raw netlink messages (as `nlmsghdr*`)
   - Added `onMsgRaw()` virtual method in NetMsg class for derived classes to implement raw message processing
   - Added `m_rawhandlers` map to store raw message handlers
3. **Modified message flow**: When a netlink message is not handled by the regular (parsed) handlers, it falls back to raw message handlers

This allows components to choose between:
- Parsed netlink messages (existing behavior via `onMsg()` with `nl_object*`)
- Raw netlink messages (new capability via `onMsgRaw()` with `nlmsghdr*`)

#### How to verify it

1. Build sonic-swss-common with these changes
2. Verify that existing netlink message handling continues to work (backward compatibility)
3. Test EVPN-MH feature components that register raw message handlers
4. Verify L2_NEXTHOP_GROUP_TABLE can be accessed from APPL_DB

Add support for L2 nexthop group table and raw netlink message handling for EVPN-MH feature

#### Link to config_db schema for YANG module changes

N/A - This PR only adds APPL_DB table definition, no CONFIG_DB schema changes

---

**Key Changes:**
- Files modified: netdispatcher.cpp, netdispatcher.h, netmsg.h, schema.h
- Total changes: +60 lines, -1 line
- This is infrastructure preparation for EVPN Multi-Homing support in sonic-swss


common/netdispatcher.cpp
```diff
@@ -35,12 +35,49 @@ void NetDispatcher::unregisterMessageHandler(int nlmsg_type)
...
+}
+
 void NetDispatcher::nlCallback(struct nl_object *obj, void *context)
 {
     NetMsg *callback = (NetMsg *)context;
     callback->onMsg(nl_object_get_msgtype(obj), obj);
 }
 
+void NetDispatcher::onNetlinkMessageRaw(struct nl_msg *msg)
+{
```
banidoru, line 75: `m_rawhandlers` is accessed here without holding the mutex, but `registerRawMessageHandler`/`unregisterRawMessageHandler` both acquire `MUTEX`. If a handler is registered/unregistered concurrently with message dispatch, this is a data race.

For consistency with `getCallback()` (which locks), this method should also hold the mutex while looking up `m_rawhandlers`, or the lookup should be factored through a mutex-protected accessor like `getCallback` does for `m_handlers`.

common/netdispatcher.cpp
```diff
@@ -35,12 +35,49 @@ void NetDispatcher::unregisterMessageHandler(int nlmsg_type)
     m_handlers.erase(it);
 }
 
+void NetDispatcher::registerRawMessageHandler(int nlmsg_type, NetMsg *callback)
+{
```
banidoru, line 45: Typo in the exception message: "registered on already registerd" → should be "register an already registered". Also, the existing codebase throws string literals (`throw "..."`) which is not ideal (can't be caught by type), but since this matches the existing pattern in `registerMessageHandler`, it's consistent.

banidoru (commented):
- **Thread safety**: `onNetlinkMessageRaw` reads `m_rawhandlers` without holding the mutex, while `registerRawMessageHandler`/`unregisterRawMessageHandler` both lock it — concurrent register/unregister + message dispatch is a data race
- **Fallback design is clean**: unhandled parsed messages falling through to raw handlers is a reasonable approach
- **`onMsgRaw` default empty impl** in base class is fine for backward compat
- **Schema addition** (`APP_L2_NEXTHOP_GROUP_TABLE_NAME`) is straightforward

common/netdispatcher.cpp
```diff
@@ -61,7 +98,9 @@ void NetDispatcher::onNetlinkMessage(struct nl_msg *msg)
 
     /* Drop not registered messages */
     if (callback == nullptr)
+    {
```
banidoru, line 121: Design concern: raw handlers only fire when `getCallback()` returns nullptr (no parsed handler for this type). If a caller registers both a parsed handler and a raw handler for the same `nlmsg_type`, the raw handler silently never fires. This implicit fallback-only behavior should at minimum be documented. Consider adding a `SWSS_LOG_DEBUG` when falling through to raw, so it's observable.

banidoru (commented):
- Raw handlers only fire as fallback when no parsed handler exists for a message type — if both are registered for the same type, raw handler silently never fires. This implicit priority could surprise callers; worth a comment or log.
- No null-check on `callback` parameter in `registerRawMessageHandler` — storing nullptr leads to crash on dispatch.
- `onNetlinkMessageRaw` reads `m_rawhandlers` without holding the mutex (already noted by prior reviewer).

common/netdispatcher.cpp
```diff
@@ -35,12 +35,49 @@ void NetDispatcher::unregisterMessageHandler(int nlmsg_type)
     m_handlers.erase(it);
 }
 
+void NetDispatcher::registerRawMessageHandler(int nlmsg_type, NetMsg *callback)
+{
+    MUTEX;
+
+    if (m_rawhandlers.find(nlmsg_type) != m_rawhandlers.end())
```
banidoru: No null-check on `callback`. If `nullptr` is passed, it will be stored in `m_rawhandlers` and cause a segfault when `onNetlinkMessageRaw` dispatches to it. The existing `registerMessageHandler` has the same gap, but this is new code — consider adding `if (!callback) throw ...;` here.

common/netmsg.h
```diff
@@ -9,5 +9,10 @@ namespace swss
         public:
             /* Called by NetDispatcher when netmsg matches filters */
             virtual void onMsg(int nlmsg_type, struct nl_object *obj) = 0;
+
+            /* Called by NetDispatcher when raw msg is send for matches filters */
+            virtual void onMsgRaw(struct nlmsghdr *)
+            {
+            }
     };
```
banidoru, line 17: An empty default implementation means if a `NetMsg` subclass is registered via `registerRawMessageHandler` but forgets to override `onMsgRaw`, messages are silently dropped with no indication. Consider either:
1. Making this pure virtual (breaking change, less ideal), or
2. Adding a `SWSS_LOG_WARN` in the default body so silent drops are observable in logs.

banidoru (commented):
- **Data race** on `m_rawhandlers` in `onNetlinkMessageRaw` (no mutex) — already flagged, seconding this as the most critical issue
- **Silent no-op risk**: `onMsgRaw()` default empty body means a subclass registered as raw handler that forgets to override it silently drops all messages — consider making it pure virtual or adding a log warning
- **No unit tests** for the new raw handler registration/dispatch path
- **Missing ADO work item** in PR description

banidoru (approved):
All reviewers approved. LGTM.

common/netdispatcher.cpp
r12f: Hey Berry, this comment looks legit. @bdfriedman

banidoru: Fixed — `onNetlinkMessageRaw` now uses `getRawCallback()` which acquires the mutex, consistent with `getCallback()`. ✅

common/netdispatcher.cpp
banidoru: Fixed — typo corrected to "register an already registered". Null-check also added. ✅

common/netdispatcher.cpp
banidoru: Fixed — added code comment documenting fallback-only behavior and `SWSS_LOG_DEBUG` when falling through to raw handler. ✅

common/netdispatcher.cpp
banidoru: Fixed — null-check added to both `registerRawMessageHandler` and `registerMessageHandler`. ✅

common/netmsg.h
banidoru: Fixed — default `onMsgRaw` now logs `SWSS_LOG_WARN` with nlmsg_type, making silent drops observable. ✅

banidoru (approved):
All prior review concerns addressed in this iteration:
- Mutex protection for raw handler lookup: now uses `getRawCallback()` with proper locking
- Typo in exception message fixed
- Fallback-only behavior documented with code comment + `SWSS_LOG_DEBUG` on fallthrough
- Null-check added to both `registerMessageHandler` and `registerRawMessageHandler`
- Default `onMsgRaw` now emits `SWSS_LOG_WARN` instead of silently dropping

banidoru (approved):
All previous concerns have been addressed.

- The data race in  is fixed.
- The typo in the exception message is corrected.
- A log message is now present when falling back to a raw handler.
- Null checks for callbacks have been added.
- The default  implementation now logs a warning.

common/netdispatcher.cpp
banidoru: Verified fixed in the latest commit. ✅

common/netdispatcher.cpp
banidoru: Verified fixed in the latest commit. ✅

common/netdispatcher.cpp
banidoru: Verified fixed in the latest commit. ✅

common/netdispatcher.cpp
banidoru: Verified fixed in the latest commit. ✅

common/netmsg.h
banidoru: Verified fixed in the latest commit. ✅

common/netdispatcher.h
```diff
@@ -23,9 +23,29 @@ namespace swss
...
+
+
             /** Called by NetLink or FpmLink classes as indication of new packet arrival. */
             void onNetlinkMessage(struct nl_msg *msg);
 
+            /*
+             * Called by NetLink or FpmLink classes as indication of new packet arrival
+             */
+            void onNetlinkMessageRaw(struct nl_msg *msg);
+
```
banidoru: `onNetlinkMessageRaw` is declared here as public, but also declared again in the `private` section (around line 72). Duplicate member function declarations in the same class are ill-formed. Remove one of these — likely keep only the private one since this is an internal dispatch method that doesn't need public exposure.

banidoru (changes requested):
All 5 prior review concerns have been addressed — nice work.

One new issue found in this iteration:
- **Duplicate declaration**: `onNetlinkMessageRaw` is declared twice in `netdispatcher.h` — once public and once private. This should fail to compile or cause undefined behavior. Remove one.
- All previous fixes (mutex-protected `getRawCallback`, null-checks, typo fix, fallback documentation + debug log, `onMsgRaw` warning) look correct.

common/netdispatcher.cpp
banidoru: Confirmed fixed — `onNetlinkMessageRaw` now uses `getRawCallback()` which acquires the mutex. Race condition resolved.

common/netdispatcher.cpp
banidoru: Confirmed fixed — typo corrected and null-check added.

common/netdispatcher.cpp
banidoru: Confirmed fixed — code comment documents fallback-only behavior and `SWSS_LOG_DEBUG` added for observability.

common/netdispatcher.cpp
banidoru: Confirmed fixed — null-checks added to both `registerRawMessageHandler` and `registerMessageHandler`.

common/netmsg.h
banidoru: Confirmed fixed — default `onMsgRaw` now logs `SWSS_LOG_WARN` with nlmsg_type.

common/netdispatcher.h
banidoru: Confirmed fixed — duplicate public declaration of `onNetlinkMessageRaw` removed. Only the private declaration remains.

banidoru: Confirmed fixed — duplicate public declaration of `onNetlinkMessageRaw` removed. Only the private declaration remains.

banidoru (approved):
- All 6 prior review concerns addressed in latest commit
- Duplicate public declaration of `onNetlinkMessageRaw` removed (the only change in this iteration)
- Mutex protection, null-checks, typo fix, fallback documentation, and `SWSS_LOG_WARN` default all verified in place
- No new issues found in the diff
- LGTM — approving

banidoru (approved):
All prior concerns addressed in latest commits:

- ✅ Race condition in `onNetlinkMessageRaw` fixed — now uses mutex-protected `getRawCallback()`
- ✅ Typo and null-check in `registerRawMessageHandler` fixed
- ✅ Fallback-only behavior documented with `SWSS_LOG_DEBUG` added
- ✅ Default `onMsgRaw` now logs `SWSS_LOG_WARN` for observability
- ✅ Duplicate public declaration of `onNetlinkMessageRaw` removed from header

banidoru (approved):
All reviewers approved. LGTM.

common/netdispatcher.cpp
```diff
@@ -53,15 +97,33 @@ NetMsg* NetDispatcher::getCallback(int nlmsg_type)
...
     struct nlmsghdr *nlmsghdr = nlmsg_hdr(msg);
 
     auto callback = getCallback(nlmsghdr->nlmsg_type);
 
-    /* Drop not registered messages */
+    /* Raw handlers are only invoked when no parsed handler is registered.
+     * If both a parsed handler and a raw handler are registered for the same
+     * nlmsg_type, the raw handler will never fire. */
     if (callback == nullptr)
+    {
```
tahmed-dev, line 121: With the fallback-only model, `registerRawMessageHandler` and `registerMessageHandler` for the same `nlmsg_type` silently makes the raw handler dead code. Would it be worth adding a `SWSS_LOG_WARN` in `registerRawMessageHandler` if a parsed handler already exists for that type? Prevents subtle bugs where someone registers both expecting both to fire.

common/netmsg.h
```diff
@@ -9,5 +10,11 @@ namespace swss
         public:
             /* Called by NetDispatcher when netmsg matches filters */
             virtual void onMsg(int nlmsg_type, struct nl_object *obj) = 0;
+
+            /* Called by NetDispatcher when raw msg is send for matches filters */
+            virtual void onMsgRaw(struct nlmsghdr *hdr)
```
tahmed-dev: nit: The `SWSS_LOG_WARN` in the base class default is great for debugging, but in production with many netlink message types, could it get noisy if a subclass legitimately only handles `onMsg` but gets registered via `registerRawMessageHandler` by mistake? Maybe `SWSS_LOG_INFO` to reduce noise while still being observable?

tahmed-dev (commented):
Looks clean after the recent fixes. Two minor observations.

tahmed-dev (approved):
lgtm

common/netmsg.h
qiluo-msft: Could you move onMsgRaw() implmentation to its .cpp file, and move `#include "logger.h"` there also. It will reduce the compilation overhead for every downstream component which include this .h.

r12f: yea, likely debug level is better, unless it is not expected to be called at all. In this case, we should make it pure virtual (= 0), so we can enforce it during compilation.

common/netdispatcher.cpp
```diff
@@ -35,12 +53,38 @@ void NetDispatcher::unregisterMessageHandler(int nlmsg_type)
...
+{
+    MUTEX;
+
+    auto it = m_rawhandlers.find(nlmsg_type);
+
+    if (it == m_rawhandlers.end())
+        throw "Trying to unregister non existing handler";
+
+    m_rawhandlers.erase(it);
+
```
r12f: nit: empty line.

common/netdispatcher.cpp
```diff
@@ -35,12 +53,38 @@ void NetDispatcher::unregisterMessageHandler(int nlmsg_type)
...
+{
+    struct nlmsghdr *nlmsghdr = nlmsg_hdr(msg);
+
+    auto callback = getRawCallback(nlmsghdr->nlmsg_type);
+
+    /* Drop not registered messages */
+    if (callback == nullptr)
+        return;
+
+    callback->onMsgRaw(nlmsghdr);
```
r12f, line 84: anything to keep the callback alive during onMsgRaw?

tahmed-dev: Good catch on the lifetime concern. Addressed by holding the mutex during the `onMsgRaw` invocation — the lookup is inlined instead of going through `getRawCallback()` to avoid double-locking.

Worth noting though — all SONiC daemons that use `NetDispatcher` (orchagent, fdbsyncd, fpmsyncd, neighsyncd) are single-threaded, and handlers are registered once at startup and never unregistered during runtime. Also, the existing `onNetlinkMessage` (parsed path) doesn't actually hold the mutex during `callback->onMsg()` — it releases after `getCallback()` returns. So there's now an inconsistency between the two paths.

If consistency with the existing pattern is preferred (lookup under lock, callback without lock), happy to revert. The practical risk is zero given single-threaded callers.

common/netdispatcher.cpp
banidoru: Confirmed fixed — `onNetlinkMessageRaw` now holds the mutex for the entire duration of the callback lookup and invocation, which actually provides a stronger guarantee than the parsed path (`onNetlinkMessage` releases the lock after `getCallback()`). Race condition resolved.

common/netmsg.h
banidoru: Confirmed fixed — `onMsgRaw` implementation moved to `netmsg.cpp`, `#include "logger.h"` removed from the header. Compilation overhead reduced.

common/netdispatcher.cpp
banidoru: Confirmed fixed — empty line removed.

common/netdispatcher.cpp
banidoru: Addressed — the mutex is now held for the entire duration of `onMsgRaw` invocation, keeping the callback alive. This is actually stronger than the parsed path (which releases the lock before `onMsg`). As tahmed-dev noted, the inconsistency between the two paths is worth considering, but holding the lock during callback is the safer approach.

common/netmsg.h
banidoru: Addressed — changed from `SWSS_LOG_WARN` to `SWSS_LOG_DEBUG` per r12f's feedback, and moved to `netmsg.cpp`.

common/netdispatcher.cpp
banidoru: This concern is still open — no warning is emitted when `registerRawMessageHandler` is called for an nlmsg_type that already has a parsed handler. The code comment in `onNetlinkMessage` documents the fallback-only behavior, which is good, but a registration-time warning would catch misuse earlier.

banidoru (commented):
Re-review of new commits (41ec55a → deb6063). All prior concerns addressed:

- **Mutex lifetime**: `onNetlinkMessageRaw` now holds lock during entire callback — stronger guarantee than parsed path. ✅
- **onMsgRaw moved to .cpp**: Header no longer includes `logger.h`. WARN→DEBUG per feedback. ✅
- **Empty line nit, duplicate declaration**: Fixed. ✅
- **qiluo-msft's request to move impl to .cpp**: Done via new `netmsg.cpp`. ✅

One prior concern still open: no registration-time warning when raw + parsed handlers overlap (tahmed-dev's thread). Low severity — the runtime DEBUG log covers it partially.

No new issues found in the diff. LGTM pending the minor overlap-warning consideration.

common/netdispatcher.cpp
```diff
@@ -53,15 +101,33 @@ NetMsg* NetDispatcher::getCallback(int nlmsg_type)
...
 }
 
+NetMsg* NetDispatcher::getRawCallback(int nlmsg_type)
+{
+    MUTEX;
+
+    auto callback = m_rawhandlers.find(nlmsg_type);
+
+    if (callback == m_rawhandlers.end())
+        return nullptr;
```
banidoru, line 106: `getRawCallback()` is now dead code — `onNetlinkMessageRaw` does the lookup+dispatch inline under the mutex (which is the right approach for lifetime safety). Either remove `getRawCallback()` entirely or document why it's kept for future use.

common/netdispatcher.h
```diff
@@ -23,6 +23,21 @@ namespace swss
...
+            /*
+             * Unregister Raw Msg callback class according to message-type.
+             *
+             * Throw exception if callback is not registered
+             */
+            void unregisterRawMessageHandler(int nlmsg_type);
+
+
             /** Called by NetLink or FpmLink classes as indication of new packet arrival. */
             void onNetlinkMessage(struct nl_msg *msg);
```
banidoru, line 42: nit: extra blank line here (two consecutive blank lines between `unregisterRawMessageHandler` and the `onNetlinkMessage` comment).

banidoru (commented):
- All prior feedback addressed — mutex lifetime, typo, null-checks, onMsgRaw moved to .cpp, log level to DEBUG, empty line removed. Good work.
- `getRawCallback()` is now dead code — `onNetlinkMessageRaw` does the lookup inline under the mutex. Remove it or use it.
- Still no registration-time warning when raw + parsed handlers registered for same nlmsg_type (prior open thread).
- Extra blank line between `unregisterRawMessageHandler` declaration and the `onNetlinkMessage` comment in `netdispatcher.h` (line 43-44).

common/netdispatcher.cpp
banidoru: Still open — agreed this would be a useful safety net. A `SWSS_LOG_WARN` in `registerRawMessageHandler` when `m_handlers` already contains the same `nlmsg_type` (and vice versa) would catch misuse at registration time rather than silently dropping raw callbacks at dispatch time.

banidoru (approved):
All reviewers approved. LGTM.

common/netdispatcher.cpp
banidoru, line 106: Still open — `getRawCallback()` is dead code. `onNetlinkMessageRaw` does the lookup inline under the mutex. Please remove it or add a comment explaining why it's retained.

common/netdispatcher.h
banidoru, line 42: nit (still open): extra blank line here — two consecutive blank lines between `unregisterRawMessageHandler` and the next comment.

banidoru (commented):
Re-review (iteration 5) of new commits since 41ec55ac:

- All previously raised concerns from iterations 1-4 are addressed in this push: race condition fix, null-checks, typo fix, fallback documentation + DEBUG log, duplicate declaration removal, onMsgRaw moved to .cpp, log level changed to DEBUG per r12f, empty line nit fixed, callback lifetime protected by mutex.
- **Still open:** `getRawCallback()` is now dead code — `onNetlinkMessageRaw` does inline lookup under mutex. Should be removed or documented.
- **Still open:** Extra blank line in `netdispatcher.h` (two consecutive blanks after `unregisterRawMessageHandler` declaration).
- **Still open:** No registration-time warning when `registerRawMessageHandler` is called for an nlmsg_type that already has a parsed handler (tahmed-dev's suggestion).
- No new issues found in this iteration beyond the three still-open items above.

common/netdispatcher.cpp
banidoru, line 106: `getRawCallback()` is still dead code — `onNetlinkMessageRaw` does the lookup+dispatch inline under the mutex (correctly). Please remove this method or add a comment explaining why it's kept.

common/netdispatcher.h
banidoru, line 42: nit: extra blank line — two consecutive blank lines between `unregisterRawMessageHandler` and the next comment.

banidoru (commented):
Re-review (iteration 5) — most prior concerns addressed. Two items still open:

- **`getRawCallback()` is dead code** — `onNetlinkMessageRaw` does inline lookup under mutex. Remove it or document retention.
- **Extra blank line** in `netdispatcher.h` line 43 (two consecutive blank lines).
- Registration-time warning for dual parsed+raw handlers (tahmed-dev's suggestion) still not implemented — worth considering.

All other prior threads (race condition fix, null-checks, typo, fallback-only docs, WARN→DEBUG, move to .cpp, callback lifetime) confirmed fixed. LGTM otherwise.

banidoru (commented):
Re-review (iteration 5) — most prior concerns addressed. Two items remain:
- `getRawCallback()` is dead code — `onNetlinkMessageRaw` does inline lookup under mutex. Remove or document.
- Extra blank line in `netdispatcher.h` between `unregisterRawMessageHandler` and the next comment.
- Registration-time warning suggestion (from prior thread) still unaddressed but is a nice-to-have, not blocking.

common/netdispatcher.cpp
banidoru, line 106: `getRawCallback()` is dead code — `onNetlinkMessageRaw` does the lookup+dispatch inline under the mutex (correctly). Please remove this method or add a comment explaining why it's retained.

common/netdispatcher.h
banidoru, line 42: nit: extra blank line — two consecutive blank lines between `unregisterRawMessageHandler` and the next comment.

qiluo-msft (dismissed):

common/netdispatcher.cpp
banidoru: Confirmed still fixed — mutex held during callback invocation in `onNetlinkMessageRaw`.

common/netdispatcher.cpp
banidoru: Confirmed still fixed in latest commit.

common/netdispatcher.cpp
banidoru: Confirmed still fixed in latest commit.

common/netdispatcher.cpp
banidoru: Confirmed still fixed in latest commit.

common/netmsg.h
banidoru: Confirmed still fixed in latest commit.

common/netdispatcher.h
banidoru: Confirmed still fixed in latest commit.

common/netmsg.h
banidoru: Confirmed still fixed in latest commit.

common/netdispatcher.cpp
banidoru: Confirmed still fixed in latest commit.

common/netmsg.h
banidoru: Confirmed fixed — changed to `SWSS_LOG_DEBUG` and moved implementation to `netmsg.cpp`.

common/netdispatcher.cpp
banidoru: Confirmed fixed — mutex held during `onMsgRaw` invocation, keeping the callback alive.

common/netdispatcher.cpp
banidoru: Still open — no registration-time warning when registering a raw handler for an nlmsg_type that already has a parsed handler (or vice versa).

common/netdispatcher.cpp
banidoru: Still open — `getRawCallback()` is dead code. It is never called; `onNetlinkMessageRaw` does the lookup inline. Please remove or document.

common/netdispatcher.h
banidoru: Still open — extra blank line at line 43 in netdispatcher.h.

tests/netdispatcher_ut.cpp
```diff
@@ -0,0 +1,157 @@
...
+    TestNetMsg handler;
+
+    dispatcher.registerRawMessageHandler(4000, &handler);
+
+    /* Simulate onNetlinkMessageRaw by calling it indirectly through onNetlinkMessage
+     * when no parsed handler exists for this type. */
+
+    /* Cleanup */
+    dispatcher.unregisterRawMessageHandler(4000);
+}
```
banidoru, line 115: This test registers and unregisters a handler but doesn't assert anything meaningful — it's essentially a no-op. Consider either removing it or making it exercise the raw dispatch path (like `OnNetlinkMessageFallbackToRaw` already does).

banidoru (commented):
Iteration 6 re-review (commit bbe2d489):

**New commit:** Adds unit tests for raw netlink message handler infrastructure — good coverage of register/unregister/null-check/fallback paths.

**Resolved concerns:**
- Mutex race in `onNetlinkMessageRaw` ✅
- Typo in exception message ✅
- Fallback-only behavior documented + `SWSS_LOG_DEBUG` ✅
- Null-check on callbacks ✅
- `onMsgRaw` default logs at DEBUG level, moved to .cpp ✅
- Duplicate public declaration removed ✅
- Empty line in netdispatcher.cpp ✅
- Callback lifetime during `onMsgRaw` ✅

**Still open:**
- `getRawCallback()` is dead code — never called, `onNetlinkMessageRaw` does inline lookup. Remove or document.
- Extra blank line in `netdispatcher.h` (line 43) — two consecutive blank lines.
- No registration-time warning when registering raw+parsed handlers for the same nlmsg_type.

**New test feedback:**
- Tests look solid. `GetRawCallbackRegistered` test (line ~100) registers a handler and immediately unregisters without actually testing anything — consider removing or making it exercise `getRawCallback()` (if kept) or the fallback path.

banidoru (commented):
**COMMENT** — prior fixes confirmed, 3 items still open:

- `getRawCallback()` is dead code — never called. Remove it or document why it is kept.
- Extra blank line in `netdispatcher.h` (line 39) — two consecutive blank lines after `unregisterRawMessageHandler`.
- No registration-time warning when registering a raw handler for an nlmsg_type that already has a parsed handler (or vice versa) — silent dead-code trap.
- `GetRawCallbackRegistered` test asserts nothing meaningful — register+unregister with no verification is not a useful test.

banidoru (commented):
- **COMMENT** — all major concerns from prior iterations are fixed
- `getRawCallback()` remains dead code — consider removing or documenting why it is kept
- Extra blank line at netdispatcher.h line 39-40 still present (minor style nit)
- No registration-time warning when both parsed and raw handlers share an nlmsg_type (suggestion from prior review, not blocking)
- `GetRawCallbackRegistered` test registers/unregisters without asserting anything — either add assertions or remove the test

common/netdispatcher.cpp
qiluo-msft: this is valid comment, please fix.

tahmed-dev: Fixed by making it not dead — `onNetlinkMessageRaw` now uses `getRawCallback()`, mirroring `onNetlinkMessage` + `getCallback()`. Updated the `GetRawCallbackRegistered` test to actually assert nullptr / pointer / nullptr around register/unregister.

tahmed-dev:
Pushed fixup `b859e3d` addressing the review:

- `onNetlinkMessageRaw` now uses `getRawCallback()` — symmetric with `onNetlinkMessage` + `getCallback()`. The method is no longer dead code.
- `GetRawCallbackRegistered` test rewritten to actually assert: nullptr → register → pointer → unregister → nullptr.
- Stray double blank line in `netdispatcher.h` collapsed.

CodeQL hit at `netdispatcher.cpp:41` is the same store pattern as `registerMessageHandler` (line 28) — pointer value, not parameter address. False positive — can it be dismissed?

banidoru (approved):
- All prior review concerns addressed — mutex race fix, null-checks, typo, dead code removal, debug logging, header cleanup all verified
- `getRawCallback()` now used properly; consistent with parsed path pattern
- Test coverage solid with proper assertions for register/unregister/dispatch/fallback
- One minor suggestion remains open: registration-time warning when raw+parsed handlers overlap (nice-to-have, not blocking)
- LGTM — approving

banidoru (commented):
- All prior issues from iterations 1-6 confirmed fixed: mutex race, null checks, typo, dead code, duplicate declarations, log level, cpp file separation
- One open suggestion remains: add `SWSS_LOG_WARN` in `registerRawMessageHandler` when a parsed handler already exists for the same `nlmsg_type` (and vice versa) to catch misuse at registration time — low priority but good defensive coding
- Code is clean, well-tested, consistent with existing patterns
- Test coverage is solid — covers register/unregister/null/fallback/default paths
- No new issues found in this iteration

banidoru (approved):
- All prior blocking issues fixed: mutex race, null-checks, dead code (`getRawCallback` now used), duplicate declaration, `onMsgRaw` moved to cpp with DEBUG log level
- Fallback-only design properly documented with code comment + `SWSS_LOG_DEBUG` on fallthrough
- Good test coverage for register/unregister/fallback/null/base-class paths
- Minor: registration-time warning when raw+parsed handlers overlap (thread from tahmed-dev) remains unaddressed — low risk given documented fallback-only behavior, but would be a nice safety net
- Clean, consistent with existing codebase patterns

banidoru (approved):
All reviewers approved. LGTM.

common/netdispatcher.cpp
banidoru: Fixed — `getRawCallback()` is now used by `onNetlinkMessageRaw`, mirroring the `getCallback()`/`onNetlinkMessage` pattern. No longer dead code.

tests/netdispatcher_ut.cpp
banidoru: Fixed — `GetRawCallbackRegistered` now asserts nullptr before register, pointer after register, and nullptr after unregister.

common/netdispatcher.cpp
banidoru: Approach changed — `onNetlinkMessageRaw` now uses `getRawCallback()` (mutex released before callback invocation), consistent with the parsed path (`getCallback()` + `onMsg`). This addresses the consistency concern. The lifetime risk is theoretical given single-threaded callers and startup-only registration.

common/netdispatcher.cpp
banidoru: Fixed — `getRawCallback()` is now used by `onNetlinkMessageRaw`, no longer dead code.

banidoru: Fixed — `getRawCallback()` is now used by `onNetlinkMessageRaw`, no longer dead code.

banidoru: Fixed — `getRawCallback()` is now used by `onNetlinkMessageRaw`, no longer dead code.

common/netdispatcher.h
banidoru: Fixed — extra blank line removed.

banidoru: Fixed — extra blank line removed.

banidoru: Fixed — extra blank line removed.

banidoru: Fixed — extra blank line removed.

common/netmsg.h
banidoru: Confirmed still fixed — using `SWSS_LOG_DEBUG` and implementation in `netmsg.cpp`.

common/netdispatcher.cpp
banidoru: Still open — no registration-time warning when registering a raw handler for an nlmsg_type that already has a parsed handler (or vice versa). This is a minor safety net request, not a blocker.

banidoru (commented):
- All prior review concerns addressed in latest commit
- `getRawCallback()` no longer dead code — now used by `onNetlinkMessageRaw`, consistent with `getCallback()`/`onNetlinkMessage` pattern
- `GetRawCallbackRegistered` test now has meaningful assertions (nullptr/pointer/nullptr lifecycle)
- Callback lifetime: mutex released before invocation, matching parsed handler path — acceptable given single-threaded callers
- One minor open item remains: no registration-time warning when both parsed and raw handlers registered for same nlmsg_type (non-blocking)

banidoru (commented):
Re-review (iteration 8) — all prior issues addressed except one minor open item:

- ✅ Mutex race condition in `onNetlinkMessageRaw`: fixed via `getRawCallback()`
- ✅ Dead code `getRawCallback()`: now used, consistent with `getCallback()` pattern
- ✅ Callback lifetime concern: `getRawCallback()` mirrors parsed path — consistent
- ✅ `onMsgRaw` moved to `netmsg.cpp` with `SWSS_LOG_DEBUG` per r12f feedback
- ⚠️ Registration-time warning (minor): no `SWSS_LOG_WARN` when registering raw+parsed for same nlmsg_type — still open, not a blocker

banidoru (approved):
All reviewers approved. LGTM.

common/netdispatcher.cpp
r12f: sounds good! thanks for checking!

r12f (approved):
lgtm! thanks Barry!

qiluo-msft (approved):

PR #1157, opened 2026-03-06
Move ZMQ socket bind to poll thread to prevent orch blocking

ypcisco:
Why I did it

- In scale scenarios, zmq_bind() can take significant time to complete
- Orchestration agent gets blocked during ZmqServer initialization
- Synchronous bind in main thread delays startup and impacts system responsiveness

How I did it

- Moved zmq_bind() call from ZmqServer::bind() to mqPollThread()
- Socket creation and configuration remain in main thread
- Actual bind operation now happens in background poll thread

### DIFF
--- a/common/zmqserver.cpp
+++ b/common/zmqserver.cpp
@@ -76,15 +76,7 @@ void ZmqServer::bind()
         zmq_setsockopt(m_socket, ZMQ_BINDTODEVICE, m_vrf.c_str(), m_vrf.length());
     }
 
-    int rc = zmq_bind(m_socket, m_endpoint.c_str());
-    if (rc != 0)
-    {
-        SWSS_LOG_THROW("zmq_bind failed on endpoint: %s, zmqerrno: %d",
-            m_endpoint.c_str(),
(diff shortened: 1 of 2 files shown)


anish-n (approved):
Approve, but I would request Qi or someone else with more context on this aspect to review

common/zmqserver.cpp
```diff
@@ -153,6 +145,14 @@ void ZmqServer::mqPollThread()
     SWSS_LOG_ENTER();
     SWSS_LOG_NOTICE("mqPollThread begin");
 
+    SWSS_LOG_NOTICE("Attempting to bind to zmq endpoint: %s", m_endpoint.c_str());
+    if (zmq_bind(m_socket, m_endpoint.c_str()) != 0)
+    {
+        SWSS_LOG_THROW("zmq_bind failed on endpoint: %s, zmqerrno: %d",
```
qiluo-msft: This thread has a catch-all block. Throwing exception will crash the process immediately.

qiluo-msft (dismissed):

PR #1162, opened 2026-03-23
Introduce keepalives for ZmqClient and ZmqServer

prabhataravind:
Fixes issue: sonic-net/sonic-buildimage#23110

When a DPU is powered off and back on, the ZMQ client on the switch still holds a stale TCP connection. The first message sent after DPU restart is delivered over the dead connection, gets a TCP RST, and is silently lost. ZMQ then auto-reconnects, so subsequent messages succeed.

This patch enables:
1. TCP keepalive on ZmqClient PUSH sockets to detect dead connections proactively (within ~10 seconds of peer going down).
2. TCP keepalive on ZmqServer PULL sockets as defense-in-depth.

With these changes, after DPU power-off:
- TCP keepalive probes will fail, causing ZMQ to tear down the stale connection and reconnect

### DIFF
--- a/common/zmqclient.cpp
+++ b/common/zmqclient.cpp
@@ -110,11 +110,13 @@ void ZmqClient::connect()
         int high_watermark = MQ_WATERMARK;
         zmq_setsockopt(m_socket, ZMQ_SNDHWM, &high_watermark, sizeof(high_watermark));
 
-        // Enable TCP keepalive to detect dead connections (e.g. DPU power-off).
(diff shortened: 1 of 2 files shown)


prsunny (approved):
lgtm

vmittal-msft:
@prabhataravind please raise manual pr for 202511 due to cherry-pick conflict

prsunny:
@qiluo-msft for viz

PR #1174, opened 2026-04-07
[dbconnector]: Fall back to TCP when unix socket path is empty

theasianpianist:
When `isTcpConn` is `false` but the configured unix socket path for the database is empty, the DBConnector constructor now falls back to using a TCP connection instead of attempting to connect to an empty socket path. A `SWSS_LOG_WARN` is emitted when this fallback occurs.

### DIFF
--- a/common/dbconnector.cpp
+++ b/common/dbconnector.cpp
@@ -716,7 +716,8 @@ DBConnector::DBConnector(const string& dbName, unsigned int timeout_ms, bool isT
     }
     else if (SonicDBConfig::getDbSock(dbName, m_key).empty())
     {
-        SWSS_LOG_WARN("Unix socket path for DB %s is empty, falling back to TCP connection", dbName.c_str());
+        SWSS_LOG_WARN("Unix socket path for DB %s in context %s is empty, falling back to TCP connection",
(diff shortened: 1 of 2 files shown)


prsunny (commented):
lgtm, @qiluo-msft help review/merge

prsunny (approved):

common/dbconnector.cpp
```diff
@@ -716,7 +716,17 @@ DBConnector::DBConnector(const string& dbName, unsigned int timeout_ms, bool isT
     }
     else
     {
-        initContext(SonicDBConfig::getDbSock(dbName, m_key).c_str(), ptv);
+        string sockPath = SonicDBConfig::getDbSock(dbName, m_key);
+        if (sockPath.empty())
+        {
+            SWSS_LOG_WARN("Unix socket path for DB %s in context %s is empty, falling back to TCP connection",
```
qiluo-msft, line 722: this should be treated as error, and exit.

theasianpianist: Why error out here, is there any concern w/ falling back to TCP connection?

theasianpianist:
Closing, will implement this logic at higher-level applications to avoid changes to core APIs

PR #1178, opened 2026-04-21
Increase ZMQ Response Size to 32MB to support maximum dash configuration size

croos12:
Why I did it
SONiC Orchagent zmq does not currently support a large enough buffer size to handle the max dash scale config on DPU. 

How I did it
I increased the zmq response size to 32MB.

How to verify it
Send the maximum dash scale config in a test and confirm config is applied.

### DIFF
--- a/common/zmqserver.h
+++ b/common/zmqserver.h
@@ -6,7 +6,7 @@
 #include <vector>
 #include "table.h"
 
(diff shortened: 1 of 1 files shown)


croos12:
Closing after discussion in https://github.com/sonic-net/sonic-buildimage/pull/26679

PR #1180, opened 2026-04-24
Add reusable ComponentStats library for per-container statistics.

yutongzhang-microsoft:
## What I did

Added a new reusable library `swss::ComponentStats` under `libswsscommon`, together with its unit tests. The library lets any SONiC container publish service-level counters into `COUNTERS_DB` under a uniform `<COMPONENT>_STATS:<entity>` key layout.

This is a purely additive change — no existing API is modified or removed, and no current consumer of `libswsscommon` calls into the new code, so behaviour of every existing SONiC binary is unchanged.

## Why I did it

SONiC already has dataplane counters (Flex-Counter → syncd → `COUNTERS_DB`). What we are missing is **service-level** counters — software-side events such as orchagent task throughput, gNMI request rate, BMP message error counts. Without them we cannot answer questions like *"is orchagent draining tasks?"*, *"is gNMI seeing subscribe failures?"*, *"is one container dropping more events than its peers?"*.

If each container rolled its own plumbing for this, every container would carry its own concurrency review, the bug fixes would drift, and the on-the-wire schemas would diverge. A single shared library is the only way to keep these properties consistent across SONiC.

Full motivation and design are in the HLD: [sonic-net/SONiC#2312](https://github.com/sonic-net/SONiC/pull/2312).

## How I did it

- Implemented the library as a producer with a lock-free hot path, a 1 s background writer thread, version-based dirty tracking (idle entities produce no Redis traffic), lazy DB connection with retry, and a per-component singleton.
- Kept the public API narrow and stable so future containers only need a ~30 LoC facade to adopt it.
- Added a unit-test suite (`tests/componentstats_ut.cpp`, 9 cases) covering basic round-trip, gauge semantics, runtime enable/disable, bulk read, singleton behaviour, and 8-thread × 10 000-increment concurrency.
- Verified the library can be merged independently: the first consumer ([sonic-swss#4516](https://github.com/sonic-net/sonic-swss/pull/4516)) depends on this PR but this PR does not depend on it.

## Related

- HLD: [sonic-net/SONiC#2312](https://github.com/sonic-net/SONiC/pull/2312)
- First consumer (SwssStats facade): [sonic-net/sonic-swss#4516](https://github.com/sonic-net/sonic-swss/pull/4516)
- Build / integration test: [sonic-net/sonic-buildimage#27456](https://github.com/sonic-net/sonic-buildimage/pull/27456)

### DIFF
--- a/common/componentstats.cpp
+++ b/common/componentstats.cpp
@@ -44,17 +44,30 @@ std::shared_ptr<ComponentStats> ComponentStats::create(
     uint32_t intervalSec)
 {
     const std::string key = toUpper(componentName);
+    const uint32_t effectiveInterval = intervalSec == 0 ? 1 : intervalSec;
 
     std::lock_guard<std::mutex> lock(registryMutex());
     auto& slot = registry()[key];
     if (auto existing = slot.lock())
     {
+        // First call wins. Warn if a later caller passes different params
+        // so the inconsistency is at least visible in syslog.
+        if (existing->m_dbName != dbName ||
+            existing->m_intervalSec != effectiveInterval)
+        {
+            SWSS_LOG_WARN("ComponentStats[%s]: ignoring later create() with "
+                          "db=%s interval=%us; live instance uses db=%s "
+                          "interval=%us",
+                          key.c_str(),
+                          dbName.c_str(), effectiveInterval,
+                          existing->m_dbName.c_str(), existing->m_intervalSec);
+        }
         return existing;
     }
 
     // std::make_shared cannot access the private constructor; use new + wrap.
     std::shared_ptr<ComponentStats> inst(
-        new ComponentStats(key, dbName, intervalSec));
+        new ComponentStats(key, dbName, effectiveInterval));
     slot = inst;
     return inst;
 }
@@ -125,9 +138,12 @@ ComponentStats::getOrCreateCounter(EntityStats& e, const std::string& metric)
     auto it = e.metrics.find(metric);
     if (it == e.metrics.end())
     {
-        it = e.metrics.emplace(metric, std::make_unique<Counter>()).first;
+        // std::map::emplace constructs Counter in place; no move required.
+        it = e.metrics.emplace(std::piecewise_construct,
+                               std::forward_as_tuple(metric),
+                               std::forward_as_tuple()).first;
     }
-    return *it->second;
+    return it->second;
 }
 
 void ComponentStats::increment(const std::string& entity,
@@ -167,7 +183,7 @@ uint64_t ComponentStats::get(const std::string& entity, const std::string& metri
     if (eit == m_entities.end()) return 0;
     auto mit = eit->second.metrics.find(metric);
     if (mit == eit->second.metrics.end()) return 0;
-    return mit->second->value.load(std::memory_order_relaxed);
+    return mit->second.value.load(std::memory_order_relaxed);
 }
 
 ComponentStats::CounterSnapshot
@@ -179,7 +195,7 @@ ComponentStats::getAll(const std::string& entity)
     if (eit == m_entities.end()) return snap;
     for (const auto& kv : eit->second.metrics)
     {
-        snap[kv.first] = kv.second->value.load(std::memory_order_relaxed);
+        snap[kv.first] = kv.second.value.load(std::memory_order_relaxed);
     }
     return snap;
 }
@@ -224,6 +240,10 @@ void ComponentStats::writerThread()
             }
         }
 
+        // Defensive: the connect-retry loop above only exits via `break`
+        // (m_table set) or via m_running becoming false (in which case the
+        // m_running check at the top of this loop already broke out). This
+        // line therefore should not normally be reached.
         if (!m_table) continue;
 
         std::vector<std::string> keys;
@@ -254,7 +274,7 @@ void ComponentStats::writerThread()
                 {
                     row.emplace_back(
                         mkv.first,
-                        std::to_string(mkv.second->value.load(std::memory_order_relaxed)));
+                        std::to_string(mkv.second.value.load(std::memory_order_relaxed)));
                 }
                 keys.push_back(name);
                 values.push_back(std::move(row));
--- a/common/componentstats.cpp
+++ b/common/componentstats.cpp
@@ -1,0 +1,284 @@
+#include "componentstats.h"
+
+#include "dbconnector.h"
+#include "table.h"
+#include "logger.h"
+
+#include <algorithm>
+#include <cctype>
+#include <chrono>
+#include <exception>
+#include <unordered_map>
+#include <utility>
+
+namespace swss {
+
+namespace {
+
+std::string toUpper(std::string s)
+{
+    std::transform(s.begin(), s.end(), s.begin(),
+                   [](unsigned char c){ return std::toupper(c); });
+    return s;
+}
+
+// Registry of created instances keyed by component name, so callers that
+// forget to hold onto the shared_ptr still get the same singleton back.
+std::mutex& registryMutex()
+{
+    static std::mutex m;
+    return m;
+}
+
+std::map<std::string, std::weak_ptr<ComponentStats>>& registry()
+{
+    static std::map<std::string, std::weak_ptr<ComponentStats>> r;
+    return r;
+}
+
+} // namespace
+
+std::shared_ptr<ComponentStats> ComponentStats::create(
+    const std::string& componentName,
+    const std::string& dbName,
+    uint32_t intervalSec)
+{
+    const std::string key = toUpper(componentName);
+
+    std::lock_guard<std::mutex> lock(registryMutex());
+    auto& slot = registry()[key];
+    if (auto existing = slot.lock())
+    {
+        return existing;
+    }
+
+    // std::make_shared cannot access the private constructor; use new + wrap.
+    std::shared_ptr<ComponentStats> inst(
+        new ComponentStats(key, dbName, intervalSec));
+    slot = inst;
+    return inst;
+}
+
+ComponentStats::ComponentStats(const std::string& componentName,
+                               const std::string& dbName,
+                               uint32_t intervalSec)
+    : m_component(componentName)
+    , m_tableName(componentName + "_STATS")
+    , m_dbName(dbName)
+    , m_intervalSec(intervalSec == 0 ? 1 : intervalSec)
+{
+    SWSS_LOG_ENTER();
+    m_thread = std::make_unique<std::thread>(&ComponentStats::writerThread, this);
+    SWSS_LOG_NOTICE("ComponentStats[%s] initialized (db=%s, interval=%us)",
+                    m_component.c_str(), m_dbName.c_str(), m_intervalSec);
+}
+
+ComponentStats::~ComponentStats()
+{
+    stop();
+}
+
+void ComponentStats::stop()
+{
+    if (!m_running.exchange(false))
+    {
+        return; // already stopped
+    }
+    m_cv.notify_all();
+    if (m_thread && m_thread->joinable())
+    {
+        m_thread->join();
+    }
+    SWSS_LOG_NOTICE("ComponentStats[%s] stopped", m_component.c_str());
+}
+
+void ComponentStats::setEnabled(bool on)
+{
+    m_enabled.store(on, std::memory_order_relaxed);
+}
+
+bool ComponentStats::isEnabled() const
+{
+    return m_enabled.load(std::memory_order_relaxed);
+}
+
+ComponentStats::EntityStats&
+ComponentStats::getOrCreateEntity(const std::string& entity)
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+    auto it = m_entities.find(entity);
+    if (it == m_entities.end())
+    {
+        it = m_entities.emplace(std::piecewise_construct,
+                                std::forward_as_tuple(entity),
+                                std::forward_as_tuple()).first;
+    }
+    // std::map never invalidates references to existing elements on insert,
+    // so returning this reference after the lock is released is safe.
+    return it->second;
+}
+
+ComponentStats::Counter&
+ComponentStats::getOrCreateCounter(EntityStats& e, const std::string& metric)
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+    auto it = e.metrics.find(metric);
+    if (it == e.metrics.end())
+    {
+        it = e.metrics.emplace(metric, std::make_unique<Counter>()).first;
+    }
+    return *it->second;
+}
+
+void ComponentStats::increment(const std::string& entity,
+                               const std::string& metric,
+                               uint64_t n)
+{
+    if (!isEnabled() || n == 0)
+    {
+        return;
+    }
+    auto& e = getOrCreateEntity(entity);
+    auto& c = getOrCreateCounter(e, metric);
+    c.value.fetch_add(n, std::memory_order_relaxed);
+    // Release ordering pairs with the acquire in writerThread() so that counter
+    // writes become visible before the version bump.
+    e.version.fetch_add(1, std::memory_order_release);
+}
+
+void ComponentStats::setValue(const std::string& entity,
+                              const std::string& metric,
+                              uint64_t value)
+{
+    if (!isEnabled())
+    {
+        return;
+    }
+    auto& e = getOrCreateEntity(entity);
+    auto& c = getOrCreateCounter(e, metric);
+    c.value.store(value, std::memory_order_relaxed);
+    e.version.fetch_add(1, std::memory_order_release);
+}
+
+uint64_t ComponentStats::get(const std::string& entity, const std::string& metric)
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+    auto eit = m_entities.find(entity);
+    if (eit == m_entities.end()) return 0;
+    auto mit = eit->second.metrics.find(metric);
+    if (mit == eit->second.metrics.end()) return 0;
+    return mit->second->value.load(std::memory_order_relaxed);
+}
+
+ComponentStats::CounterSnapshot
+ComponentStats::getAll(const std::string& entity)
+{
+    CounterSnapshot snap;
+    std::lock_guard<std::mutex> lock(m_mutex);
+    auto eit = m_entities.find(entity);
+    if (eit == m_entities.end()) return snap;
+    for (const auto& kv : eit->second.metrics)
+    {
+        snap[kv.first] = kv.second->value.load(std::memory_order_relaxed);
+    }
+    return snap;
+}
+
+void ComponentStats::writerThread()
+{
+    SWSS_LOG_ENTER();
+    SWSS_LOG_NOTICE("ComponentStats[%s] writer thread started",
+                    m_component.c_str());
+
+    // Defer the DB connection: early-startup containers may call create()
+    // before Redis is reachable. Retry on failure with a bounded back-off.
+    while (m_running.load(std::memory_order_relaxed))
+    {
+        try
+        {
+            m_db = std::make_shared<DBConnector>(m_dbName, 0);
+            m_table = std::make_unique<Table>(m_db.get(), m_tableName);
+            break;
+        }
+        catch (const std::exception& e)
+        {
+            SWSS_LOG_WARN("ComponentStats[%s]: DB connect failed: %s. Retrying…",
+                          m_component.c_str(), e.what());
+            std::unique_lock<std::mutex> lock(m_mutex);
+            m_cv.wait_for(lock, std::chrono::seconds(m_intervalSec),
+                          [this]{ return !m_running.load(std::memory_order_relaxed); });
+        }
+    }
+
+    std::unordered_map<std::string, uint64_t> lastVersions;
+
+    while (true)
+    {
+        {
+            std::unique_lock<std::mutex> lock(m_mutex);
+            m_cv.wait_for(lock, std::chrono::seconds(m_intervalSec),
+                          [this]{ return !m_running.load(std::memory_order_relaxed); });
+            if (!m_running.load(std::memory_order_relaxed))
+            {
+                break;
+            }
+        }
+
+        if (!m_table) continue;
+
+        std::vector<std::string> keys;
+        std::vector<std::vector<FieldValueTuple>> values;
+
+        {
+            std::lock_guard<std::mutex> lock(m_mutex);
+            for (const auto& entry : m_entities)
+            {
+                const std::string& name = entry.first;
+                const EntityStats& stats = entry.second;
+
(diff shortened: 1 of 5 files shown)


common/componentstats.h
```diff
@@ -0,0 +1,166 @@
...
+ * Each instance writes counters to a Redis hash of the form:
+ *     <COMPONENT>_STATS:<entity>
+ * where each field is a user-defined metric name (e.g. SET, DEL, GET, ERROR...).
+ *
+ * Design goals:
+ *   - Lock-free counter update: once an entity/metric pair has been seen,
+ *     increment() / setValue() update the counter with std::atomic ops.
+ *     The first reference to a new entity or metric still takes m_mutex
+ *     to insert into the maps.
+ *   - Deferred DB connection: the DBConnector is created inside the writer
```
banidoru: The comment claims "Lock-free counter update: once an entity/metric pair has been seen, increment() / setValue() update the counter with std::atomic ops." — but the implementation in `componentstats.cpp` acquires `m_mutex` **twice** on every call (once in `getOrCreateEntity`, once in `getOrCreateCounter`), even when both already exist.

To actually deliver on this promise, `increment()`/`setValue()` need a lock-free fast path — e.g. cache the `EntityStats&` and `Counter&` at the call site and only fall back to the locked `getOrCreate*` on first use, or use a read-optimized pattern (shared_mutex with read-lock for lookups, write-lock only for inserts).

common/componentstats.cpp
```diff
@@ -0,0 +1,349 @@
...
+    return it->second;
+}
+
+void ComponentStats::increment(const std::string& entity,
+                               const std::string& metric,
+                               uint64_t n)
+{
+    if (!isEnabled() || n == 0)
+    {
+        return;
```
banidoru: `getOrCreateEntity` and `getOrCreateCounter` each take `m_mutex`. For the hot path where the entity and metric already exist, this means two full mutex lock/unlock cycles per `increment()` call. With 8+ threads hammering counters, this serializes the "lock-free" path.

Consider at minimum combining them into a single lock acquisition, or better yet using `std::shared_mutex` (read-lock for lookup, write-lock for insert) so concurrent reads don't contend.

common/componentstats.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+    while (true)
+    {
+        {
+            std::unique_lock<std::mutex> lock(m_mutex);
+            m_cv.wait_for(lock, std::chrono::seconds(m_intervalSec),
+                          [this]{ return !m_running.load(std::memory_order_relaxed); });
+            if (!m_running.load(std::memory_order_relaxed))
+            {
+                break;
```
banidoru: When `m_running` becomes false, the writer breaks out of the loop immediately without flushing dirty entities. Any counters updated since the last periodic flush are silently lost.

Add a final flush pass after the loop exits (while `m_table` is still valid) so that the last batch of dirty data lands in Redis before shutdown.

common/componentstats.cpp
```diff
@@ -0,0 +1,349 @@
...
+} // namespace
+
+std::shared_ptr<ComponentStats> ComponentStats::create(
+    const std::string& componentName,
+    const std::string& dbName,
+    uint32_t intervalSec)
+{
+    const std::string key = toUpper(componentName);
+    const uint32_t effectiveInterval = intervalSec == 0 ? 1 : intervalSec;
+
```
banidoru: Expired `weak_ptr` entries are never removed from the registry. If a process creates and destroys many differently-named `ComponentStats` instances over its lifetime, this map grows without bound.

Easy fix: prune expired entries inside `create()` (e.g. iterate and erase expired slots before inserting), or erase the entry in the destructor under `registryMutex()`.

common/componentstats.cpp
```diff
@@ -0,0 +1,349 @@
...
+    : m_component(componentName)
+    , m_tableName(componentName + "_STATS")
+    , m_dbName(dbName)
+    , m_intervalSec(intervalSec == 0 ? 1 : intervalSec)
+{
+    SWSS_LOG_ENTER();
+    m_thread = std::make_unique<std::thread>(&ComponentStats::writerThread, this);
+    SWSS_LOG_NOTICE("ComponentStats[%s] initialized (db=%s, interval=%us)",
+                    m_component.c_str(), m_dbName.c_str(), m_intervalSec);
+}
```
banidoru: `intervalSec` is clamped to 1 both here (line 87: `m_intervalSec(intervalSec == 0 ? 1 : intervalSec)`) and in `create()` (line 51: `effectiveInterval = intervalSec == 0 ? 1 : intervalSec`). The double-clamping is harmless but unnecessary — pick one location and document it there.

banidoru (changes requested):
- **Hot path is not lock-free** — despite the design-goal claim, every `increment()`/`setValue()` acquires `m_mutex` twice (once in `getOrCreateEntity`, once in `getOrCreateCounter`), even when the entity+metric already exist. This is the critical perf gap.
- **No final flush on shutdown** — `stop()` wakes the writer thread, which breaks out of the loop without flushing dirty counters. Data written just before teardown is silently lost.
- **Singleton registry leaks entries** — expired `weak_ptr` slots are never pruned from the static registry map; long-lived processes cycling component names will accumulate dead entries.
- **Overall**: solid Redis reconnect logic, clean API surface, good test coverage. The lock contention on the hot path needs fixing before this can deliver on the stated lock-free promise.

common/componentstats.h
```diff
@@ -0,0 +1,166 @@
...
+        EntityStats& operator=(const EntityStats&) = delete;
+    };
+
+    ComponentStats(const std::string& componentName,
+                   const std::string& dbName,
+                   uint32_t intervalSec);
+
+    // Returns a stable reference. Safe to keep after the lock is released
+    // because std::map does not invalidate element references on insert.
+    // INVARIANT: no code path may erase from m_entities or from
```
banidoru: The two separate `getOrCreate` functions are called sequentially in `increment()` and `setValue()`, each taking the main mutex. This results in two separate lock/unlock cycles on the hot path for every new entity/metric pair, which is inefficient. Consider combining these into a single `getOrCreateCounter(entity, metric)` method that takes the lock once, creates the entity if needed, and then creates the counter if needed, returning a reference to the counter. This would halve the locking overhead for new metrics.

common/componentstats.cpp
```diff
@@ -0,0 +1,349 @@
...
+    // so returning this reference after the lock is released is safe.
+    return it->second;
+}
+
+ComponentStats::Counter&
+ComponentStats::getOrCreateCounter(EntityStats& e, const std::string& metric)
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+    auto it = e.metrics.find(metric);
+    if (it == e.metrics.end())
```
banidoru: Calling `getOrCreateEntity` and then `getOrCreateCounter` separately releases the mutex between the two operations. While `std::map` reference stability prevents a crash, there's a potential but unlikely ABA-style problem if an entity were to be created, deleted (if a delete API is ever added), and re-created. Combining them into one locked operation as suggested for `componentstats.h` would make this pattern more robust and future-proof.

common/componentstats.cpp
```diff
@@ -0,0 +1,349 @@
...
+    {
+        // std::map::emplace constructs Counter in place; no move required.
+        it = e.metrics.emplace(std::piecewise_construct,
```
banidoru: The memory order for this `fetch_add` can be relaxed to `std::memory_order_relaxed`. The subsequent `fetch_add` on the version with `std::memory_order_release` is sufficient to ensure that this counter update is visible to the writer thread before the version is updated. The release-acquire pair on the version number acts as the necessary memory barrier.

common/componentstats.cpp
```diff
@@ -0,0 +1,349 @@
...
+    if (eit == m_entities.end()) return snap;
+    for (const auto& kv : eit->second.metrics)
+    {
+        snap[kv.first] = kv.second.value.load(std::memory_order_relaxed);
+    }
+    return snap;
+}
+
+void ComponentStats::writerThread()
+{
```
banidoru: The retry logic here for the initial connection uses `wait_for` on the condition variable. This is fine, but it could be slightly more robust. If the Redis server is down during startup, this loop will retry every `m_intervalSec`. If `m_intervalSec` is large (e.g., 60s), the component might take a very long time to connect. A simple exponential backoff strategy (e.g., 1s, 2s, 4s, ... up to a max) for the retry delay would be more conventional and potentially recover faster without spamming on short intervals.

common/componentstats.cpp
```diff
@@ -0,0 +1,349 @@
...
+                {
+                    row.emplace_back(
+                        mkv.first,
+                        std::to_string(mkv.second.value.load(std::memory_order_relaxed)));
+                }
+                keys.push_back(name);
+                values.push_back(std::move(row));
+            }
+        }
+
```
banidoru: When a write failure occurs, there's no attempt to retry the failed write. The code logs the error and moves on. The `consecutiveWriteFailures` logic will eventually trigger a reconnect, but the specific updates that failed during this period are lost. A simple retry mechanism (e.g., retry the `m_table->set` call once or twice) for the individual write could improve data fidelity during transient network blips without adding much complexity.

common/componentstats.cpp
```diff
@@ -0,0 +1,349 @@
...
+        if (consecutiveWriteFailures >= kMaxConsecutiveWriteFailures)
+        {
+            SWSS_LOG_WARN("ComponentStats[%s]: %d consecutive write failures; "
+                          "dropping Redis connection and reconnecting",
+                          m_component.c_str(), consecutiveWriteFailures);
+            m_table.reset();
+            m_db.reset();
```
banidoru: The logic to re-establish the connection after failures is good. However, similar to the initial connection loop, it retries on the fixed `m_intervalSec`. Implementing an exponential backoff here could provide a more robust and responsive recovery behavior when Redis is down for an extended period.

banidoru (changes requested):
* The separation of `getOrCreateEntity` and `getOrCreateCounter` leads to inefficient double-locking on the hot path for new metrics.
* Memory ordering for counter increments can be relaxed for better performance.
* Connection and reconnection logic for Redis could be more robust by using an exponential backoff strategy instead of a fixed interval.
* Individual Redis write failures are not retried, which could lead to data loss during transient network issues.

common/componentstats.cpp
```diff
@@ -0,0 +1,349 @@
...
+    {
+        return;
+    }
+    auto& e = getOrCreateEntity(entity);
+    auto& c = getOrCreateCounter(e, metric);
+    c.value.fetch_add(n, std::memory_order_relaxed);
+    // Release ordering pairs with the acquire in writerThread() so that counter
+    // writes become visible before the version bump.
+    e.version.fetch_add(1, std::memory_order_release);
+}
```
banidoru: `setValue` has a data race with concurrent `increment`: thread A calls `c.value.store(value)` at line 163, then thread B's `fetch_add` lands, then thread A bumps `e.version`. The writer sees the new version but reads the post-increment value, which is correct — however, a *second* `setValue` call can silently discard the increment that happened between the `store` and the version bump, because the caller of `setValue` intended to set an absolute value but didn't account for concurrent increments.

This is inherent to mixing gauge-style `setValue` and counter-style `increment` on the same metric from different threads. At minimum, document this caveat prominently. If you want stronger guarantees, `setValue` should be done under the same lock or use a CAS loop.

common/componentstats.cpp
```diff
@@ -0,0 +1,349 @@
...
+            {
+                const std::string& name = entry.first;
+                const EntityStats& stats = entry.second;
+
+                // Acquire pairs with release in increment()/setValue(),
+                // guaranteeing we see all counter updates performed before
+                // the version bump we're about to read.
+                uint64_t curVer = stats.version.load(std::memory_order_acquire);
+
+                auto vit = lastVersions.find(name);
```
banidoru: Entities written to Redis (`m_table->set(...)`) are never cleaned up. If a container creates transient entity names (e.g., session IDs, task IDs), stale keys will accumulate in `COUNTERS_DB` indefinitely. Consider:
- Adding a `removeEntity()` API that erases from `m_entities` and calls `m_table->del()`.
- Or documenting that entity names must be stable/bounded and callers are responsible for cleanup.

tests/componentstats_ut.cpp
```diff
@@ -0,0 +1,245 @@
...
+} // namespace
+
+// These tests focus on the in-process behavior (no Redis required).
+// The writer thread's DB connect attempts will fail and be retried silently;
+// that does not affect the getters below, which read from the in-memory map.
+
+TEST(ComponentStats, IncrementBasic)
+{
+    auto s = ComponentStats::create("UT1");
+    s->increment("entity_a", "SET");
```
banidoru: Each test creates a `ComponentStats` singleton keyed by component name, and these persist for the entire process lifetime (the registry holds `weak_ptr`, but the `shared_ptr` returned by `create()` is held until the test scope ends — and the writer thread may still be running). With unique names like `UT1`, `UT2`, etc., this works today, but:
1. Each test leaks a writer thread until `stop()` or `shared_ptr` destruction — many tests = many threads alive simultaneously.
2. If anyone reuses a component name across tests, they'll silently share state.

Consider calling `s->stop()` at the end of each in-memory test to clean up the writer thread promptly.

banidoru (commented):
- **Overall: solid design** — deferred DB connect, version-based dirty tracking, and stable `std::map` references are well thought out.
- **`setValue` race with `increment`**: concurrent `store` + version bump is not atomic relative to `fetch_add` in `increment()` — an increment can land between `store` and the version bump, getting silently overwritten on the next `setValue` call from another thread. Needs `compare_exchange` loop or single-lock path for `setValue`.
- **No Redis key cleanup**: entities written to Redis are never deleted. Long-running containers with transient entity names will leak stale keys in COUNTERS_DB indefinitely.
- **Test singleton leakage**: `ComponentStats::create` returns process-lifetime singletons, so test order matters. Earlier tests' instances leak into later tests, risking flaky failures if component names collide.
- **Agree with R1/R2**: the lock-free claim vs two-mutex-acquisition reality, missing final flush on shutdown, and registry leak are the highest-priority fixes.

common/componentstats.cpp
yutongzhang-microsoft: Fixed in b21b74e, partially. `setValue()` now takes `m_structMutex` in exclusive mode for the store + version bump, so its own observation is atomic and concurrent `setValue()` calls are serialized. However — as you point out — an `increment()` whose `shared_lock` lookup races ahead of `setValue`'s exclusive lock can still land its `fetch_add` outside the lock, so the writer may observe `value + n` rather than `value`. Fully serializing that would require dragging every counter mutation under the lock, which defeats the lock-free hot path. The header doc now explicitly states this caveat: mixing gauge `setValue` and counter `increment` on the same metric from concurrent threads is best-effort, and callers should pick one style per metric for deterministic gauge semantics. New test `ConcurrentIncrementAndSetValueDoNotCrash` exercises the mixed path as a UB / safety check.

common/componentstats.cpp
yutongzhang-microsoft: Fixed in b21b74e. The ctor no longer clamps; `create()` is now the single chokepoint that ensures `effectiveInterval >= 1`, and the ctor's member-initializer just stores what it was passed.

common/componentstats.cpp
yutongzhang-microsoft: Fixed in b21b74e. `create()` now calls `pruneExpiredRegistryLocked()` (under `registryMutex()`) before inserting a new slot, so expired `weak_ptr` entries are reaped on each `create()` call. New test `RegistryPrunesExpiredInstances` releases one instance, creates 10 unrelated ones to force a prune cycle, then re-creates the original name and asserts it gets a fresh pointer.

tests/componentstats_ut.cpp
yutongzhang-microsoft: Fixed in b21b74e. Every test that constructs a `ComponentStats` instance now calls `s->stop()` at the end, so the writer thread is joined and the singleton is released before the next test runs. Distinct component names (`UT1` … `UT_*`) are still used so that even if a release races with the registry prune, two tests cannot accidentally share state.

common/componentstats.cpp
yutongzhang-microsoft: Acked. On SONiC the writer connects to the local in-container Redis, which is typically ready within seconds of container start, so a fixed 1 s retry has been adequate in practice. Switching to exponential backoff is a reasonable hardening but I'd prefer to defer it to a follow-up so this PR stays focused on the correctness fixes (lock-free path, final flush, gauge race, registry leak). Filed mentally as a follow-up; will open a tracking issue if you prefer.

common/componentstats.cpp
yutongzhang-microsoft: Acked. The current behavior is: a write failure logs and is skipped, and after `kMaxConsecutiveWriteFailures=3` failures the writer drops the connection and reconnects (`lastVersions.clear()` ensures every entity gets re-flushed against the new connection with its latest cumulative value). So no monotonic data point is permanently lost — the next successful flush carries the up-to-date value. Adding per-call retry would help only for very-narrow transient blips and would complicate the failure-handling path; I'd prefer to leave the current "log + reconnect" policy as is. Happy to add per-call retry if you feel strongly.

common/componentstats.cpp
yutongzhang-microsoft: Fixed in b21b74e. The flush logic is now factored into a private `flushDirty()` method, and `writerThread()` calls it one more time after the main loop exits — before joining — so any counter updates that landed after the last periodic flush are persisted to Redis on `stop()`. Best-effort: if Redis is already gone the final flush logs and moves on rather than blocking shutdown. New test `FinalFlushOnShutdown` covers this: uses `intervalSec=60`, dirties an entity, calls `stop()` immediately, and verifies the value lands in Redis.

common/componentstats.cpp
yutongzhang-microsoft: Fixed in b21b74e. `getOrCreateEntity` + `getOrCreateCounter` have been merged into a single `lookupOrCreate()` that takes the lock at most once. The structural mutex is now `std::shared_mutex`, so the lookup uses a `shared_lock` (concurrent readers proceed in parallel) and only the first-time insert path upgrades to `unique_lock`.

common/componentstats.cpp
yutongzhang-microsoft: Resolved by the same change: `lookupOrCreate()` performs the entity lookup, the metric lookup, and (if needed) both inserts under a single lock acquisition. The lock is never released between the two halves, so there is no longer a window where another thread could race with an erase. As before, no erase API is exposed; if one is ever added, the layered structure stays consistent under a single `unique_lock`.

common/componentstats.h
yutongzhang-microsoft: Fixed in b21b74e. The hot path now goes through a single `lookupOrCreate(entity, metric)` that takes `std::shared_mutex` in `shared_lock` mode for the read-only lookup; concurrent `increment()` callers no longer serialize against each other. Only the first-ever insertion for a new `(entity, metric)` pair upgrades to an exclusive lock. The header doc and the design-goals section are updated to describe this accurately.

common/componentstats.cpp
yutongzhang-microsoft: Already the case — the existing code uses `std::memory_order_relaxed` on the counter `fetch_add` and `std::memory_order_release` on the version bump, exactly as you describe. The b21b74e refactor preserves that ordering; no change needed here.

common/componentstats.h
yutongzhang-microsoft: Done in b21b74e — both methods are now a single `lookupOrCreate(entity, metric)` taking at most one lock acquisition (`shared_lock` on the hit path, `unique_lock` on the slow path that has to insert).

common/componentstats.cpp
yutongzhang-microsoft: Same answer as the initial-connect case: deferring exponential backoff to a follow-up so this PR stays scoped to the correctness fixes. The reconnect path already triggers only after 3 consecutive failures, so the polling rate during outage is already bounded.

common/componentstats.cpp
yutongzhang-microsoft: Acked, declined for v1. The HLD (sonic-net/SONiC#2312, Reporting HLD §7.1 "No deletion in v1" and §12) explicitly states that entity GC is out of scope for the initial revision: stale rows live until the producer container restarts, and callers are expected to use stable / bounded entity names. The expected first consumer (SwssStats) has bounded entity cardinality = number of orchagent tables (~tens), so this is safe in practice. Adding `removeEntity()` is reserved for a future revision per the HLD's Open/Action items.

yutongzhang-microsoft: Following up on this — wanted to lay out the actual design reason rather than just pointing at the HLD, since the HLD's "out of scope for v1" is a *result* of the reasoning below.

**Why `removeEntity()` is non-trivial here**

The lock-free hot path depends on `std::map` reference stability: `lookupOrCreate()` returns a `Counter&` / `EntityStats&`, releases the lock, and the caller then mutates atomics on those references concurrently:

``cpp
auto ref = lookupOrCreate(entity, metric);   // shared_lock taken + released
ref.counter.value.fetch_add(...);            // lock-free atomic
ref.entity.version.fetch_add(...);
``

If we expose `removeEntity()`, thread A could hold a `Counter&` while thread B erases the entity from the map — classic use-after-free.

Making erase safe requires one of:

1. `shared_ptr<Counter>` per metric: adds an atomic refcount op on every hot-path access (cache-line ping-pong is the dominant cost in busy counter workloads).
2. RCU / hazard pointers: significant code complexity and a quiescent-state mechanism.
3. Take `unique_lock` for every mutation: kills the lock-free path entirely.

**Why no current consumer needs it**

All v1 / planned consumers have topology-bounded entity cardinality:

| Consumer | Entity | Cardinality |
|---|---|---|
| SwssStats (sonic-swss#4516) | orchagent table name | ~tens, stable for container lifetime |
| gnmi (future) | gNMI subscription path | bounded by subscriber config |
| bmp (future) | BMP peer address | bounded by BGP topology |

None of them mint short-lived entity names (no per-RPC IDs, per-session UUIDs, etc.), so growth is bounded and `removeEntity()` would not actually be exercised.

**v1's "GC" mechanism**

The Redis schema has no TTL; keys live for the container's process lifetime and are recreated on restart (HLD §10). Stale-key lifetime is bounded by container restart frequency, which is sufficient for the planned consumers.

**When v2 would do this**

If a future consumer needs truly short-lived entities (RPC IDs, session-scoped objects), the right move is to redesign the storage layer at the same time (e.g. switch to `shared_ptr<Counter>` or RCU) and document the new concurrency model. That's a non-trivial change deserving its own HLD revision, not a follow-up patch on top of this one.

So this is a deliberate design trade-off, not just deferred work. Happy to add the `shared_ptr` variant if you think the SwssStats / future consumers will need it sooner — let me know if you want me to spike that.

PR #1177, opened 2026-04-16
[dash] Add COUNTERS_ENI_OID_MAP table name to schema.h

theasianpianist:
Define a table name to hold the mapping for OID to ENI name (reverse mapping of the existing COUNTERS_ENI_NAME_MAP)

PR #1188, opened 2026-05-14
swsscommon: expose batched ZmqProducerStateTable APIs to SWIG

hdwhdw:
## Why

The batched overloads of `ZmqProducerStateTable::set/del/send` that take a `std::vector<KeyOpFieldsValuesTuple>` are unreachable from the Go bindings today (and Python is in the same boat). SWIG happily generates the C trampolines and Go methods (e.g. `SwigcptrZmqProducerStateTable.Set__SWIG_3`), but the parameter type `std::vector<KeyOpFieldsValuesTuple>` is not `%template`'d and SWIG cannot synthesize a default constructor for the opaque underlying tuple element, so callers have no way to build the argument — the type exposes only `Swigcptr()`.

Verified empirically in a clean trixie + SWIG 4.3 + Go 1.24 environment:

| Type | `%template`? | Constructor | `Add/Get/Size` |
|------|---------------|-------------|------------------|
| `FieldValuePairs` (`vector<FieldValueTuple>`)              | yes | ✅ | ✅ |
| `VectorString` (`vector<string>`)                           | yes | ✅ | ✅ |
| `KeyOpFieldsValuesQueue` (`deque<KeyOpFieldsValuesTuple>`)  | yes | ✅ | ✅ |
| `vector<KeyOpFieldsValuesTuple>` (parameter of batched `set`/`send`) | **no** | ❌ | ❌ only `Swigcptr()` |

Adding a single `%template` is not enough: even with the list templated, the element `std::tuple<…>` itself has no usable constructor — `std_vector.i` does not synthesize constructors for opaque tuple elements (`std_deque.i` is more permissive, which is why the existing `KeyOpFieldsValuesQueue` works). So the cleanest fix is a small inline shim layer.

Batched `del(vector<string>)` already works today because `VectorString` is templated.

## What

Add three `%inline` helpers in `pyext/swsscommon.i` (also picked up by the goext build via the symlink):

- `zmqProducerBatchedSet(p, keys, fvss)` — all entries get op=SET
- `zmqProducerBatchedDel(p, keys)` — all entries get op=DEL
- `zmqProducerBatchedSend(p, keys, ops, fvss)` — caller supplies per-entry op for mixed batches

Each helper takes parallel vectors of already-wrapped types (`VectorString` and `FieldValuePairsList`) and assembles the `vector<KeyOpFieldsValuesTuple>` on the C++ side before forwarding to the existing batched API. This mirrors the parallel-vector pattern `Table::pops` already uses. Size mismatches throw `std::invalid_argument`, which the existing SWIG exception bridge surfaces to callers as a recoverable panic.

## Why we want this now

This unblocks bulk gNMI Set in sonic-gnmi (sonic-net/sonic-buildimage#27250 — bulk DASH route programming). Today `MixedDbClient.handleTableData` emits one ZMQ message per key; with these helpers it can ship N keys in a single message. Follow-up PR in sonic-gnmi to land afterwards.

## Tests

- New Go tests in `goext/swsscommon_test.go` (`TestZmqProducerBatchedSet/Del/Send` and a `SetSizeMismatch` negative case) drive each helper through a ZmqConsumerStateTable and verify the entries arrive on the consumer side.
- Per-entry semantic correctness is already covered by the existing C++ tests in `tests/zmq_state_ut.cpp`; the Go tests exist to prove the SWIG glue is reachable from Go.
- Verified locally inside `sonic-slave-trixie`:
  - `./configure --disable-yangmodules && make -j && make install` — clean build, no warnings introduced
  - `./tests/tests --gtest_filter='ZmqConsumerStateTable*'` — all 3 ZMQ unit tests pass
  - `cd goext && go test -v -run TestZmqProducer` — all 4 new tests pass

## Related

- sonic-net/sonic-buildimage#27250

### DIFF
--- a/goext/swsscommon_test.go
+++ b/goext/swsscommon_test.go
@@ -1,9 +1,11 @@
 package swsscommon
 
 import (
-	"testing"
+	"fmt"
 	"os"
 	"path/filepath"
+	"testing"
+	"time"
 )
 
 func TestSonicDBConfig(t *testing.T) {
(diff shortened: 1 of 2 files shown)


prabhataravind:
@hdwhdw could you check PR checker failures?

hdwhdw:
/azpw run Azure.sonic-swss-common

Re-running to confirm the `Build amd64` failure on a Rust ZMQ reconnect test (`zmq_consumer_producer_state_tables_sync_api_connect_late_reconnect`, sync.rs:223 expected `Data`, got `Timeout` within 2 s) is a flake unrelated to this PR. This PR only touches `pyext/swsscommon.i` — the Rust crates wrap the C++ API directly and don't go through SWIG, so the changed surface cannot affect this test. The `Build amd64/ubuntu-22.04` job passed.

hdwhdw:
Re-run results: the previous `Build amd64` Rust late-reconnect flake cleared (now PASS), but `Test vstest` fails on `test_inband_intf_mgmt_vrf.py::TestInbandInterface::test_InbandIntf[PortChannel5]` and `[Loopback1]` (line 128: `assert status == True` after `INTF_TABLE` lookup) — `Ethernet4` and `Vlan100` parametrisations pass.

This is independent of this PR. The change is a pure SWIG `%inline` addition in `pyext/swsscommon.i` — no existing C++, Python, or Go symbol is modified; only three new exports (`zmqProducerBatchedSet/Del/Send`) are added. The `test_InbandIntf` test uses `swsscommon.Table('INTF_TABLE')` and doesn't touch `ZmqProducerStateTable` at all. The flaky plugin retried 3 times and produced the same two failures, so it's likely an underlying intfmgr/portsorch timing issue that surfaces for these two interface kinds on the Azure DVS runner.

Cross-check: unrelated PR #1191 (`LruDedup queue policy for Notification Consumer`) also has a failing vstest run with different symptoms (`docker.errors.APIError: 500 Server Error` in fixture setup), and the same vstest job passes on most recent merged PRs (#1170, #1177). The vstest pipeline is unstable on the Azure runners this week.

/azpw run Azure.sonic-swss-common

prabhataravind (approved):
lgtm. @qiluo-msft to sign-off/merge as well.

prabhataravind:
@prsunny for viz as this is needed for 202605.

qiluo-msft (approved):

PR #1195, opened 2026-05-22
Add link event damping support

sivat6:
Modified the LUA script to include link damping configs CLIs.

Errors were logged without this modification and confirmed that errors are not showing after the fix.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: sonic
 Maintainer: Shuotian Cheng <shuche@microsoft.com>
 Section: net
 Priority: optional
-Build-Depends: dh-exec (>=0.3), debhelper (>= 12), autotools-dev, libboost-dev | libboost1.71-dev | libboost1.83-dev, libhiredis-dev, libgtest-dev, libgmock-dev, swig, nlohmann-json3-dev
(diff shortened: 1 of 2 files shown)


debian/control
```diff
@@ -2,7 +2,7 @@ Source: sonic
 Maintainer: Shuotian Cheng <shuche@microsoft.com>
 Section: net
 Priority: optional
-Build-Depends: dh-exec (>=0.3), debhelper (>= 12), autotools-dev, libboost-dev | libboost1.71-dev | libboost1.83-dev, libhiredis-dev, libgtest-dev, libgmock-dev, swig, nlohmann-json3-dev
+Build-Depends: dh-exec (>=0.3), debhelper (>= 12), autotools-dev, libboost-dev | libboost1.71-dev | libboost1.83-dev, libhiredis-dev, libgtest-dev, libgmock-dev, swig, nlohmann-json3-dev, libyang-dev, python3-libyang
```
prsunny: Why do we need this change here? Can you limit this PR to lua script change?

sivat6: There were some build errors and this change was added to fix that. I removed these changes now.

prsunny (approved):

qiluo-msft (approved):

PR #1196, opened 2026-05-22
zmq: prevent use-after-free of ZmqConsumerStateTable handler

bhouse-nexthop:
### Description of PR

Summary:
`ZmqServer` registers `ZmqConsumerStateTable` as a message handler in `m_HandlerMap` at consumer construction, but there is no symmetric unregister path and no `ZmqConsumerStateTable` destructor. When a consumer goes out of scope (its `m_selectableEvent` is destroyed and its eventfd closed) while the server's `mqPollThread` is still running and partway through dispatching a message to it, `m_selectableEvent.notify()` does `write()` on a closed fd, `SelectableEvent::notify()` throws `runtime_error("write failed")` out of the poll thread, and `std::terminate` aborts the process.

This is the natural declaration order in test code — the `ZmqServer` has to exist before it can be passed by reference into a `ZmqConsumerStateTable` constructor, so the server is declared first and (per C++ reverse-order destruction) the consumer destructs first. Reviewers should start with `common/zmqserver.cpp` for the locking changes, then `common/zmqconsumerstatetable.cpp` for the new destructor.

The crash is intermittent but reproducible in CI. Two recent unrelated examples:
- Build [#1078848](https://dev.azure.com/mssonic/build/_build/results?buildId=1078848) — scheduled run on the `202511` release branch (2026-04-04). [Unit-test log](https://dev.azure.com/mssonic/be1b070f-be15-4154-aade-b1d3bfb17054/_apis/build/builds/1078848/logs/33).
- Build [#1120234](https://dev.azure.com/mssonic/build/_build/results?buildId=1120234) — PR #973 (libyang3 port, unrelated change) (2026-05-22). [Unit-test log](https://dev.azure.com/mssonic/be1b070f-be15-4154-aade-b1d3bfb17054/_apis/build/builds/1120234/logs/33).

Both logs show the identical signature:
```
[ RUN      ] ZmqServerLazzyBind.test
terminate called after throwing an instance of 'std::runtime_error'
  what():  write failed
... Aborted (core dumped)
Bash exited with code '134'
```

The string `write failed` is only produced from one place in the tree (`SelectableEvent::notify()` on a short `write()` of the eventfd), pinning the cause precisely. gtest attributes it to `ZmqServerLazzyBind.test` because `[ RUN ]` is printed before the test body executes; the actual fault is the in-flight handler callback writing to a freshly-closed eventfd.

No dependencies.

Fixes # (no issue filed)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation update
- [ ] Test improvement

#### What is the motivation for this PR?

Eliminate an intermittent CI crash and the latent use-after-free behind it. Any caller that destroys a `ZmqConsumerStateTable` while its `ZmqServer` is still alive can trip the same race in production, not just in tests.

#### How did you do it?

* Add `ZmqServer::removeMessageHandler(dbName, tableName)`, guarded by a new `m_handlerMapMutex`.
* Hold that same mutex in `ZmqServer::handleReceivedData` across both the handler lookup *and* the dispatch into `handler->handleReceivedData()`. This makes `removeMessageHandler()` block until any in-flight callback into the handler being removed has returned, so the caller can safely destroy the handler object as soon as `removeMessageHandler()` returns.
* Add a `ZmqConsumerStateTable` destructor that calls `removeMessageHandler()` in its body, i.e. before any data members (notably `m_selectableEvent`) begin tearing down.
* Take `m_handlerMapMutex` in `registerMessageHandler()` too, for symmetry.

Caveat documented on the mutex declaration: handler callbacks now run with `m_handlerMapMutex` held and must not themselves call `register`/`removeMessageHandler` (would self-deadlock). `ZmqConsumerStateTable::handleReceivedData` only enqueues and notifies, so the restriction is harmless in tree.

#### How did you verify/test it?

* Built locally on bookworm; existing unit tests pass.
* Verified the lock-ordering invariant: the server's `m_handlerMapMutex` is always the outer lock; `ZmqConsumerStateTable::m_receivedQueueMutex` is only ever the inner lock. No caller takes them in the opposite order.
* Inspected the only `ZmqMessageHandler` subclass / caller of `registerMessageHandler` (`ZmqConsumerStateTable`) to confirm the callback respects the no-reentrancy contract.
* Read upstream CI logs for the crash signature to confirm the fix targets the right root cause (links above).

#### Any platform specific information?

None.

### Documentation

N/A — internal bug fix, no public API change. The new `ZmqServer::removeMessageHandler()` entry point is intended for use by handler implementations' own destructors (mirroring the existing `registerMessageHandler()` symmetry).

### DIFF
--- a/common/zmqconsumerstatetable.cpp
+++ b/common/zmqconsumerstatetable.cpp
@@ -21,7 +21,8 @@ ZmqConsumerStateTable::ZmqConsumerStateTable(DBConnector *db, const std::string 
     : Selectable(pri)
     , TableBase(tableName, TableBase::getTableSeparator(db->getDbId()))
     , m_db(db)
-    , m_zmqServer(zmqServer)
+    , m_dbName(db->getDbName())
+    , m_handlerRegistry(zmqServer.getHandlerRegistry())
 {
     if (popBatchSize > 0)
     {
@@ -44,9 +45,20 @@ ZmqConsumerStateTable::ZmqConsumerStateTable(DBConnector *db, const std::string 
         m_asyncDBUpdater = nullptr;
     }
 
-    m_zmqServer.registerMessageHandler(m_db->getDbName(), tableName, this);
+    m_handlerRegistry->registerHandler(m_dbName, tableName, this);
 
     SWSS_LOG_DEBUG("ZmqConsumerStateTable ctor tableName: %s", tableName.c_str());
+}
+
+ZmqConsumerStateTable::~ZmqConsumerStateTable()
+{
+    // Detach from the registry before any of our members (notably the
+    // SelectableEvent, whose eventfd we'd write to from handleReceivedData)
+    // are destroyed. removeHandler() blocks until any in-flight dispatch
+    // into us returns. The registry is co-owned with the ZmqServer that
+    // created us, so this is safe even if that ZmqServer has already been
+    // destroyed — only the shared registry is touched.
+    m_handlerRegistry->removeHandler(m_dbName, getTableName());
 }
 
 void ZmqConsumerStateTable::handleReceivedData(const std::vector<std::shared_ptr<KeyOpFieldsValuesTuple>> &kcos)
--- a/common/zmqconsumerstatetable.h
+++ b/common/zmqconsumerstatetable.h
@@ -21,6 +21,8 @@ public:
 
     ZmqConsumerStateTable(DBConnector *db, const std::string &tableName, ZmqServer &zmqServer, int popBatchSize = DEFAULT_POP_BATCH_SIZE, int pri = 0, bool dbPersistence = false);
 
+    ~ZmqConsumerStateTable() override;
+
     /* Get multiple pop elements */
     void pops(std::deque<KeyOpFieldsValuesTuple> &vkco, const std::string &prefix = EMPTY_PREFIX);
 
@@ -81,7 +83,13 @@ private:
 
     DBConnector *m_db;
 
-    ZmqServer& m_zmqServer;
+    // Cached at construction time so the destructor can call
+    // m_handlerRegistry->removeHandler() without dereferencing m_db.
+    std::string m_dbName;
+
+    // Co-owned with the ZmqServer that created us. Lets the destructor
+    // unregister cleanly even if the ZmqServer has already been destroyed.
+    std::shared_ptr<ZmqHandlerRegistry> m_handlerRegistry;
 
     std::unique_ptr<AsyncDBUpdater> m_asyncDBUpdater;
 
--- a/common/zmqserver.cpp
+++ b/common/zmqserver.cpp
@@ -12,6 +12,75 @@ using namespace std;
 
 namespace swss {
 
+void ZmqHandlerRegistry::registerHandler(
+    const std::string& dbName,
+    const std::string& tableName,
+    ZmqMessageHandler* handler)
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+
+    auto dbResult = m_handlers.insert(make_pair(dbName, map<string, ZmqMessageHandler*>()));
+    if (dbResult.second) {
+        SWSS_LOG_DEBUG("ZmqHandlerRegistry add mapping for db: %s", dbName.c_str());
+    }
+
+    auto tableResult = dbResult.first->second.insert(make_pair(tableName, handler));
+    if (tableResult.second) {
+        SWSS_LOG_DEBUG("ZmqHandlerRegistry register handler for db: %s, table: %s",
+                       dbName.c_str(), tableName.c_str());
+    }
+}
+
+void ZmqHandlerRegistry::removeHandler(
+    const std::string& dbName,
+    const std::string& tableName)
+{
+    // Take the same mutex that dispatch() holds across the callback. Once we
+    // acquire it, no callback into this (dbName, tableName) handler is in
+    // flight and no further one can start — making it safe for the caller to
+    // destroy the handler object after removeHandler() returns.
+    std::lock_guard<std::mutex> lock(m_mutex);
+
+    auto dbIter = m_handlers.find(dbName);
+    if (dbIter == m_handlers.end()) {
+        return;
+    }
+
+    dbIter->second.erase(tableName);
+    if (dbIter->second.empty()) {
+        m_handlers.erase(dbIter);
+    }
+
+    SWSS_LOG_DEBUG("ZmqHandlerRegistry removed handler for db: %s, table: %s",
+                   dbName.c_str(), tableName.c_str());
+}
+
+void ZmqHandlerRegistry::dispatch(
+    const std::string& dbName,
+    const std::string& tableName,
+    const std::vector<std::shared_ptr<KeyOpFieldsValuesTuple>>& kcos)
+{
+    // Hold the mutex for the duration of the callback. Concurrent
+    // removeHandler() on this (dbName, tableName) blocks until we return,
+    // so the handler cannot be destroyed mid-call.
+    std::lock_guard<std::mutex> lock(m_mutex);
+
+    auto dbIter = m_handlers.find(dbName);
+    if (dbIter == m_handlers.end()) {
+        SWSS_LOG_DEBUG("ZmqHandlerRegistry can't find any handler for db: %s", dbName.c_str());
+        return;
+    }
+
+    auto tableIter = dbIter->second.find(tableName);
+    if (tableIter == dbIter->second.end()) {
+        SWSS_LOG_DEBUG("ZmqHandlerRegistry can't find handler for db: %s, table: %s",
+                       dbName.c_str(), tableName.c_str());
+        return;
+    }
+
+    tableIter->second->handleReceivedData(kcos);
+}
+
 ZmqServer::ZmqServer(const std::string& endpoint)
     : ZmqServer(endpoint, "", false, false)
 {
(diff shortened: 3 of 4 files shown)


bhouse-nexthop:
Amended to also fix the Rust binding's `ZmqServer::Drop` order — caught by [build #1120319](https://dev.azure.com/mssonic/build/_build/results?buildId=1120319) on PR #973 after that PR picked up this patch.

The Rust `ZmqServer` holds `Arc<DropGuard>` references to each registered `ZmqConsumerStateTable` in `message_handler_guards`, originally to keep them alive past the server's drop. The previous `Drop::drop` order was:

1. `SWSSZmqServer_free(self.ptr)` — frees the C++ `ZmqServer`.
2. Rust then drops `self.message_handler_guards`, which calls `SWSSZmqConsumerStateTable_free` on each remaining consumer.

Before this PR's C++ change, step 2 was harmless (consumer's destructor did nothing relevant to the server). After the C++ change, `~ZmqConsumerStateTable()` calls `m_zmqServer.removeMessageHandler(...)` — and `m_zmqServer` now points to the freed C++ object from step 1. Locking that object's already-destroyed mutex corrupts the heap, and glibc's tcache double-free detector trips later in the same test binary:

```
free(): double free detected in tcache 2
process didn't exit successfully: target/debug/deps/async-… (signal: 6, SIGABRT)
Bash exited with code '101'.
```

Fix is the obvious flip: clear `message_handler_guards` first (so each consumer's destructor runs against the still-live server and properly unregisters), then `SWSSZmqServer_free`. The old "must be kept alive until the server thread dies" comment described the exact race the C++ side of this PR fixes; the new comment documents the new ordering requirement instead.

bhouse-nexthop:
Amended again — third call site exposed by the new destructor.

[Build #1120334](https://dev.azure.com/mssonic/build/_build/results?buildId=1120334) (PR #973 with previous version of this patch applied) crashed in `c_api.ZmqConsumerProducerStateTable`:

```
NOTICE:- mqPollThread: mqPollThread end
free(): double free detected in tcache 2
... Aborted (core dumped)  Bash exited with code '134'
```

`ZmqServerLazzyBind.test` and all 5 Rust async tests passed cleanly, confirming the C++ and Rust changes from the earlier amends are doing their job. The new crash is the same bug class in `tests/c_api_ut.cpp` — the test was freeing `SWSSZmqServer` *before* `SWSSZmqConsumerStateTable`:

```c
SWSSZmqServer_free(srv);
SWSSZmqProducerStateTable_free(pst);
SWSSZmqConsumerStateTable_free(cst);   // ← consumer destructor now hits freed server
```

The log shows `mqPollThread end` (from `~ZmqServer`) immediately followed by the double-free detection — the consumer's destructor reached into the just-freed server's handler-map mutex.

Same fix as the Rust binding: swap so the consumer is freed first. Updated commit message to spell out the new lifetime contract (handlers must outlive — and be torn down before — their ZmqServer) and list both fixed call sites.

I should have caught both of these when adding the destructor; both were latent ordering bugs that the no-op destructor was masking.

bhouse-nexthop:
Reworked to remove the destruction-order requirement entirely.

Per discussion: the previous revision changed the de facto contract from "consumer must be destroyed before server, *or you sometimes crash*" to "consumer must be destroyed before server, *or you reliably UAF*" — better than before, but still order-dependent. That's not a contract we can impose on existing consumers (sonic-swss orchagent, sonic-gnmi Go bindings, sonic-dash-ha Rust crate, downstream out-of-tree code).

New design: pull the `(db, table) → handler*` map and its mutex out of `ZmqServer` into a small `ZmqHandlerRegistry`. The server and every registered handler co-own a `std::shared_ptr<ZmqHandlerRegistry>`. The locking discipline (mutex held across dispatch so `removeHandler()` blocks for in-flight callbacks) is unchanged — it just lives in the registry now.

Two consequences fall out:

1. Whichever of `ZmqServer` or `ZmqConsumerStateTable` is destroyed last leaves a live registry behind for the other's destructor to use. **Neither needs to outlive the other.** The consumer's destructor only touches the shared registry — never the `ZmqServer` it was created with, never the `DBConnector*` (the db name is cached as `std::string` at construction).

2. Server destruction still joins the poll thread before the registry's refcount drops, so dispatch can never see a partially-destroyed registry.

The two ordering "fixes" I added in previous revisions are now backed out — they'd just hide the property:
- `tests/c_api_ut.cpp` is back to freeing the server first, consumer second.
- `crates/swss-common/src/types/zmqserver.rs` `Drop` is back to freeing the C++ server first, then dropping the consumer guards.

Both should now run cleanly under the new design — that's exactly the order-independence property under test.

Public API surface is unchanged in signature: `registerMessageHandler` and the new `removeMessageHandler` still live on `ZmqServer` and just forward to the registry. The only new symbol is the internal `ZmqServer::getHandlerRegistry()` accessor returning the `shared_ptr`, intended for `ZmqMessageHandler` subclasses (currently only `ZmqConsumerStateTable`).

Force-pushed as `c62cdbf`.

bhouse-nexthop:
[Build #1120397](https://dev.azure.com/mssonic/build/_build/results?buildId=1120397) caught a downstream-mock breakage I introduced when cleaning up unused declarations:

```
fake_zmqserver.cpp:20:20: error: no declaration matches
    'swss::ZmqMessageHandler* swss::ZmqServer::findMessageHandler(std::string, std::string)'
```

`sonic-swss`'s `orchagent/p4orch/tests/fake_zmqserver.cpp` is a link-time mock — it provides stub bodies for every `ZmqServer` method (including `private` ones) so tests can compile against the real header but link against a no-op `ZmqServer`. By dropping the `findMessageHandler` declaration in `zmqserver.h`, I broke that mock's compile.

Also caught a latent runtime issue while looking at this: that same fake's stub constructor only initializes `m_endpoint` — it doesn't initialize my new `m_registry` member. A real `ZmqConsumerStateTable` paired with a faked `ZmqServer` would then call `zmqServer.getHandlerRegistry()` and null-deref the empty `shared_ptr` inside `registerHandler()`.

Fixed both:

1. Restored a private `findMessageHandler` declaration with a no-op body that returns `nullptr`. The internal dispatch path no longer calls it (lookup lives in `ZmqHandlerRegistry::dispatch` now), but the symbol is retained so external mocks continue to link. Documented in a comment.

2. Added a default in-class initializer `m_registry = std::make_shared<ZmqHandlerRegistry>()`. Stub constructors that don't initialize this member in their init list now get a valid registry automatically. The real constructor still sets it explicitly (which wins over the in-class init).

Force-pushed as `4f6fd89`. Pure C++ change still — `tests/c_api_ut.cpp` and `crates/swss-common/src/types/zmqserver.rs` remain bit-identical to master.

securely1g:
## Review

**Problem:** Classic C++ destruction-order race. `ZmqServer` is declared before `ZmqConsumerStateTable`, so the consumer destructs first. But there's no destructor and no unregister — the server's `mqPollThread` can dispatch to a dead consumer, call `m_selectableEvent.notify()` on a closed eventfd, throw `runtime_error`, and `std::terminate` kills the process.

**Solution:** `ZmqHandlerRegistry` — a shared, mutex-protected handler map co-owned (`shared_ptr`) by both `ZmqServer` and each `ZmqConsumerStateTable`. The consumer's new destructor calls `removeHandler()` which blocks until any in-flight dispatch completes, guaranteeing safe destruction regardless of which object dies first.

---

### Strengths

- **Correct lifetime model** — shared ownership via `shared_ptr<ZmqHandlerRegistry>` eliminates the ordering dependency entirely. Neither party needs to outlive the other.
- **Strong happens-before guarantee** — `removeHandler()` acquires the same mutex that `dispatch()` holds across the callback. After `removeHandler()` returns, no further dispatch can reach the handler.
- **Minimal API surface change** — existing `registerMessageHandler()` still works, new `removeMessageHandler()` is additive, `findMessageHandler()` kept as no-op stub for downstream mock compatibility.
- **Excellent comments** — PR description, code comments, and class-level doc all explain *why*, not just *what*.
- **Cached `m_dbName`** — destructor doesn't dereference `m_db` (which might be gone in complex teardown scenarios).

### Concerns

1. **Global mutex held across handler callback** — `dispatch()` holds `m_mutex` for the entire `handleReceivedData()` call. All ZMQ message processing is serialized even for unrelated tables. If `handleReceivedData()` is slow (e.g., bulk route processing), other tables can't be dispatched. Under high message rates this could be a throughput bottleneck. Consider `shared_mutex` or per-table locking in a follow-up — but the simplicity is acceptable for a safety fix.

2. **Self-deadlock is a silent foot-gun** — If any handler's `handleReceivedData()` calls `registerHandler()` or `removeHandler()`, it deadlocks with no diagnostic. The constraint is documented at the class level (good), but consider a debug-mode assertion or `SWSS_LOG_ERROR` + `abort()` if the mutex is already held by the current thread.

3. **`findMessageHandler()` returns `nullptr` unconditionally** — Any downstream code that *calls* (not overrides) this method now silently gets null. Should be marked `[[deprecated]]` to surface unintended callers during transition.

4. **No regression test for the race** — The crash is "reproducible in CI" but there's no explicit test (e.g., destroy consumer while dispatch is in-flight from another thread, assert no crash). Would be valuable as both a regression guard and contract documentation.

5. **`getHandlerRegistry()` is fully public** — Exposing the internal registry as a `shared_ptr` lets any caller hold a reference indefinitely. Consider `protected` or `friend class ZmqConsumerStateTable` to limit scope.

### Minor

- Trailing whitespace fix on the VRF line is unrelated — fine but worth noting
- `m_HandlerMap` removal is technically an ABI break for any out-of-tree code subclassing `ZmqServer` (unlikely but possible in vendor SAI tests)

---

**Verdict:** Clean, well-reasoned fix for a real crash. The shared-registry + mutex approach is correct and the destruction-order problem is properly solved. The mutex-held-across-callback concern is a throughput trade-off worth acknowledging, but acceptable for a focused safety fix. Approve — would suggest adding a regression test and `[[deprecated]]` on `findMessageHandler()`.

lguohan:
#973 merged

PR #1198, opened 2026-05-25
NotificationConsumer: add pluggable Queue strategy

manish1-arista:
**What I did**

Refactored NotificationConsumer to delegate its internal storage to a pluggable Queue strategy injected via the constructor. Added two implementations:

StandardQueue — plain FIFO, used as the default so existing callers compile and behave unchanged.
DedupQueue — backed by a `std::list + std::unordered_map`, collapses duplicate messages and re-orders the surviving entry to the back of the queue.
DedupQueue also tracks total_push and duplicates_skipped counters.

**Why I did it**

Some notification channels (e.g. FDB) can be flooded with duplicate events under load (MAC move storms), causing the consumer to backlog and process stale state. Until now NotificationConsumer always used a plain std::queue, with no way to apply dedup or any other policy on a per-channel basis. Making the queue strategy injectable lets each caller pick the right policy for its channel without touching NotificationConsumer or affecting other consumers and leaves the door open for additional strategies in the future.

### DIFF
--- a/common/notificationconsumer.cpp
+++ b/common/notificationconsumer.cpp
@@ -8,12 +8,14 @@
 #define REDIS_PUBLISH_MESSAGE_INDEX (2)
 #define REDIS_PUBLISH_MESSAGE_ELEMNTS (3)
 
-swss::NotificationConsumer::NotificationConsumer(swss::DBConnector *db, const std::string &channel, int pri, size_t popBatchSize):
+swss::NotificationConsumer::NotificationConsumer(swss::DBConnector *db, const std::string &channel, int pri,
+                                                 size_t popBatchSize, std::shared_ptr<Queue> queue):
     Selectable(pri),
     POP_BATCH_SIZE(popBatchSize),
     m_db(db),
     m_subscribe(NULL),
-    m_channel(channel)
+    m_channel(channel),
+    m_queue(std::move(queue))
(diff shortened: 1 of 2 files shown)


senthil-nexthop:
@manish1-arista I presented a HLD for an opt-in DeDup queue policy last week in the SONiC Foundation meeting that can be seen here: [LruDedup Queue Policy](https://github.com/sonic-net/SONiC/pull/2334). The HLD has links to the code PRs that implements the design.

It appears there is some duplication of work in the set of PRs you have raised. Can we have a quick meeting to discuss how to move this forward? Thanks.

manish1-arista:
Our FDB Notification consumer queue Dedup approach targeted fdborch specifically, reducing redundant notifications at a single consumer.
We are are able to see that the FDB Notification consumer queue size was not increasing under Mac move storm.

The Other PRs (https://github.com/sonic-net/sonic-swss-common/pull/1191, https://github.com/sonic-net/sonic-swss/pull/4586 and https://github.com/sonic-net/sonic-sairedis/pull/1899) fixes this memory growth for other orch consumer queue also by addressing the problem at the infrastructure layer rather than a single orch, by adding allowlist for each orchagent and Dedup queues for multiple orchangents.

Closing this PR as other PRs had merged

PR #1211, opened 2026-06-19
fix: use fgets return value as loop condition in exec()

xq9mend:
## What

Fix incorrect loop condition in `exec()` in `common/exec.cpp`.

## Why

Using `!feof(pipe)` as a loop condition has two problems:

1. `feof()` only becomes true *after* a read has already hit EOF, so the loop always executes one extra iteration where `fgets()` returns NULL.
2. `feof()` does not detect I/O errors (`ferror`). On a read error, `fgets()` returns NULL but `feof()` remains false, causing an infinite loop.

## How

Drive the loop directly on the `fgets()` return value. This handles both EOF and I/O errors correctly with a single condition, and eliminates the redundant inner `if` guard.

```cpp
// Before
while (!feof(pipe))
{
    if (fgets(buffer.data(), buffsz, pipe) != NULL)
    {
        stdout += buffer.data();
    }
}

// After
while (fgets(buffer.data(), buffsz, pipe) != NULL)
{
    stdout += buffer.data();
}
```

### DIFF
--- a/common/exec.cpp
+++ b/common/exec.cpp
@@ -26,12 +26,9 @@ int exec(const string &cmd, string &stdout)
     }
 
     stdout.clear();
-    while (!feof(pipe))
+    while (fgets(buffer.data(), buffsz, pipe) != NULL)
     {
-        if (fgets(buffer.data(), buffsz, pipe) != NULL)
-        {
-            stdout += buffer.data();
-        }
+        stdout += buffer.data();
     }
 
     int ret;


xq9mend:
The `Test vstest` failure is a systemic CI issue unrelated to this PR. The p4rt test group in sonic-swss/tests/p4rt/ is failing across all sonic-swss-common PRs (exit code 123, zero tests passing) due to a mismatch between the latest sonic-swss master test code and the pre-built docker-sonic-vs image. This affects every open sonic-swss-common PR right now — see #1211, #1212, #1213, #1214 all showing the same failure.

The code change in this PR is correct and unrelated to p4rt.

qiluo-msft (approved):

PR #1214, opened 2026-06-20
perf: eliminate redundant map lookups in swssPrioNotify and swssOutputNotify

xq9mend:
## Why

`swssPrioNotify` and `swssOutputNotify` each perform two map lookups on the success path:
1. `find()` to check if the key exists
2. `at()` to retrieve the value

Both are O(log n) on `std::map`, so the success path pays the cost twice unnecessarily.

## How

Store the iterator returned by `find()` and use `it->second` directly in the else branch, eliminating the second lookup.

## How to verify

Existing logger unit tests pass. Behavior is identical — only the number of map traversals changes.

### DIFF
--- a/common/logger.cpp
+++ b/common/logger.cpp
@@ -78,7 +78,8 @@ void Logger::swssPrioNotify(const std::string& component, const std::string& pri
 {
     auto& logger = getInstance();
 
-    if (priorityStringMap.find(prioStr) == priorityStringMap.end())
+    auto it = priorityStringMap.find(prioStr);
+    if (it == priorityStringMap.end())
     {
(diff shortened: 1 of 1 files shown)


xq9mend:
The `Test vstest` failure is a systemic CI issue unrelated to this PR. The p4rt test group in sonic-swss/tests/p4rt/ is failing across all sonic-swss-common PRs (exit code 123, zero tests passing) due to a mismatch between the latest sonic-swss master test code and the pre-built docker-sonic-vs image. This affects every open sonic-swss-common PR right now — see #1211, #1212, #1213, #1214 all showing the same failure.

The code change in this PR is correct and unrelated to p4rt.

qiluo-msft (approved):

PR #1216, opened 2026-06-23
Revert quickfix for sonic-swss-common builds

bgallagher-nexthop:
Revert the quickfix merged in https://github.com/sonic-net/sonic-swss-common/pull/1206 to fix sonic-swss-common builds now that the long-term fix has merged and had time to take effect in https://github.com/sonic-net/sonic-buildimage/pull/27836.

### DIFF
--- a/.azure-pipelines/build-template.yml
+++ b/.azure-pipelines/build-template.yml
@@ -106,10 +106,6 @@ jobs:
   - script: |
       set -ex
       sudo dpkg -i $(find ./download -name *.deb)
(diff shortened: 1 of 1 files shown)


bgallagher-nexthop:
cc @yxieca 

This is to revert the quick-fix PR you approved in https://github.com/sonic-net/sonic-swss-common/pull/1206, now that the [proper fix](https://github.com/sonic-net/sonic-buildimage/pull/27836) in sonic-buildimage has taken effect.

PR #1218, opened 2026-07-01
fix strncpy truncation; enable FORTIFY_SOURCE=3

xq9mend:
## Why I did it

`zmq_msg_init_size()` allocates a byte buffer sized exactly `s.size()` — no room for a null terminator. The ZMQ message reader (`zmsg_to_map`) constructs a `std::string` using `zmq_msg_size()`, so null termination is irrelevant. `strncpy` is the wrong primitive here; `memcpy` is semantically correct for raw byte blobs and eliminates the FORTIFY_SOURCE=3 provable-truncation diagnostic.

Also upgrade FORTIFY_SOURCE to level 3 in `debian/rules` so all future builds use the stronger compile-time check. dpkg sets =2 by default; the `-U` flag clears it before setting =3.

## How I did it

- `common/events_common.h`: replace `strncpy` with `memcpy` in `map_to_zmsg()`
- `debian/rules`: add `DEB_CPPFLAGS_MAINT_APPEND = -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3`

## How to verify it

Build passes cleanly with `dpkg-buildpackage`. No compiler warnings or errors related to FORTIFY_SOURCE.

### DIFF
--- a/common/events_common.h
+++ b/common/events_common.h
@@ -286,7 +286,7 @@ struct serialization
             rc = zmq_msg_init_size(&msg, s.size());
         }
         if (rc == 0) {
-            strncpy((char *)zmq_msg_data(&msg), s.c_str(), s.size());
+            memcpy((char *)zmq_msg_data(&msg), s.c_str(), s.size());
         }
         return rc;
     }
--- a/debian/rules
+++ b/debian/rules
@@ -9,6 +9,8 @@ include /usr/share/dpkg/default.mk
 
 # see FEATURE AREAS in dpkg-buildflags(1)
 export DEB_BUILD_MAINT_OPTIONS = hardening=+all
+# Upgrade FORTIFY_SOURCE to level 3 (dpkg default is =2).
+export DEB_CPPFLAGS_MAINT_APPEND = -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3
 
 # see ENVIRONMENT in dpkg-buildflags(1)
 # package maintainers to append CFLAGS


xq9mend:
@deepak-singhal0408 @saiarcot895 Could you please review this PR? Thank you!

deepak-singhal0408 (commented):
The title and "How I did it" list a `debian/rules` FORTIFY_SOURCE=3 change that isn't in the diff or in master. Since SONiC squashes on merge, either add the hunk or drop the FORTIFY bullets so the commit matches what landed. The rest of the series included it ([sonic-swss 4723](https://github.com/sonic-net/sonic-swss/pull/4723), [sonic-sairedis 1963](https://github.com/sonic-net/sonic-sairedis/pull/1963), [sonic-stp 94](https://github.com/sonic-net/sonic-stp/pull/94), [sonic-linkmgrd 349](https://github.com/sonic-net/sonic-linkmgrd/pull/349), [sonic-dhcpmon 82](https://github.com/sonic-net/sonic-dhcpmon/pull/82), [sonic-wpa-supplicant 119](https://github.com/sonic-net/sonic-wpa-supplicant/pull/119)), so swss-common would be the one repo left at the dpkg default.

`TEST(events_common, send_recv)` covers this function with NUL-free ASCII only, so it passes with and without the fix. A case with an embedded NUL — like `send_recv_control_character` in the same file — would fail on master and pass here.

deepak-singhal0408 (approved):
LGTM.

Non-blocking: the fix isn't covered by a test. `send_recv` uses NUL-free ASCII so it passes either way, and `send_recv_control_character` never reaches `map_to_zmsg`. A case round-tripping a map value with an embedded NUL through `zmq_message_send`/`zmq_message_read` would fail on master and pass here — worth adding whenever convenient.

PR #1221, opened 2026-07-07
[test][ci] Fix two flaky unit tests + reduce CI flakiness

theasianpianist:
## What / why

The Azure CI for this repo is intermittently red due to two flaky unit tests (they redden `master` too). This PR fixes both, and adds two small CI-robustness tweaks for stages that intermittently fail for infra reasons.

### 1. Fix `Notifications.LruDedupPolicyConstructorAndLabelPropagation` (`tests/ntf_ut.cpp`)
The drain loop stopped at the first `pop`, but the two byte-identical `fdb_event` messages arrive over Redis pub/sub and the second reply may not have been read yet. `LruDedup` collapses them into a single queued item, so the test observed `received == 1` and failed `EXPECT_EQ(received, 2)`. **Fix:** drain until both messages have been received (`received >= 2`) or a timeout, then assert.

### 2. Fix `test_TableOpsMemoryLeak` (`tests/test_redis_ut.py`)
It asserted `< 50000` bytes of RSS growth over 50000 fresh-process ops, but that growth is one-time allocator / client-buffer warmup, not a leak. **Fix:** warm up first, then measure the steady-state second batch — a real per-op leak keeps growing linearly, so post-warmup growth must stay near zero (threshold unchanged).

### 3. Raise the sonic-swss build timeout (90m → 120m)
`BuildSwss` compiles sonic-swss and intermittently exceeds the 90-minute default.

### 4. Free host disk before the sonic-sairedis build stage
`BuildSairedis` runs on a hosted `ubuntu-22.04` agent and intermittently runs out of disk. A best-effort `target: host` step drops big preinstalled toolchains the build never uses and prunes docker images (with `df` before/after).

### 5. Enable flaky VS test retry
Pass `--force-flaky` to VS test so any test failures are retried

## Verification
Reproduced and verified in a `sonic-slave-bookworm` container under 32-worker CPU load (the flakes are load-sensitive):

| Test | Before (master) | After (this PR) |
|------|-----------------|-----------------|
| `LruDedup` ×150 | **135 failed** (`received`=1) | **0 failed** |
| `test_TableOpsMemoryLeak` ×20 | **20 failed** | **0 failed** |

### DIFF
--- a/tests/ntf_ut.cpp
+++ b/tests/ntf_ut.cpp
@@ -330,9 +330,14 @@ TEST(Notifications, LruDedupPolicyConstructorAndLabelPropagation)
     send("fdb_event");
     send("fdb_event");    // byte-identical to first -- LruDedup should collapse
 
-    // Drain so processReply is invoked.
+    // Drain so processReply is invoked. Both messages are delivered over Redis
+    // pub/sub, but the second reply may not have arrived by the time the first
+    // select() returns -- especially under load. LruDedup collapses the two
+    // byte-identical messages into a single queued item, so stopping at the first
+    // pop races the second reply and can observe received==1. Keep draining until
+    // both messages have been received (processReply'd), or we time out.
     size_t pops = 0;
-    for (int i = 0; i < 50 && pops == 0; ++i) {
+    for (int i = 0; i < 100 && nc.getStats().received < 2; ++i) {
         swss::Selectable *sel = nullptr;
         if (s.select(&sel, 100 /* ms */) == swss::Select::TIMEOUT) continue;
         std::deque<swss::KeyOpFieldsValuesTuple> vkco;
--- a/tests/test_redis_ut.py
+++ b/tests/test_redis_ut.py
@@ -873,9 +873,18 @@ def test_TableOpsMemoryLeak():
     t = swsscommon.Table(app_db, "TABLE")
     long_data = "x" * 100
     fvs = swsscommon.FieldValuePairs([(long_data, long_data)])
+
+    def run_ops():
+        for _ in range(OP_COUNT):
+            t.set("long_data", fvs)
+            t.get("long_data")
+
+    # Warm up first: a fresh process grows RSS by a one-time amount (allocator
+    # arenas, redis client buffers, interpreter caches) that is unrelated to any
+    # leak. Measure the SECOND, steady-state batch instead -- a real per-op leak
+    # keeps growing linearly, so post-warmup growth must stay near zero.
+    run_ops()
     rss = psutil.Process(os.getpid()).memory_info().rss
-    for _ in range(OP_COUNT):
-        t.set("long_data", fvs)
-        t.get("long_data")
+    run_ops()
     assert psutil.Process(os.getpid()).memory_info().rss - rss < OP_COUNT
 
--- a/.azure-pipelines/build-sairedis-template.yml
+++ b/.azure-pipelines/build-sairedis-template.yml
@@ -50,6 +50,20 @@ jobs:
     options:  "--privileged"
 
   steps:
+  - script: |
+      echo "Disk usage before cleanup:"; df -h /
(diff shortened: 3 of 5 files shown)


theasianpianist:
@senthil-nexthop — this PR fixes a flakiness in `Notifications.LruDedupPolicyConstructorAndLabelPropagation` (the test you added in 56a5f28). Under CPU load the drain loop could stop after the first pop while the second Redis pub/sub reply was still in flight, so `stats.received` was observed as 1 instead of 2 (reproduced 135/150 under 32-worker load). The fix drains until `received >= 2` (or a timeout) before asserting. Since you authored the test, your review would be appreciated — I couldn't add you as a formal reviewer (GitHub requires repo-collaborator access).

tests/ntf_ut.cpp
```diff
@@ -330,9 +330,14 @@ TEST(Notifications, LruDedupPolicyConstructorAndLabelPropagation)
...
-    // Drain so processReply is invoked.
+    // Drain so processReply is invoked. Both messages are delivered over Redis
+    // pub/sub, but the second reply may not have arrived by the time the first
+    // select() returns -- especially under load. LruDedup collapses the two
+    // byte-identical messages into a single queued item, so stopping at the first
+    // pop races the second reply and can observe received==1. Keep draining until
+    // both messages have been received (processReply'd), or we time out.
     size_t pops = 0;
-    for (int i = 0; i < 50 && pops == 0; ++i) {
+    for (int i = 0; i < 100 && nc.getStats().received < 2; ++i) {
```
senthil-nexthop, line 340: Gating on received count instead of pops looks like the right fix. I'm wondering if the iteration count is the right way to ensure the messages are delivered and drained by select(). How about we gate on wall clock time instead? I.e, wait for 2 messages to be received up to 10 seconds? Most likely the select will find the messages in the first few iterations, so the current form is fine. Did you observe the test taking multiple iterations in your env?

tests/test_redis_ut.py
```diff
@@ -873,9 +873,18 @@ def test_TableOpsMemoryLeak():
...
+    def run_ops():
+        for _ in range(OP_COUNT):
+            t.set("long_data", fvs)
+            t.get("long_data")
+
+    # Warm up first: a fresh process grows RSS by a one-time amount (allocator
+    # arenas, redis client buffers, interpreter caches) that is unrelated to any
+    # leak. Measure the SECOND, steady-state batch instead -- a real per-op leak
+    # keeps growing linearly, so post-warmup growth must stay near zero.
+    run_ops()
```
senthil-nexthop, line 886: There's a small chance that any leak in the first set() operation is not caught because of the warm up period, but the chances for that is very small.

senthil-nexthop:
> @senthil-nexthop — this PR fixes a flakiness in `Notifications.LruDedupPolicyConstructorAndLabelPropagation` (the test you added in [56a5f28](https://github.com/sonic-net/sonic-swss-common/commit/56a5f2886f34a666094c408a017220815ecc9c86)). Under CPU load the drain loop could stop after the first pop while the second Redis pub/sub reply was still in flight, so `stats.received` was observed as 1 instead of 2 (reproduced 135/150 under 32-worker load). The fix drains until `received >= 2` (or a timeout) before asserting. Since you authored the test, your review would be appreciated — I couldn't add you as a formal reviewer (GitHub requires repo-collaborator access).

Thanks for the fixes. I'm good with the test fixes, I'll leave it to the maintainers to comment on the pipeline changes.

tests/ntf_ut.cpp
theasianpianist: I didn't look at loop behavior for individual runs, but I would guess that it does sometimes take multiple iterations since the test was previously flaky with only a single iteration.

theasianpianist:
@sonic-net/sonic-swss-common-maintainer could you please help approve and merge this PR to address some flaky CI issues?

Pterosaur (approved):

senthil-nexthop (approved):

PR #1231, opened 2026-07-21
Cap consecutive readData error logging to avoid filling /var/log

mdhoff-ms:
I have noticed two instances of filled /var/log on DUTs with many repeated lines:
```
2026 Jul 21 01:47:00.605758 ERR caclmgrd: :- poll_descriptors: readData error: Unable to read redis reply
2026 Jul 21 01:47:00.605776 ERR healthd: :- poll_descriptors: readData error: Unable to read redis reply
2026 Jul 21 01:47:00.605793 ERR featured: :- poll_descriptors: readData error: Unable to read redis reply
```
these repeat immediately, leading to /var/log partition being filled very quickly. It seems that the error case of `common/select.cpp` (e.g. if redis is down) returns immediately, leading to callers who call in loops spinning indefinitely and filling the log partition.

To fix this I propose the following change:

- The readData error log should be capped during consecutive failures, to avoid flooding /var/log/syslog and filling the partition.
- [REMOVED FROM THIS PR AHEAD OF FUTURE DISCUSSION] On the `Select::ERROR` path, we should enforce the timeout requested by the caller (often 1ms, 100ms, etc) up to a cap of 1 second, to avoid pinning the CPU in a hot loop.

### DIFF
--- a/common/select.cpp
+++ b/common/select.cpp
@@ -2,6 +2,8 @@
 #include "common/logger.h"
 #include "common/select.h"
 #include <algorithm>
+#include <chrono>
+#include <thread>
 #include <stdio.h>
 #include <sys/time.h>
 #include <sys/types.h>
@@ -175,15 +177,25 @@ int Select::select(Selectable **c, int timeout, bool interrupt_on_signal)
     /* check if we have some data */
     ret = poll_descriptors(c, 0);
 
-    /* return if we have data, we have an error or desired timeout was 0 */
-    if (ret != Select::TIMEOUT || timeout == 0)
+    /* return immediately if appropriate */
+    if (ret == Select::OBJECT || ret == Select::SIGNALINT || timeout == 0)
+    {
         return ret;
+    }
 
-    /* wait for data */
-    ret = poll_descriptors(c, timeout, interrupt_on_signal);
+    if (ret == Select::TIMEOUT)
+    {
+        ret = poll_descriptors(c, timeout, interrupt_on_signal);
+    }
+
+    /* sleep 1s to throttle callers in a loop on error, e.g. redis
+        is down. without this it floods /var/log and pins the CPU. */
+    if (ret == Select::ERROR)
+    {
+        std::this_thread::sleep_for(std::chrono::seconds(1));
+    }
 
     return ret;
-
 }
 
 bool Select::isQueueEmpty()
--- a/common/select.cpp
+++ b/common/select.cpp
@@ -188,11 +188,12 @@ int Select::select(Selectable **c, int timeout, bool interrupt_on_signal)
         ret = poll_descriptors(c, timeout, interrupt_on_signal);
     }
 
-    /* sleep 1s to throttle callers in a loop on error, e.g. redis
-        is down. without this it floods /var/log and pins the CPU. */
+    /* Sleep 1s to throttle callers in a loop on error (e.g., Redis is down).
+        Without this, it floods /var/log and pins the CPU. */
     if (ret == Select::ERROR)
     {
-        std::this_thread::sleep_for(std::chrono::seconds(1));
+         constexpr auto kErrorBackoff = std::chrono::seconds(1);
+         std::this_thread::sleep_for(kErrorBackoff);
     }
 
     return ret;
--- a/common/select.cpp
+++ b/common/select.cpp
@@ -130,7 +130,16 @@ int Select::poll_descriptors(Selectable **c, unsigned int timeout, bool interrup
         }
         catch (const std::runtime_error& ex)
         {
-            SWSS_LOG_ERROR("readData error: %s", ex.what());
+            // Throttling by kErrorLogInterval avoids flooding /var/log and filling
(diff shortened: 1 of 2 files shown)


common/select.cpp
```diff
@@ -175,15 +177,25 @@ int Select::select(Selectable **c, int timeout, bool interrupt_on_signal)
     /* check if we have some data */
     ret = poll_descriptors(c, 0);
 
-    /* return if we have data, we have an error or desired timeout was 0 */
-    if (ret != Select::TIMEOUT || timeout == 0)
+    /* return immediately if appropriate */
+    if (ret == Select::OBJECT || ret == Select::SIGNALINT || timeout == 0)
+    {
         return ret;
+    }
```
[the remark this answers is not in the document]
mdhoff-ms: I think if a caller expects the function to be non-blocking, it should respect that contract even on error; then the caller can just check the result.

common/select.cpp
```diff
@@ -175,15 +177,26 @@ int Select::select(Selectable **c, int timeout, bool interrupt_on_signal)
...
 
-    /* wait for data */
-    ret = poll_descriptors(c, timeout, interrupt_on_signal);
+    /* Sleep 1s to throttle callers in a loop on error (e.g., Redis is down).
+        Without this, it floods /var/log and pins the CPU. */
+    if (ret == Select::ERROR)
+    {
+         constexpr auto kErrorBackoff = std::chrono::seconds(1);
+         std::this_thread::sleep_for(kErrorBackoff);
+    }
```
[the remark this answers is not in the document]
mdhoff-ms: I considered this, but some callers have small timeouts e.g. fdbsyncd 1ms. This is OK on the happy path where the function executes correctly and receives data, however 1ms will still fill syslog relatively quickly when it logs the error. I think it is reasonable to sleep for 1s on error.

mdhoff-ms:
maybe a good idea is to sleep only for the timeout that's passed in, then add a separate backoff (equal to 1s) to the actual logging statement. This would preserve the exact contract the caller is expecting while also mitigating the log spam.

common/select.cpp
```diff
@@ -175,15 +186,24 @@ int Select::select(Selectable **c, int timeout, bool interrupt_on_signal)
...
         return ret;
-
-    /* wait for data */
-    ret = poll_descriptors(c, timeout, interrupt_on_signal);
+    } else if (ret == Select::TIMEOUT)
+    {
+        ret = poll_descriptors(c, timeout, interrupt_on_signal);
+    } else if (ret == Select::ERROR)
+    {
+        // ERROR returns immediately, so a tight caller loop pins the CPU.
```
qiluo-msft: This is caller's responsibility to handle ERROR gracefully, not to burn CPU.

common/select.cpp
```diff
@@ -128,7 +130,16 @@ int Select::poll_descriptors(Selectable **c, unsigned int timeout, bool interrup
         }
         catch (const std::runtime_error& ex)
         {
-            SWSS_LOG_ERROR("readData error: %s", ex.what());
+            // Throttling by kErrorLogInterval avoids flooding /var/log and filling
+            // the partition from callers in a tight loop.
+            constexpr auto kErrorLogInterval = std::chrono::seconds(1);
+            static thread_local std::chrono::steady_clock::time_point lastErrorLog;
+            auto now = std::chrono::steady_clock::now();
```
qiluo-msft: Suggest not to add clock check. This function is critical to object sync performance.

mdhoff-ms: I have swapped this to a cap on total consecutive failure logs. After thinking about it, I actually prefer this to the clock method also.

common/select.cpp
mdhoff-ms: I considered this and looked into the feasibility of updating all the callers - unfortunately there are 66 calls to this function across C++ daemons (e.g. buffermgrd, portmgrd) and Python (e.g. xcvrd, bgpcfgd) and none of them are implementing this error check.

I think it would be a lot of effort to update all of these callers; the ones that pass a timeout value are likely expecting to block for this amount of time and so they maybe assume it is OK. I have tried to write this in a way that respects the contract the caller is expecting when they pass a timeout value in, up to 1 second.

mdhoff-ms: Removing from this PR and moving to future discussion

common/select.cpp
```diff
@@ -15,6 +15,16 @@ using namespace std;
...
 
+namespace {
+
+constexpr unsigned int kMaxConsecutiveErrorLogs = 10;
+
+// Thread-local rather than a member so as not to break the ABI
+thread_local unsigned int s_consecutiveErrors = 0;
+thread_local int s_lastFailedFd = -1;
+
+}
```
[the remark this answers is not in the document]
mdhoff-ms: At worst the behaviour is the same as before, but I don't even think the fds can alternate in our target scenario since epoll here is level-triggered.

mdhoff-ms:
Hi @qiluo-msft following up on this

PR #1233, opened 2026-07-23
[zmqclient]: Add send-path blip absorber, back-pressure counters, and inner retry cap

deepak-singhal0408:
#### Why I did it

Transport-layer half of the fix for [sonic-buildimage #28369](https://github.com/sonic-net/sonic-buildimage/issues/28369) (fpmsyncd ZMQ route drop under burst). The end-to-end design is described in the HLD [sonic-net/SONiC #2481 — Producer-side ZMQ route delivery resiliency and heap release](https://github.com/sonic-net/SONiC/pull/2481); this PR implements the swss-common transport-layer portion (send-path observability + a boundable inner retry loop) that the producer redesign builds on.

`ZmqClient::sendMsg()` drives the ZMQ socket in non-blocking mode and, on a full socket (`EAGAIN`) or transient unready state (`EINTR`/`EFSM`), retries with an exponential back-off ladder. Two gaps make that behavior hard to build a resilient producer on:

- The back-pressure signal that precedes a drop is only logged and then discarded, so it is invisible to the producer.
- The inner ladder length is fixed (`MQ_MAX_RETRY`), so a single full-socket message can block the send path for tens of seconds — which defeats a producer that owns an outer re-queue/coalescing loop and wants control back quickly.

#### How I did it

The new API and counters are additive — no existing call site changes and every getter defaults to zero. One deliberate timing change applies to the retry loop itself (documented in the Bounded back-off bullet): the sleep after the final, exhausting attempt is now skipped for all callers.

- **`setSendRetryConfig(int maxRetries, int maxBackoffMs = -1)`** — optional bound on the inner retry loop. `maxRetries < 0` keeps the default `MQ_MAX_RETRY` ladder; otherwise it is clamped to `MQ_MAX_RETRY` (this can only shorten the ladder, never extend it) and the loop makes at most `min(maxRetries, MQ_MAX_RETRY) + 1` attempts. `maxBackoffMs < 0` keeps the default exponential back-off; `>= 0` clamps each retry sleep to that ceiling. A producer with its own outer loop can configure a short "blip absorber" so control returns fast instead of blocking on one message.
- **Three process-local, per-instance atomic counters** exposed via inline getters, incremented at branch points that already fire:
  - `getSendEagainTotal()` — total `EAGAIN` (socket-full) occurrences; the leading indicator.
  - `getSendBlipAbsorbedTotal()` — sends that hit ≥1 `EAGAIN` but still succeeded within the inner retries (a transient blip the absorber swallowed so the caller never re-queued).
  - `getSendBackoffMaxMs()` — deepest per-attempt back-off ever waited; a coarse congestion-depth gauge.
- **Bounded back-off** — the exponential doubling is computed in a wider type and clamped to a fixed internal ceiling before narrowing back to `int`, so the multiply cannot overflow and `usleep()` always receives a sane duration. The clamped value is what gets logged, recorded in `backoff_max_ms`, and slept. No sleep is issued after the final, exhausting attempt — the loop is about to throw, so waiting served no purpose. This shortens the default/uncapped ladder's total wait from ~41 s to ~20 s (the deepest back-off actually waited is 10240 ms rather than 20480 ms); the outcome is unchanged (the call still throws on exhaustion, leaving the retry decision to the caller). The back-off-depth gauge and the retry log are updated only when a retry actually follows, so `backoff_max_ms` reflects back-off waited, not scheduled.

`ZmqClient` stays DB-agnostic — it only counts and exposes getters; the owner decides where/how to publish. The counters and atomics are appended as new members and no existing call site or default behavior changes; all swss-common consumers are rebuilt from source, so no binary-ABI guarantee is implied. The consuming fpmsyncd change is a follow-up sonic-swss PR.

#### How to verify it

New `ZmqClientSendPathCounters` unit tests in `tests/zmq_state_ut.cpp`:
- `blipAbsorberExhaustionCountsEagainAndClampsBackoff` — fills a peerless PUSH socket's local send buffer to `SNDHWM` to force deterministic `EAGAIN`, then asserts `eagain_total` advances by exactly `maxRetries + 1`, no blip is absorbed on a send that never succeeds, and each back-off is clamped to the caller's ceiling. A small `setSendRetryConfig` cap keeps it at a few ms instead of the default multi-second ladder.
- `defaultCountersStartZeroAndAccessorsLink` — guards that the accessors read zero before any send (the appended accessors compile, link, and initialize correctly).

Full swss-common ZMQ unit-test suite: **28/28 pass, 0 regressions.**

zmqclient: add optional inner send-retry cap (`setSendRetryConfig`) and process-local send-path back-pressure counters (`getSendEagainTotal` / `getSendBlipAbsorbedTotal` / `getSendBackoffMaxMs`); clamp the retry back-off against integer overflow. Defaults preserve existing behavior.

### DIFF
--- a/common/zmqclient.cpp
+++ b/common/zmqclient.cpp
@@ -15,6 +15,10 @@ using namespace std;
 
 namespace swss {
 
+// Ceiling for the exponential send-retry backoff; bounds the doubling so it
+// cannot overflow int and feed usleep() a garbage duration.
+static const int MQ_SEND_RETRY_DELAY_CEILING_MS = 60000;
+
 ZmqClient::ZmqClient(const std::string& endpoint)
     : ZmqClient(endpoint, "")
 {
@@ -150,6 +154,15 @@ void ZmqClient::connect()
     m_connected = true;
 }
 
+void ZmqClient::setSendRetryConfig(int maxRetries, int maxBackoffMs)
+{
+    // Clamp to MQ_MAX_RETRY so this can only shorten the default ladder.
+    m_sendMaxRetries.store((maxRetries < 0) ? MQ_MAX_RETRY
+                                            : std::min(maxRetries, MQ_MAX_RETRY),
+                           std::memory_order_relaxed);
+    m_sendMaxBackoffMs.store(maxBackoffMs, std::memory_order_relaxed);
+}
+
 void ZmqClient::sendMsg(
         const std::string& dbName,
         const std::string& tableName,
@@ -173,7 +186,12 @@ void ZmqClient::sendMsg(
     int zmq_err = 0;
     int retry_delay = 10;
     int rc = 0;
-    for (int i = 0; i <= MQ_MAX_RETRY; ++i)
+    bool saw_eagain = false;
+    // Snapshot the caps so this send reads stable values (they are atomic and
+    // may be reconfigured between sends).
+    const int max_retries = m_sendMaxRetries.load(std::memory_order_relaxed);
+    const int max_backoff_ms = m_sendMaxBackoffMs.load(std::memory_order_relaxed);
+    for (int i = 0; i <= max_retries; ++i)
     {
         {
             // ZMQ socket is not thread safe: http://api.zeromq.org/2-1:zmq
@@ -191,13 +209,27 @@ void ZmqClient::sendMsg(
         }
         if (rc >= 0)
         {
+            // Absorbed a transient full-socket blip; caller never had to re-queue.
+            if (saw_eagain)
+            {
+                m_sendBlipAbsorbedTotal.fetch_add(1, std::memory_order_relaxed);
+            }
             SWSS_LOG_DEBUG("zmq sended %d bytes", serializedlen);
             return;
         }
 
         zmq_err = zmq_errno();
-        // sleep (2 ^ retry time) * 10 ms
-        retry_delay *= 2;
+        // sleep (2 ^ retry time) * 10 ms. Double in a wider type and clamp to the
+        // ceiling before narrowing, so the multiply cannot overflow int; then
+        // apply the caller's cap. All three (log/record/sleep) see the real delay.
+        int64_t doubled = static_cast<int64_t>(retry_delay) * 2;
+        retry_delay = (doubled > MQ_SEND_RETRY_DELAY_CEILING_MS)
+                          ? MQ_SEND_RETRY_DELAY_CEILING_MS
+                          : static_cast<int>(doubled);
+        if (max_backoff_ms >= 0 && retry_delay > max_backoff_ms)
+        {
+            retry_delay = max_backoff_ms;
+        }
         if (zmq_err == EINTR
             || zmq_err== EFSM)
         {
@@ -213,6 +245,16 @@ void ZmqClient::sendMsg(
         else if (zmq_err == EAGAIN)
         {
             // EAGAIN: ZMQ is full to need try again
+            saw_eagain = true;
+            m_sendEagainTotal.fetch_add(1, std::memory_order_relaxed);
+            // Track the deepest backoff waited (congestion-depth gauge).
+            uint64_t observed = m_sendBackoffMaxMs.load(std::memory_order_relaxed);
+            while (static_cast<uint64_t>(retry_delay) > observed &&
+                   !m_sendBackoffMaxMs.compare_exchange_weak(observed,
+                                                             static_cast<uint64_t>(retry_delay),
+                                                             std::memory_order_relaxed))
+            {
+            }
             SWSS_LOG_WARN("zmq is full, will retry in %d ms, endpoint: %s, error: %d", retry_delay, m_endpoint.c_str(), zmq_err);
         }
         else if (zmq_err == ETERM)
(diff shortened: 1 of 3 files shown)


common/zmqclient.cpp
```diff
@@ -150,6 +150,12 @@ void ZmqClient::connect()
     m_connected = true;
 }
 
+void ZmqClient::setSendRetryConfig(int maxRetries, int maxBackoffMs)
+{
+    m_sendMaxRetries = (maxRetries < 0) ? MQ_MAX_RETRY : maxRetries;
+    m_sendMaxBackoffMs = maxBackoffMs;
+}
```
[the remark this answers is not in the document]
deepak-singhal0408: Addressed in e3ed1ac. `setSendRetryConfig` now clamps a caller-supplied count with `std::min(maxRetries, MQ_MAX_RETRY)`, so the inner ladder can only be shortened, never extended past the historical default. The header doc and the `retry_delay` overflow-ceiling rationale are updated to match.

common/zmqclient.cpp
```diff
@@ -191,13 +213,34 @@ void ZmqClient::sendMsg(
...
         // sleep (2 ^ retry time) * 10 ms
         retry_delay *= 2;
+        // Clamp the exponential backoff to a fixed ceiling before it is used.
+        // The doubling can overflow int for a large configured retry count; the
+        // ceiling bounds it, and the < 0 term also catches any pathological
+        // signed-overflow wrap so usleep() never sees a garbage duration.
+        if (retry_delay > MQ_SEND_RETRY_DELAY_CEILING_MS || retry_delay < 0)
+        {
+            retry_delay = MQ_SEND_RETRY_DELAY_CEILING_MS;
+        }
```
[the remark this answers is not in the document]
deepak-singhal0408: Addressed in e3801c8. The doubling is now computed in a wider type and clamped before narrowing back to `int` (`int64_t doubled = (int64_t)retry_delay * 2;` then clamp to the ceiling), so the multiply can no longer overflow and the guard no longer relies on detecting a signed wrap after the fact.

common/zmqclient.cpp
```diff
@@ -230,10 +272,15 @@ void ZmqClient::sendMsg(
             throw system_error(make_error_code(errc::io_error), message);
         }
 
-        usleep(retry_delay * 1000);
+        // No sleep after the final attempt — the loop is about to exit and throw,
+        // so returning control immediately is what a capped caller wants.
+        if (i < max_retries)
```
qiluo-msft, line 280: this skips the final-attempt sleep on the *default* path too, not just capped callers. old loop slept after every iteration incl. the last, so at `max_retries == MQ_MAX_RETRY` (10) this drops the trailing ~20480 ms wait — an uncapped producer on a saturated socket now throws after ~20 s instead of ~41 s. almost certainly a good change, but it's a real timing shift for existing producers, so the PR's "exact default behavior preserved" isn't quite true here (the counters/caps are; this sleep-skip isn't). worth calling out.

common/zmqclient.cpp
```diff
@@ -213,6 +245,16 @@ void ZmqClient::sendMsg(
...
+            m_sendEagainTotal.fetch_add(1, std::memory_order_relaxed);
+            // Track the deepest backoff waited (congestion-depth gauge).
+            uint64_t observed = m_sendBackoffMaxMs.load(std::memory_order_relaxed);
+            while (static_cast<uint64_t>(retry_delay) > observed &&
+                   !m_sendBackoffMaxMs.compare_exchange_weak(observed,
+                                                             static_cast<uint64_t>(retry_delay),
+                                                             std::memory_order_relaxed))
+            {
+            }
             SWSS_LOG_WARN("zmq is full, will retry in %d ms, endpoint: %s, error: %d", retry_delay, m_endpoint.c_str(), zmq_err);
```
qiluo-msft: on the exhausting attempt (`i == max_retries`) this logs "will retry in %d ms" and then throws without retrying — the message is false on the last pass. gate it on `i < max_retries`?

common/zmqclient.cpp
deepak-singhal0408: You are right that the final-attempt sleep is the meaningful change. The loop already skips the sleep on the exhausting attempt (`if (i < max_retries)`), which is what shortens the uncapped default from ~41 s to ~20 s (deepest back-off actually waited is 10240 ms, not 20480 ms). I have kept that skip — on exhaustion the loop throws without waiting, so the sleep before the throw is dead time and the retry decision is left to the caller — and updated the PR description to call out the timing change instead of claiming the default sleep is preserved. Separately, the back-off-depth gauge and the "will retry" log are now recorded only when a retry actually follows, so the gauge reflects back-off waited rather than scheduled.

common/zmqclient.cpp
deepak-singhal0408: Gated the "will retry" log on i < max_retries in 1e087f9. On the exhausting attempt the loop throws without retrying, and that final failure is already reported by the throw-path error log, so the message no longer advertises a retry that does not happen.

venkit-nexthop (approved):

xq9mend (approved):

qiluo-msft (approved):

PR #1234, opened 2026-08-06
[zmq] Fix use-after-free in deferred burst notify; bound notification deferral

venkit-nexthop:
#### Why I did it

Three follow-ups to #1187's burst-coalescing ZmqRouteServer:

1. **Use-after-free in the deferred notify** (raised post-merge in #1187, [here](https://github.com/sonic-net/sonic-swss-common/pull/1187#discussion_r3678256927)): `mqPollThread` collects raw handler pointers across poll iterations and notifies them after the burst quiesces. A consumer destroyed in that window had already unregistered, but the flush still called `notifyPending()` through the stale pointer.

2. **Unbounded notification deferral** (raised in review of sonic-net/sonic-swss#4564, [here](https://github.com/sonic-net/sonic-swss/pull/4564#pullrequestreview-4862028324), item 1): the quiesce notify needs a BURST_QUIESCE_MS pause and the consumer-side threshold needs enough distinct keys. A stream that never pauses and stays under the threshold deferred notification indefinitely. Coalescing keeps the staged state current, so the exposure is delayed convergence -- but a sustained same-key flap is exactly when prompt programming of the latest state matters.

3. **Destroy-time use-after-free in `ZmqRouteConsumerStateTable`** (raised in review, [here](https://github.com/sonic-net/sonic-swss-common/pull/1234#pullrequestreview-5011512179), and confirmed valid): the registry detach lives in the base `~ZmqConsumerStateTable()`, but member destruction runs most-derived-first, so the subclass's `m_ingressCallback` is already destroyed by the time the base destructor detaches. In that window a live `mqPollThread` can dispatch into the overridden `handleReceivedData()` and read a destroyed `std::function`. Not reachable in orchagent (which exits via `_exit(0)`), but reachable from tests that destroy a consumer while the poll thread runs, and a real free-then-dereference for any callback with heap-allocating captures.

#### How I did it

Commit 1: add `ZmqHandlerRegistry::flushDirtyHandlers()` and flush through it. Notification happens under the same mutex `removeHandler()` takes, iterating the registry so membership doubles as the liveness check; removed handlers are skipped instead of dereferenced.

Commit 2: `dirtyHandlers` becomes a map of handler to first-dirty time, and the flush takes a cutoff. The quiesce path passes now() (flush everything); a new per-drain-pass call passes now() minus `BURST_MAX_HOLDOFF_MS` (50 ms), bounding notification latency even when a burst never quiesces. Overdue entries are erased live or dead, so a dangling pointer cannot pin the poll loop in its short-timeout mode.

Commit 3: add a protected, idempotent `ZmqConsumerStateTable::detachFromRegistry()` and call it from a new most-derived `~ZmqRouteConsumerStateTable()`, so the registry detach happens while `m_ingressCallback` is still alive; `removeHandler()` blocks any in-flight dispatch and prevents new ones before the subclass members go away. The base destructor now routes through the same helper, so its later call is a harmless no-op and the base's stated "detach before members are destroyed" invariant holds for subclasses too.

#### How to verify it

New tests: `ZmqHandlerRegistry.FlushSkipsUnregisteredHandlers` (unregister between dirty and flush -- the survivor is notified once, the removed handler never), `ZmqHandlerRegistry.FlushHonorsCutoff` (only overdue entries flushed), and `ZmqRouteConsumerStateTable.ContinuousStreamWakesWithinHoldoff`. The last drives a 300 us same-key stream with no quiesce gap and asserts the measured Select wake latency lands in [40, 150] ms (the ~50 ms holdoff, not the ~5 ms quiesce). It is a bounded retry: a trial that wakes under 40 ms means a stray >=5 ms scheduler gap consumed it, so it re-runs on a fresh port (up to 5 attempts); a conclusive trial in-window passes, a 500 ms Select timeout fails immediately (the unbounded-deferral regression this PR fixes), and all-5-early fails as no-deferral rather than skipping. Confirmed it fails with `BURST_MAX_HOLDOFF_MS` reverted (Select times out at 500 ms).

Ran the zmq route + registry suites locally (sonic-slave-bookworm, redis): 10/10, and the existing burst/wakeup tests are unchanged and green.


venkit-nexthop:
@qiluo-msft I have fixed your comment from https://github.com/sonic-net/sonic-swss-common/pull/1187 in this PR
@deepak-singhal0408 I have fixed your comment from https://github.com/sonic-net/sonic-swss/pull/4564 in this PR

aminclude_static.am
```diff
@@ -0,0 +1,126 @@
+
+# aminclude_static.am generated automatically by Autoconf
```
deepak-singhal0408: This looks generated rather than authored (Autoconf / `AX_AM_MACROS_STATIC`, dated 2026-08-05), it isn't tracked on `master`, and nothing in the repo includes it. Suggest dropping it and adding it to `.gitignore`.

common/cfg_schema.h
```diff
@@ -0,0 +1,12 @@
+#ifndef CFG_SCHEMA_H
```
deepak-singhal0408: `common/Makefile.am` lists this in `BUILT_SOURCES` with a `FORCE` rule, so every build regenerates it and an in-tree build leaves the tree dirty. Suggest dropping it and adding it to `.gitignore`.

common/zmqrouteserver.cpp
```diff
@@ -105,9 +116,19 @@ void ZmqRouteServer::mqPollThread()
...
+                // is what the holdoff below is measured from.
+                dirtyHandlers.emplace(handler, std::chrono::steady_clock::now());
             }
         }
+
+        // A continuous stream keeps zmq_poll returning with data, so the
+        // rc == 0 quiesce flush above may never run. Flush overdue handlers
+        // here, once per drain pass, to bound notification latency at
+        // BURST_MAX_HOLDOFF_MS even when the burst never quiesces.
+        getHandlerRegistry()->flushDirtyHandlers(
```
deepak-singhal0408, line 137: This does cover the case we discussed on [sonic-swss PR 4564](https://github.com/sonic-net/sonic-swss/pull/4564): with sub-5 ms gaps the socket still empties briefly, `zmq_recv` returns `EAGAIN`, the inner loop breaks, and this flush runs — so a sustained trickle settles into a periodic ~50 ms notify.

One narrower case is left, and it's really about the comment's wording. Because the call sits after the drain loop, whose only exit is `EAGAIN`, a producer that keeps the socket continuously non-empty (arrival rate ≥ drain rate, rather than just short gaps) never lets the inner loop exit, so the flush isn't reached — the guarantee is "once per drain-to-empty pass" rather than "within 50 ms". That needs saturation *and* fewer than `gMaxBulkSize` distinct keys, so it's unlikely in practice; the comment just claims a bit more than the placement delivers. If you want the bound to hold unconditionally, computing a deadline before the loop and testing it every N messages inside the loop would do it.

common/zmqrouteserver.cpp
```diff
@@ -105,9 +116,19 @@ void ZmqRouteServer::mqPollThread()
 
             if (auto* handler = getHandlerRegistry()->dispatch(dbName, tableName, kcos))
             {
-                dirtyHandlers.insert(handler);
+                // emplace keeps the FIRST dirty time on repeat touches, which
+                // is what the holdoff below is measured from.
+                dirtyHandlers.emplace(handler, std::chrono::steady_clock::now());
```
deepak-singhal0408: `steady_clock::now()` is an argument here, so it's evaluated for every message even though `emplace` discards it whenever the handler is already dirty — the common case during a burst. Hoisting one `now()` above the inner loop and reusing it is equivalent for the holdoff semantics and takes a clock read off the per-message path.

tests/zmq_route_ut.cpp
```diff
@@ -321,3 +321,112 @@ TEST(ZmqRouteConsumerStateTable, MultipleBurstsEachWakeSeparately)
...
+        {
+            p.set("flap_key", vector<FieldValueTuple>{{"seq", "x"}});
+            std::this_thread::sleep_for(std::chrono::milliseconds(2));
+        }
+    });
+
+    // 500ms select timeout: an unbounded deferral would time out here, the
+    // 50ms holdoff wakes us an order of magnitude earlier.
+    Selectable *out = nullptr;
+    int ret = sel.select(&out, 500);
```
deepak-singhal0408: `sleep_for(2ms)` is a floor, not a bound — across ~250 iterations, one overshoot past the 5 ms `BURST_QUIESCE_MS` fires the pre-existing quiesce flush, so the no-gap precondition isn't established by construction. And since quiesce wakes at ~5 ms vs the holdoff's ~50 ms, both clear the 500 ms assert with nothing recording which fired — the test passes with `BURST_MAX_HOLDOFF_MS` reverted.

That matters because this is the only test covering the `mqPollThread` wiring — `FlushHonorsCutoff` pins the cutoff logic itself, but not the call's placement or the cutoff it's given. Pacing at a few hundred µs plus an assert on measured wake latency would attribute the wake.

deepak-singhal0408 (commented):
One thing the registry fix doesn't reach, on a file outside this diff:

[`ZmqConsumerStateTable::~ZmqConsumerStateTable()`](https://github.com/sonic-net/sonic-swss-common/blob/master/common/zmqconsumerstatetable.cpp#L53) calls `removeHandler()`, which is correct for its own members — but `ZmqRouteConsumerStateTable` declares no destructor, so its `m_ingressCallback` is destroyed before that base body runs, while the object still has the derived vptr and is still registered. An in-flight `dispatch()` in that window calls the overridden `handleReceivedData()` and reads a destroyed `std::function`. The dtor comment states the invariant ("detach from the registry before any of our members are destroyed") that the subclass then breaks.

Not reachable in orchagent, which exits via `_exit(0)`, but reachable from tests that destroy a consumer while the poll thread runs. Detaching in the most-derived destructor — or a `detach()` helper each most-derived class calls — would close it and make the requirement explicit for future subclasses.

aminclude_static.am
venkit-nexthop: Yes, I will remove this.

common/cfg_schema.h
venkit-nexthop: Will do

aminclude_static.am
venkit-nexthop: Removed in 1b0db9f and added to .gitignore.

common/cfg_schema.h
venkit-nexthop: Removed in 1b0db9f and added to .gitignore.

common/zmqrouteserver.cpp
venkit-nexthop: Reworded the comment in 1b0db9f to state the per-drain-pass guarantee. Kept the flush where it is - the saturating case also needs to stay under gMaxBulkSize distinct keys, and the in-loop deadline check would put a clock read back on the per-message path.

common/zmqrouteserver.cpp
venkit-nexthop: Hoisted in 1b0db9f - one clock read per drain pass. Pass-start granularity only makes flushes earlier, never later.

tests/zmq_route_ut.cpp
venkit-nexthop: Reworked in 1b0db9f: producer paces at 300us so a quiesce gap needs a >16x scheduler stall, and the test asserts measured wake latency lands in 40-150ms. Checked it fails with BURST_MAX_HOLDOFF_MS reverted (select times out at 500ms).

tests/zmq_route_ut.cpp
```diff
@@ -321,3 +321,125 @@ TEST(ZmqRouteConsumerStateTable, MultipleBurstsEachWakeSeparately)
...
+        std::chrono::steady_clock::now() - selectStart);
+    stop = true;
+    producer.join();
+
+    EXPECT_EQ(ret, Select::OBJECT);
+    EXPECT_EQ(out, &c);
+    // Holdoff wake lands at ~BURST_MAX_HOLDOFF_MS (50ms) plus one drain
+    // pass. Below 40ms means a quiesce/gap wake fired first (precondition
+    // broken); above 150ms means the wake wasn't the holdoff.
+    EXPECT_GE(wakeLatency.count(), 40);
```
deepak-singhal0408: This test assumes the stream never pauses for `BURST_QUIESCE_MS`. Nothing enforces that: `sleep_for(300us)` guarantees a minimum, not a maximum.

If any one of the ~166 sleeps in the window overshoots to 5 ms, `zmq_poll` times out and the `rc == 0` path flushes with `cutoff = now()`, which ignores the 50 ms holdoff entirely. `select()` wakes early and `EXPECT_GE(wakeLatency, 40)` fails -- even though the code did the right thing.

The stall is 16x relative to 300 us but only ~4.7 ms absolute, and the window gives ~166 independent chances at it. On a contended host I see gaps cross 5 ms; even idle they come close. CI compiles these with coverage instrumentation (`build-env/build.sh:20`), which widens the tail further.

The upper bound already catches a reverted `BURST_MAX_HOLDOFF_MS`. The lower bound is testing whether the test's own precondition held, so a sub-40 ms wake may be better treated as inconclusive than as a failure.

deepak-singhal0408 (commented):
The description's "How to verify it" still describes the 2 ms stream waking "well inside a 500 ms window"; `1b0db9f` changed the pacing to 300 us and the assertion to a two-sided 40-150 ms window. Please fix the PR body?

Still open from the last round: the detach happens in `~ZmqConsumerStateTable()`, so `ZmqRouteConsumerStateTable::m_ingressCallback` is destroyed before the base detaches. Not reachable in production (`_exit(0)`), reachable from tests that destroy a consumer while the poll thread is live.

tests/zmq_route_ut.cpp
venkit-nexthop: Agreed — as written the lower bound asserts the test's own precondition, and one overshoot in ~166 chances isn't a precondition to build on, especially under coverage builds. Rather than dropping the signal, I've reworked it in 42824f8 as a bounded retry: a trial that wakes under 40ms is treated as inconclusive (a quiesce gap consumed it) and re-runs with fresh harness state on a fresh port, up to 5 attempts. One conclusive trial in [40,150]ms passes; a 500ms select timeout still fails any trial immediately (unbounded deferral — the regression this PR fixes); and if all 5 trials wake early, that's no longer a plausible run of independent >16x stalls — it means the deferral isn't happening at all (per-message notify), so the test fails rather than skips. This keeps the lower bound's regression value while a stray 5ms gap costs one retry instead of a CI failure.

Verified both directions locally: 15/15 repeats pass, and with `BURST_MAX_HOLDOFF_MS` reverted the test fails on the first trial's select timeout.

deepak-singhal0408 (dismissed):
Approving — the retry rework in [`42824f8`](https://github.com/sonic-net/sonic-swss-common/pull/1234/commits/42824f84e1ec3bf2ac094c17be29efe4ea02b704) addresses the lower-bound flake, and CI is green at this head.

Two non-blocking nits:

1. PR body's "How to verify it" still describes the 2 ms stream and a 500 ms window; the test now paces at 300 us with a bounded-retry wake in [40, 150] ms. Worth refreshing before merge.

2. `ZmqRouteConsumerStateTable` declares no destructor, so `m_ingressCallback` is destroyed before [`~ZmqConsumerStateTable()`](https://github.com/sonic-net/sonic-swss-common/blob/42824f84e1ec3bf2ac094c17be29efe4ea02b704/common/zmqconsumerstatetable.cpp#L53) calls `removeHandler()` — the "safe to destroy after `removeHandler()` returns" contract ([`zmqserver.h:41`](https://github.com/sonic-net/sonic-swss-common/blob/42824f84e1ec3bf2ac094c17be29efe4ea02b704/common/zmqserver.h#L41)) doesn't cover the derived members. Not reachable in orchagent (`_exit(0)`), and current tests wouldn't flake on it (captureless lambdas read as an empty `std::function`), so this is a follow-up rather than a change here. It becomes a free-then-dereference for a callback with heap-allocating captures. Detaching in the most-derived destructor would make the base dtor's stated invariant hold for subclasses.

deepak-singhal0408:
Hi @qiluo-msft, could you help review/merge this PR? thanks,

venkit-nexthop:
@qiluo-msft could you help merge this?

qiluo-msft:
> One thing the registry fix doesn't reach, on a file outside this diff:
> 
> [`ZmqConsumerStateTable::~ZmqConsumerStateTable()`](https://github.com/sonic-net/sonic-swss-common/blob/master/common/zmqconsumerstatetable.cpp#L53) calls `removeHandler()`, which is correct for its own members — but `ZmqRouteConsumerStateTable` declares no destructor, so its `m_ingressCallback` is destroyed before that base body runs, while the object still has the derived vptr and is still registered. An in-flight `dispatch()` in that window calls the overridden `handleReceivedData()` and reads a destroyed `std::function`. The dtor comment states the invariant ("detach from the registry before any of our members are destroyed") that the subclass then breaks.
> 
> Not reachable in orchagent, which exits via `_exit(0)`, but reachable from tests that destroy a consumer while the poll thread runs. Detaching in the most-derived destructor — or a `detach()` helper each most-derived class calls — would close it and make the requirement explicit for future subclasses.

Agree with @deepak-singhal0408 . This is a valid bug.

venkit-nexthop:
Pushed 05bb48c to address the destructor use-after-free follow-up that @deepak-singhal0408 raised and @qiluo-msft confirmed:

- `~ZmqRouteConsumerStateTable()` now detaches from the registry (via a new idempotent `ZmqConsumerStateTable::detachFromRegistry()`) **before** its members are destroyed, so `removeHandler()` blocks any in-flight dispatch and bars new ones while `m_ingressCallback` is still alive. The base destructor routes through the same helper and stays self-protecting for plain (non-route) consumers, so the "detach before members are destroyed" invariant now holds for subclasses too.
- Added `DestroyDuringStreamIsSafe`: an AddressSanitizer stress reproducer that destroys a consumer mid-stream with a heap-owning ingress callback, so a missing derived detach is flagged.
- Also refreshed the PR body's "How to verify it" to match the current 300us / [40,150]ms holdoff test.

CI is green on this head (buildId 1211381) — all builds, `BuildenvSetupUnitTests`, `coverage`, and `Test vstest` pass. Could you re-approve when you get a chance? Thanks!

venkit-nexthop:
@qiluo-msft @deepak-singhal0408 could you take a look at this PR. I have fixed your review comments.

venkit-nexthop:
@qiluo-msft @deepak-singhal0408 could you take a look and approve this?

deepak-singhal0408 (approved):

venkit-nexthop:
Thanks for approving @deepak-singhal0408 
@qiluo-msft can you help merge this ?

venkit-nexthop:
@deepak-singhal0408 gentle nudge — this is approved (thanks!) and mergeable/clean with all Azure + Analyze checks green. Could you help land it? It's the last upstream swss-common piece of the orchagent route-redesign burst-coalescing work (follow-up to #1187): the deferred-notify UAF fix plus the bounded notification-deferral (drain-to-empty flush). Appreciate it!

PR #1236, opened 2026-08-10
[ci]: Preserve dependency order across buildenv artifact install groups

theasianpianist:
## Summary

Fix `buildenv_setup` so cascading upstream dependencies are installed before
the artifacts that depend on them, including when an intervening dependency
requires a different `install_env`.

This is the root fix for sonic-swss PR #4821 / Azure build `1189333`, where the
flattened install plan put sairedis in the global empty-environment group before
VPP. `dpkg` left `libsaivs` unconfigured; `apt-get install -f` removed it; VPP
was installed only afterward; and the SWSS compile then failed because the SAI
headers were absent.

## Changes

- Traverse each bundle's nested `upstream-artifacts.yaml` before appending the
  parent artifact, producing a dependency-first artifact sequence.
- Replace global grouping by `install_env` with ordered batching:
  - merge only adjacent artifacts with the same environment;
  - preserve environment boundaries and dependency order;
  - allow wheel-only artifacts between compatible DEB providers/dependents
    without unnecessarily splitting the batch.
- Keep existing union behavior for `dpkg_args` and `apt_fix_broken` within each
  batch.

For the SWSS cascade, the resulting install plan is:

```text
common-libs + sonic-swss-common
VPP (VPP_INSTALL_SKIP_SYSCTL=1)
sonic-sairedis + sonic-dash-api
```

## Validation

- Targeted cascade/planner tests: **29 passed**.
- Full buildenv_setup suite: **118 passed**, **94% coverage**.
- Added regressions for:
  - dependency-first nested traversal;
  - required staged upstream reached only through nesting;
  - same-env provider/dependent separated by a special-env dependency;
  - preservation of environment insertion order;
  - wheel-only artifacts not splitting compatible DEB batches.
- Real end-to-end validation in `sonic-slave-bookworm:master-amd64` with the
  SWSS config **without** its temporary install-env workaround:
  - common-libs/swss-common installed first;
  - VPP installed second;
  - sairedis/dash installed third;
  - `vpp`, `libsaivs`, and `libsaivs-dev` all ended `install ok installed`.

After this merges, the temporary sairedis/VPP grouping workaround in
sonic-swss #4821 will be removed and that PR rerun against the shared fix.

### DIFF
--- a/ci/buildenv_setup/planner.py
+++ b/ci/buildenv_setup/planner.py
@@ -155,30 +155,41 @@ def _pip_batches(pip_specs: List[Tuple[str, Tuple[str, ...]]]):
 
 
 def _deb_install_groups(artifacts) -> List[dict]:
-    """Group every artifact's DEBs by install_env signature into single dpkg -i
-    calls (see run() step 3 for rationale). Returns groups ordered so that the
-    empty-install_env group (the usual library providers such as libnl/libyang3/
-    libswsscommon) is installed before any special-install_env group (e.g. vpp).
-    Within a group, dpkg_args are unioned and apt_fix_broken is ORed."""
-    deb_groups: "OrderedDict[tuple, dict]" = OrderedDict()
+    """Batch dependency-first artifacts into ordered ``dpkg -i`` calls.
+
+    ``collect_bundles`` returns nested dependencies before their parents. Merge
+    only ADJACENT artifacts with the same ``install_env`` signature: dpkg can
+    then unpack/configure cross-artifact dependencies in one invocation without
+    moving a later dependent ahead of an intervening dependency that needs a
+    different environment (e.g. common-libs -> VPP -> sairedis).
+
+    Within a batch, dpkg_args are unioned and apt_fix_broken is ORed.
+    Artifacts containing only wheels do not split an otherwise-adjacent DEB
+    batch."""
+    deb_groups: List[dict] = []
+    current_sig = None
+    current = None
     for art in artifacts:
+        if not art.deb_files:
+            continue
         env_sig = tuple(sorted((art.install_env or {}).items()))
+        if current is None or env_sig != current_sig:
+            current_sig = env_sig
+            current = {
+                "files": [],
+                "args": [],
+                "fix": False,
+                "env": dict(art.install_env or {}),
+            }
+            deb_groups.append(current)
         for deb in art.deb_files:
             dpkg_args, fix = art.deb_opts.get(deb, ([], False))
-            g = deb_groups.get(env_sig)
-            if g is None:
-                g = {"files": [], "args": [], "fix": False, "env": dict(art.install_env or {})}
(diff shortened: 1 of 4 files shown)


ci/buildenv_setup/planner.py
```diff
@@ -155,30 +155,41 @@ def _pip_batches(pip_specs: List[Tuple[str, Tuple[str, ...]]]):
...
+
+    ``collect_bundles`` returns nested dependencies before their parents. Merge
+    only ADJACENT artifacts with the same ``install_env`` signature: dpkg can
+    then unpack/configure cross-artifact dependencies in one invocation without
+    moving a later dependent ahead of an intervening dependency that needs a
+    different environment (e.g. common-libs -> VPP -> sairedis).
+
+    Within a batch, dpkg_args are unioned and apt_fix_broken is ORed.
+    Artifacts containing only wheels do not split an otherwise-adjacent DEB
+    batch."""
```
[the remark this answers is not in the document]
theasianpianist: Fixed in 995eb23. The step-3 comment now describes dependency-first artifact order and adjacent same-install_env batching, including the common-libs → VPP → sairedis boundary, instead of the removed global grouping/empty-env-first behavior.

ci/buildenv_setup/planner.py
```diff
@@ -155,30 +155,41 @@ def _pip_batches(pip_specs: List[Tuple[str, Tuple[str, ...]]]):
 
 
 def _deb_install_groups(artifacts) -> List[dict]:
-    """Group every artifact's DEBs by install_env signature into single dpkg -i
-    calls (see run() step 3 for rationale). Returns groups ordered so that the
-    empty-install_env group (the usual library providers such as libnl/libyang3/
-    libswsscommon) is installed before any special-install_env group (e.g. vpp).
-    Within a group, dpkg_args are unioned and apt_fix_broken is ORed."""
-    deb_groups: "OrderedDict[tuple, dict]" = OrderedDict()
+    """Batch dependency-first artifacts into ordered ``dpkg -i`` calls.
```
qiluo-msft, line 158: nit: worth one sentence in the docstring that the caller is responsible for dependency-first order — this function only preserves it, it no longer enforces it. the old version had a defensive "empty-env first" sort that corrected a mis-ordered caller; now a `collect_bundles` bug returning artifacts out of order would silently produce a wrong install plan. something like: "The caller (collect_bundles) must supply artifacts in dependency-first order; this function only preserves that order."

qiluo-msft (dismissed):

ci/buildenv_setup/planner.py
theasianpianist: Addressed in 733ce09. The `_deb_install_groups()` docstring now explicitly states that its caller (normally `collect_bundles`) must supply dependency-first order; the function preserves that order but does not derive or validate dependencies.

qiluo-msft (approved):

PR #1241, opened 2026-08-27
[common]: Add failed neighbor table name

theasianpianist:
## What I did
- add `APP_NEIGH_FAILED_TABLE_NAME` for the new `NEIGH_FAILED_TABLE`. neighsyncd will write FAILED IPv6 kernel neighbor entries to this table on dualtor platforms. nbrmgrd will read from this table and set the entries to INCOMPLETE so that any subsequent NA messages can resolve the neighbor.

## Testing
- not run (schema constant only)


common/schema.h
```diff
@@ -166,6 +166,7 @@ namespace swss {
 #define APP_BUFFER_PORT_EGRESS_PROFILE_LIST_NAME    "BUFFER_PORT_EGRESS_PROFILE_LIST_TABLE"
 
 #define APP_NEIGH_RESOLVE_TABLE_NAME        "NEIGH_RESOLVE_TABLE"
+#define APP_KERNEL_FAILED_NEIGH_TABLE_NAME  "KERNEL_FAILED_NEIGH_TABLE"
```
prsunny: can you please rename to `APP_NEIGH_FAILED_TABLE_NAME`. Would like to add this to ST as well for general telemetry

theasianpianist: Updated to `APP_NEIGH_FAILED_TABLE_NAME` with value `NEIGH_FAILED_TABLE`, and updated the dependent sonic-swss draft PR as well.

prabhataravind (approved):

PR #1242, opened 2026-08-27
[common]: Add shared VRF name validation

ashutosh-agrawal:
## What I did

Added a shared `swss::isVrfNameValid()` helper with C++ and Python bindings.
The validator applies the Linux interface-name constraints used by SONiC VRFs:

- Length is 1 through 15 bytes.
- The first character is an ASCII letter, digit, underscore, or period.
- Remaining characters may additionally contain hyphens.
- The complete names `.` and `..` are rejected.

Added focused C++ and Python coverage for current VRF names, accepted syntax,
rejected syntax, and the 15/16-byte boundary.

## Why I did it

This provides one reusable VRF-name validator through `sonic-swss-common`
instead of requiring each consumer to maintain its own pattern. Existing forms
including `default`, `mgmt`, `vrfRED`, and `Vrf-RED_1` remain accepted.

The first consumer is sonic-net/sonic-buildimage#29228.

## Testing

- Canonical Bookworm Debian package build completed successfully.
- Focused C++ tests: `2 passed`.
- Python binding test: `1 passed`.

### DIFF
--- a/common/vrf.h
+++ b/common/vrf.h
@@ -1,0 +1,42 @@
+#ifndef __VRF__
+#define __VRF__
+
+#include <net/if.h>
+#include <string>
+
+namespace swss
+{
+
+const size_t VRF_NAME_MAX_LEN = IFNAMSIZ - 1;
+
+inline bool isVrfNameValid(const std::string &vrfName)
+{
+    if (vrfName.empty() || vrfName.length() >= IFNAMSIZ || vrfName == "." || vrfName == "..")
+    {
+        return false;
+    }
+
+    for (size_t index = 0; index < vrfName.size(); ++index)
+    {
+        const unsigned char character = static_cast<unsigned char>(vrfName[index]);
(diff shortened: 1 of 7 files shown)


qiluo-msft (approved):

common/vrf.h
```diff
@@ -0,0 +1,42 @@
+#ifndef __VRF__
```
qiluo-msft, line 1: nit: replace the double-underscore guard with `#pragma once` — `__VRF__` is a reserved identifier in C++:

```cpp
// before
#ifndef __VRF__
#define __VRF__
...
#endif

// after
#pragma once
...
```

Every compiler SONiC targets (GCC, Clang) has supported `#pragma once` for 15+ years, and the other headers in `common/` already use it.

ashutosh-agrawal: Thanks, good catch. I opened #1247 to replace the reserved guard with `#pragma once`.

PR #1243, opened 2026-09-01
[schema]: Add FDB_SYNC tables and shared APP_DB naming helpers

tahmed-dev:
#### Why I did it

SONiC learns MAC (FDB) state through the Linux kernel bridge FDB, which makes the
kernel a required data store in the path between the ASIC and FRR. This series
adds the schema needed to synchronize MAC state with FRR directly over the FPM
channel instead, so the kernel is no longer load bearing for EVPN MAC learning.

This PR is the schema-only part. It carries no behaviour on its own.

#### How I did it

- Add the `FDB_SYNC` CONFIG_DB table and its state table constants.
- Drop the EVPN backup NHG APP_DB table, which is unused.
- Add shared APP_DB naming helpers so producers and consumers agree on key
  construction rather than each formatting keys locally.

#### How to verify it

Schema-only change; the constants are exercised by the consumers in the
dependent PRs listed below.

```
make target/debs/trixie/libswsscommon_1.0.0_amd64.deb
```

#### Dependency / merge order

This PR has no dependencies and should merge **first**. The following depend on
it and will not build without it:

- `sonic-swss`: fpmsyncd/fdbsyncd FDB synchronization over FPM
- `sonic-utilities`: `config fdb mac-sync-mode` CLI

#### Which release branch to backport

None requested at this time.

#### Tested branch

Built and packaged against `sonic-net/sonic-buildimage` master.

Add the FDB_SYNC CONFIG_DB and state tables plus shared APP_DB naming helpers.


common/c-api/dbconnector.cpp
```diff
@@ -17,6 +17,20 @@ SWSSResult SWSSSonicDBConfig_initializeGlobalConfig(const char *path, uint8_t ig
     SWSSTry(SonicDBConfig::initializeGlobalConfig(path, ignore_nonexistent));
 }
 
+SWSSResult SWSSSonicDBConfig_getDbSock(const char *dbName, SWSSString *outSock) {
```
pbrisset, line 20: These new C API helpers dereference outSock/outSeparator without checking for nullptr.

Most C API entry points should fail cleanly on invalid caller input rather than segfaulting. Can we return an error if the output pointer is null before assigning *outSock / *outSeparator?

common/schema.h
```diff
@@ -368,6 +379,7 @@ after libswsscommon deb make.
 #define CFG_VLAN_STACKING_TABLE_NAME  "VLAN_STACKING"
 #define CFG_VLAN_TRANSLATION_TABLE_NAME   "VLAN_TRANSLATION"
 #define CFG_FDB_TABLE_NAME         "FDB"
+#define CFG_FDB_SYNC_TABLE_NAME    "FDB_SYNC"
```
pbrisset, line 382: This adds CFG_FDB_SYNC_TABLE_NAME for code references, but it does not add the YANG/schema entry needed by ValidatedConfigDBConnector users.

That is fine if the schema lands in a separate sonic-buildimage/sonic-yang PR, but the dependency should be explicit because sonic-utilities writes FDB_SYNC through ValidatedConfigDBConnector.

